Issue #60, opened 2020-03-16
Play nice with Docker CE installed from docker.com repositories


haplo:
dangerzone fails to install (in Ubuntu 19.10 at least) when Docker is installed through the docker.com repositories.

Steps:
1. Install Docker CE following [these instructions](https://docs.docker.com/install/linux/docker-ce/ubuntu/).
2. Try to install Dangerzone following the wiki (Ubuntu instructions).

Expected: Dangerzone installs just fine.

Got: error because docker.io package cannot be installed (it conflicts with `docker-ce`):

```
$ sudo apt install dangerzone
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 dangerzone : Depends: docker.io but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

micahflee:
I've confirmed your PR fixes the problem. Thanks!

haplo:
Great! Thank you for Dangerzone @micahflee!

PR #61, opened 2020-03-16
Allow dependency on docker-ce package

haplo:
This allows dangerzone to be installed by users who install Docker through the docker.com
repositories, which use the `docker-ce` package instead of `docker.io`.

See #60 .


haplo:
With this patch the DEB package installs. I haven't got time to properly test Dangerzone with the `docker-ce` package yet, but if there are issues I will try to submit other PR.

PR #67, opened 2020-03-28
Fix for docker_installer.py

davidjmeier:
The `is_docker_ready` function passes an invalid argument to `exec_dangerzone_container`.  The passed argument contains a dash instead of an underscore.  This creates the logic situation wherein the UI states that Docker is not installed on Linux derivatives and also issues a `Failed to start docker service` command line error.


davidjmeier:
This bug was found by running dangerzone by cloning and running manually as per documentation for `Development Environment`.

micahflee:
Same as in #68

PR #68, opened 2020-03-28
Fix for tasks.py

davidjmeier:
The `container.py` file has two functions that are incorrectly referenced in `tasks.py`.  These are `document_to_pixels` and `pixels_to_pdf`.  Renaming the args in the `tasks.py` file to appropriately reference those functions.  Without these fixes the software crashes with an error of: `Error: No such command "document-to-pixels"` and `Error: No such command "pixels_to_pdf"`.

### DIFF
--- a/dangerzone/tasks.py
+++ b/dangerzone/tasks.py
@@ -72,7 +72,7 @@ class ConvertToPixels(TaskBase):
     def run(self):
         self.update_label.emit("Converting document to pixels")
         args = [
-            "document-to-pixels",
+            "document_to_pixels",
             "--document-filename",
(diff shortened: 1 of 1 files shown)


davidjmeier:
This bug was found by running dangerzone by cloning and running manually as per documentation for Development Environment.

micahflee:
It turns out this issue was actually because Ubuntu 18.04's python-click package uses an older version of the click module, and the older versions and newer versions deal with command names separately.

I just fixed this in #73 by removing hyphens and underscores from the command names, so it works in both newer and older versions of click.

Issue #118, opened 2021-06-14
Remove Docker Desktop dependency for Windows and Mac


micahflee:
It would amazing if Dangerzone could be a standalone program without requiring the user to separately install Docker Desktop. Dangerzone relies on Linux containers though, and so Docker Desktop runs a Linux VM in order to run the containers in.

If we wanted to remove the Docker Desktop dependency we would have to figure out how to launch our own Linux VM in each OS. I think this would be a considerable amount of work and we would have to come up with a solution for each separate platform, but it would be a great improvement.

If we did this it would also make sense to switch from Docker to bubblewrap for Linux: #95 

### Further evidence / notes (added by @deeplow)

From user interviews, we learned that this was a painpoint for an ex-dangerzone user.

micahflee:
We'll want to use this: https://www.mobyproject.org/

micahflee:
I'm testing on macOS to start with. Specifically, we should use HyperKit to run the actual VM: https://github.com/moby/hyperkit and LinuxKit to create the VM image that runs: https://github.com/linuxkit/linuxkit

Installing LinuxKit:

```sh
brew tap linuxkit/linuxkit
brew install --HEAD linuxkit
```

Grab the linuxkit.yml file and build the image:

```sh
wget https://raw.githubusercontent.com/linuxkit/linuxkit/master/linuxkit.yml
linuxkit build linuxkit.yml
```

Now installing hyperkit (it's also easy to build from source):

```sh
brew install hyperkit
```

Then to run the VM:

```sh
hyperkit -f kexec,linuxkit-kernel
```

And it works!

```sh
$ hyperkit -f kexec,linuxkit-kernel                                                                              
Using fd 5 for I/O notifications
vmx_set_ctlreg: cap_field: 4 bit: 12 unspecified don't care: bit is 0
vmx_set_ctlreg: cap_field: 4 bit: 20 unspecified don't care: bit is 0
vmx_set_ctlreg: cap_field: 3 bit: 13 unspecified don't care: bit is 0
rdmsr to register 0x3a on vcpu 0
rdmsr to register 0x140 on vcpu 0
rdmsr to register 0x64d on vcpu 0
rdmsr to register 0x64e on vcpu 0
rdmsr to register 0x34 on vcpu 0
```

micahflee:
I copy `linuxkit.yml` to `dangerzone.yml` and remove the nginx service, so here's my `dangerzone.yml`:

```yaml
kernel:
[...]

PR #120, opened 2021-06-17
Switch back to Podman in Linux

micahflee:
This removes support for Ubuntu 20.04 (but still works in 20.10, 21.04, and newer) and Debian buster (but still works in bullseye), because they both don't have podman packaged.

This is way smoother in linux because it no longer requires running `dangerzone-container` as root, so I was able to rip out all the logic related to that. Now you don't need root for any of it, and there's no need to install Docker CE from a separate repo. You can just `apt install dangerzone` or `dnf install dangerzone` and then ran `dangerzone` and `dangerzone-cli` and it just works.


dangerzone/gui/__init__.py
```diff
@@ -71,18 +71,6 @@ def gui_main(custom_container, filename):
...
-            return
-        try:
-            if not gui_common.ensure_docker_service_is_started():
-                click.echo("Failed to start docker service")
-                return
-        except AuthorizationFailed:
-            click.echo("Authorization failed")
-            return
-
     # See if we need to install Docker...
```
rugk, line 74: So for MacOS you still need/install Docker? (If so, then the doc, i.e. `BUILD.md` etc. should probably reflect this.)

Or does this also need to be removed?

rugk (changes requested):

dangerzone/gui/__init__.py
micahflee: macOS and Windows will still need docker. Whether it's docker or podman, these are all Linux containers, and if you're not using Linux you'll need a Linux VM to run the containers in. I'm hoping to be able to ship the macOS version with a VM of its own to remove the Docker Desktop dependency though (#118)

micahflee: Add instructions :) https://github.com/firstlookmedia/dangerzone/pull/120/commits/9536a4db41bf0e584f510b88a15cf215ad9ea446

PR #161, opened 2022-05-28
Support Python 3.10

gmarmstrong:
Bump PySide2 to 5.15.2.1, which added support for Python 3.10.

PR #164, opened 2022-05-31
[WIP] Upgrade to Qt 6

gmarmstrong:
***This draft PR is a work in progress.***

*Commits are based on https://github.com/freedomofpress/dangerzone/pull/161 but could be rebased onto main if preferred.*

**Switch from PySide2 to PySide6, which upgrades Dangerzone from Qt 5 to Qt 6.** My immediate motivation was that PySide6 prebuilt binaries support [a wider range of CPU architectures](https://wiki.qt.io/Qt_for_Python#Qt_for_Python) (namely Apple Silicon/M1).

Introduces the following issues, to be resolved before continuing:
- [ ] Build instructions and CI/CD configurations need to be updated to address the dependency change. The apt and rpm repositories don't have `python3-pyside6` packages, so they'll need to be installed through pip or poetry (or for packaging, through stdeb.cfg and pyp2rpm?)
- [ ] [Upstream bug re-introduced by Qt 6.2](https://bugreports.qt.io/browse/QTBUG-87267), causes a minor performance issue on startup and prints a warning like: `qt.qpa.fonts: Populating font family aliases took 76 ms. Replace uses of missing font family "Menlo" with one that exists to avoid this cost.` (font and delay vary by machine)
- [x] [`dangerzone.gui.main_window.ConvertThread`](https://github.com/freedomofpress/dangerzone/blob/851124e863b1bf39d91aaa855a5fb8e0955f0063/dangerzone/gui/main_window.py#L488-L533) causes a TypeError upon finishing (but this doesn't cause any actual issues for an otherwise successful document conversion):
    ```
    [...]
    100% Safe PDF created
      File "/Users/guthrie/Projects/dangerzone/dangerzone/gui/main_window.py", line 512, in run
        self.finished.emit(self.error)
    TypeError: finished() only accepts 0 argument(s), 1 given!
    ```

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -4,7 +4,7 @@ import tempfile
 import subprocess
 import json
 import shutil
-from PySide2 import QtCore, QtGui, QtWidgets
+from PySide6 import QtCore, QtGui, QtWidgets
 from colorama import Style, Fore
 
 from ..common import Common
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -486,7 +486,7 @@ class SettingsWidget(QtWidgets.QWidget):
 
 
 class ConvertThread(QtCore.QThread):
-    finished = QtCore.Signal(bool)
+    is_finished = QtCore.Signal(bool)
     update = QtCore.Signal(bool, str, int)
 
     def __init__(self, global_common, common):
@@ -509,7 +509,7 @@ class ConvertThread(QtCore.QThread):
             ocr_lang,
             self.stdout_callback,
         ):
-            self.finished.emit(self.error)
+            self.is_finished.emit(self.error)
 
     def stdout_callback(self, line):
         try:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,13 +6,13 @@ authors = ["Micah Lee <micah.lee@theintercept.com>"]
 license = "MIT"
 
 [tool.poetry.dependencies]
-python = ">=3.7,<3.10"
+python = ">=3.7,<3.11"
 click = "*"
 appdirs = "*"
 requests = "*"
 macholib = "*"
 termcolor = "*"
-PySide2 = "5.15.2"
+PySide2 = "5.15.2.1"
 colorama = "*"
 psutil = "*"
 cx_freeze = {version = "*", platform = "win32"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ appdirs = "*"
 requests = "*"
 macholib = "*"
 termcolor = "*"
-PySide2 = "5.15.2.1"
+PySide6 = "6.2.*" # 6.2 is LTS
 colorama = "*"
 psutil = "*"
 cx_freeze = {version = "*", platform = "win32"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ appdirs = "*"
 requests = "*"
 macholib = "*"
 termcolor = "*"
-PySide6 = "6.2.*" # 6.2 is LTS
+PySide6 = "6.2.4" # 6.2 is LTS
 colorama = "*"
 psutil = "*"
 cx_freeze = {version = "*", platform = "win32"}
--- a/BUILD.md
+++ b/BUILD.md
@@ -58,7 +58,7 @@ Install Xcode from the App Store.
 
 Install [Docker Desktop](https://www.docker.com/products/docker-desktop). Make sure to choose your correct CPU, either Intel Chip or Apple Chip.
 
-Install Python 3.9.9 [from python.org](https://www.python.org/downloads/release/python-399/).
+Install Python 3.10.4 [from python.org](https://www.python.org/downloads/release/python-3104/).
 
 Install Python dependencies:
 
@@ -110,7 +110,7 @@ The output is in the `dist` folder.
 
 Install [Docker Desktop](https://www.docker.com/products/docker-desktop).
 
-Install Python 3.9.9 (x86) [from python.org](https://www.python.org/downloads/release/python-399/). When installing it, make sure to check the "Add Python 3.9 to PATH" checkbox on the first page of the installer.
+Install Python 3.10.4 (x86) [from python.org](https://www.python.org/downloads/release/python-3104/). When installing it, make sure to check the "Add Python 3.10 to PATH" checkbox on the first page of the installer.
 
 Install [poetry](https://python-poetry.org/). Open PowerShell, and run:
 
@@ -155,7 +155,7 @@ Open a command prompt, cd into the dangerzone directory, and run:
 poetry run python .\setup-windows.py build
 ```
 
-In `build\exe.win32-3.9\` you will find `dangerzone.exe`, `dangerzone-cli.exe`, and all supporting files.
+In `build\exe.win32-3.10\` you will find `dangerzone.exe`, `dangerzone-cli.exe`, and all supporting files.
 
 ### To build the installer
 
--- a/dangerzone/container.py
+++ b/dangerzone/container.py
@@ -135,7 +135,7 @@ def convert(input_filename, output_filename, ocr_lang, stdout_callback):
             success = True
 
     # Clean up
-    shutil.rmtree(tmpdir.name)
+    tmpdir.cleanup()
 
     return success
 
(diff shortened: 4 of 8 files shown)


gmarmstrong (commented):
This was done to fix a TypeError:

```
  File "/Users/guthrie/Projects/dangerzone/dangerzone/gui/main_window.py", line 492, in run
    self.finished.emit(self.error)
TypeError: finished() only accepts 0 argument(s), 1 given!
```

After a second look, I'm not sure about this solution. I believe the TypeError was caused by Qt 6 [defining `void QThread::finished()`](https://doc.qt.io/qt-6.2/qthread.html#finished) as `[private signal]`, whereas Qt 5 [defined it](https://doc.qt.io/qt-5/qthread.html#finished) as a `[signal]`. So, PySide6 raises an exception when we try to manually `emit` that private signal.

We may want to instead use [`void QThread::exit`](https://doc.qt.io/qt-6.2/qthread.html#exit).

dangerzone/gui/main_window.py
```diff
@@ -486,7 +486,7 @@ def start_button_clicked(self):
 
 
 class ConvertThread(QtCore.QThread):
-    finished = QtCore.Signal(bool)
+    is_finished = QtCore.Signal(bool)
```
gmarmstrong, line 489: This fixed a TypeError, `line 492, in run self.finished.emit(self.error) TypeError: finished() only accepts 0 argument(s), 1 given!`

But after a second look, I'm not sure about this solution. I believe the TypeError was caused by Qt 6 [defining `void QThread::finished()`](https://doc.qt.io/qt-6.2/qthread.html#finished) as `[private signal]`, whereas Qt 5 had [defined it](https://doc.qt.io/qt-5/qthread.html#finished) as a `[signal]`. PySide6 raises an exception when we try to manually `emit` that private signal.

dangerzone/gui/main_window.py
```diff
@@ -509,7 +509,7 @@ def run(self):
             ocr_lang,
             self.stdout_callback,
         ):
-            self.finished.emit(self.error)
+            self.is_finished.emit(self.error)
```
gmarmstrong, line 512: To fix the `finished` TypeError, we may want to instead use [`void QThread::exit`](https://doc.qt.io/qt-6.2/qthread.html#exit).

dangerzone/gui/main_window.py
gmarmstrong: A more elegant solution could come from avoiding subclassing QThread at all.

deeplow: This might actually have been a bug introduced in https://github.com/freedomofpress/dangerzone/commit/ea47a2e92ce957f0f9e104290bc433f047dddd99. Among other things, this commit adds a bool argument to the `finished` signal. However this didn't make much sense because [the slot received zero arguments](https://github.com/freedomofpress/dangerzone/commit/ea47a2e92ce957f0f9e104290bc433f047dddd99#diff-cca675cf576ebaa4fde44bfc6790124ccc293f2edcbae1d4068c347cc8a83102R540).

It sounds like when the argument became private (which sounds weird in python lingo, where nothing is really private) it now started complaining. So in a way, this was a bug that just didn't show itself before.

deeplow, line 489: Per my testing the signal `finished` is emitted anyways from a QThread when it finishes anyways. So we don't even need it. This avoids the whole typing issue.

dangerzone/gui/main_window.py
deeplow:
```suggestion
```
If we remove this (as it doesn't seem to be needed), we avoid the typing issue altogether.

deeplow (changes requested):
Thanks for yet another PR.

gmarmstrong:
> > * [Upstream bug re-introduced by Qt 6.2](https://bugreports.qt.io/browse/QTBUG-87267), causes a minor performance issue on startup and prints a warning like: `qt.qpa.fonts: Populating font family aliases took 76 ms. Replace uses of missing font family "Menlo" with one that exists to avoid this cost.` (font and delay vary by machine)
> 
> I haven't been able to reproduce this in either a windows machine or a linux one (running fedora). In which machine did this happen?

On macOS Monterey 12.4 (M1 chip). The comments on the Qt bug report suggest that the issue might have only returned for macOS.

pyproject.toml
```diff
@@ -6,13 +6,13 @@ authors = ["Micah Lee <micah.lee@theintercept.com>"]
...
 [tool.poetry.dependencies]
-python = ">=3.7,<3.10"
+python = ">=3.7,<3.11"
 click = "*"
 appdirs = "*"
 requests = "*"
 macholib = "*"
 termcolor = "*"
-PySide2 = "5.15.2"
+PySide6 = "6.2.*" # 6.2 is LTS
```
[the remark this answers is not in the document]
gmarmstrong:
```suggestion
PySide6 = "6.2.4" # 6.2 is LTS
```

dangerzone/gui/main_window.py
gmarmstrong: Nice, looks good to me if that works on your end (I haven't tested it myself).

gmarmstrong: Merged your change into this PR branch, see 5af3d73.

dangerzone/gui/main_window.py
gmarmstrong: Resolved, see 5af3d73.

deeplow: resolved via https://github.com/freedomofpress/dangerzone/commit/5af3d7396e4004ce83272710db615fbd1be446a7

gmarmstrong:
> > * The apt and rpm repositories don't have `python3-pyside6` packages, so they'll need to be installed through pip or poetry (or for packaging, through stdeb.cfg and pyp2rpm?)
> 
> 
> 
> I've been looking into this a bit. The issue is that even if we are able to install it with poetry, when building RPMs and DEBs won't be able to include these depetencies. Perhaps it can be installed via poetry at install time, but it feels like we'd be stepping over the distro's package manager and that could lead to issues in the future and definitely feels to "hacky" to be stable.
> 
> 
> 
> To me this seems like a blocker for this PR, unfortunately. But if I find a viable solution, I'll post it here.

Are there really no other packages in the APT and RPM repositories that depend on Qt6?

deeplow:
> Are there really no other packages in the APT and RPM repositories that depend on Qt6?

I think this is exactly the case! The closest I've been able to find is the persepolis package, which supports qt5 (via PyQt5) and PySide6. In their Debian package they ship the PyQt5 as a dependency instead of PySide6.

And this approach of supporting [is not super elegant](https://github.com/persepolisdm/persepolis/blob/d963e8dca7b78fd33497dd5c7ddb02aa3724d71f/persepolis/scripts/about.py#L16), let's say:

```
try:
    from PySide6.QtCore import Qt, QSize, QPoint, QFile, QIODevice, QTextStream
    from PySide6.QtWidgets import QWidget
    from PySide6.QtGui import QIcon
except:
    from PyQt5.QtCore import Qt, QSize, QPoint, QFile, QIODevice, QTextStream
    from PyQt5.QtWidgets import QWidget
    from PyQt5.QtGui import QIcon
```

In our case, it would probably look the same but with PySide2 instead of PyQt5.

gmarmstrong:
Well, that's pretty disappointing, especially since #166 and #168 depend on this. This is what I get for writing PRs that depend upon other PRs. 😅

I see several options here:

1. Review the dependent PRs and, if you want to add those changes, then keep the whole commit history and downgrade to Qt 5 on top of it all.

2. Rebase the dependent PRs onto main (or onto #161).

3. Wait some time, hoping that 9 months from now, [when Qt 5 reaches EOL on 26 May 2023](https://endoflife.date/qt), the APT and RPM repositories will by then include a `python3-pyside6` package (or an equivalent).

---

I will note that #161, #167, and #170 are all standalone changes that should neither depend upon nor conflict with this chain of PRs.

deeplow:
Yes, the review complexity is quite high because of that. I'd go for 2. at least for #168. Is this something you can do? Or shall I? Unit tests are probably needed soon and since you already made a base for it...

eloquence:
We're not adding this to the next release milestone, as it's not strictly necessary to support M1/M2 Macs, but should pick this up again soon given the looming EOL for Qt5.

deeplow:
We won't be adding support for PySide6 on Linux at this time, but we will add it to Windows and MacOS.

Closing this in favor https://github.com/freedomofpress/dangerzone/pull/296. Thanks @gmarmstrong for starting this discussion and suggesting a change. The new PR includes some of your suggestions.

PR #175, opened 2022-07-20
Update CI

deeplow:
Removes EOL distros and adds more recent ones.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,6 +132,17 @@ jobs:
               echo
             done
 
+  build-ubuntu-jammy:
+    docker:
+      - image: ubuntu:22.04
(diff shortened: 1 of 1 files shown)


eloquence:
As I understand it from https://endoflife.date/ubuntu both Hirsute and Impish are EOL as well - what do you think of removing those as well?

deeplow:
> 

Oh, that's

> As I understand it from https://endoflife.date/ubuntu both Hirsute and Impish are EOL as well - what do you think of removing those as well?

Oh, that's right! Thanks for spotting it. I didn't know about that website -- super useful.

deeplow:
That would explain the repo fetch failures they were getting in the CI.

PR #181, opened 2022-08-10
Replace prints with logging

### LINKED ISSUE #144: Certain characters in the filename stop Dangerzone from working
It looks like the characters `“` and `”` prevent Dangerzone from working, as in the filename: `“Wash Brains, Cleanse Hearts”_ Evidence from Chinese Government Documents about the Nature and Extent of Xinjiang’s Extrajudicial Internment Campaign _.pdf`

deeplow:
fixes #144: printing non-ascii characters in a macOS application
opened directly from finder would sometimes lead to an error
message in /var/log/system.log similar to this:

  Failed to execute script 'dangerzone' due to unhandled exception:
  'ascii' codec can't encode character '\u201c' in position 1:
  ordinal not in range(128)

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -2,6 +2,7 @@ import os
 import sys
 import json
 import click
+import logging
 from colorama import Fore, Style
 
 from .global_common import GlobalCommon
@@ -19,6 +20,7 @@ def print_header(s):
 @click.option("--ocr-lang", help="Language to OCR, defaults to none")
 @click.argument("filename", required=True)
 def cli_main(output_filename, ocr_lang, filename):
+    setup_logging()
     global_common = GlobalCommon()
     common = Common()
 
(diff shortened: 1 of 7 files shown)


deeplow:
What's missing is a way to send this to macOS's syslog. So far I haven't had luck with logging library, [even following some tips](https://stackoverflow.com/questions/30125412/how-can-i-log-from-python-to-syslog-with-either-sysloghandler-or-syslog-on-mac-o). The official docs [mention nothing about this](https://docs.python.org/3/library/logging.handlers.html#sysloghandler).

So far I've only had luck with sending stuff to macOS's syslog via the `syslog` module, but I'm using logging everywhere and I'd prefer not to have to make a wrapper logging function just to deal with this edge case of MacOS debugging.

deeplow:
From an umerged [python doc PR](https://github.com/python/cpython/pull/94803/files):
> **note** On macOS 12.x (Monterey), Apple has changed the behaviour of their
syslog daemon - it no longer listens on a domain socket. Therefore, you cannot
expect :class:`SysLogHandler` to work on this system.
> 
> See https://github.com/python/cpython/issues/91070 for more information.

So it seems this is a problem with the logging module and is appears unlikely to be resolved soon.

PR #183, opened 2022-08-15
Container hardening: run in container as non-root, prevent privesc & disable capabilities

### LINKED ISSUE #169: Container hardening (revisited): non root in container + no-new-privileges
### User privileges in the container

- The containers currently run the conversion processes as root (root user of the containers, not of the host). This is easily fixed without causing any issues, at least on macOS, by adding `USER user` just below `RUN adduser [...] user` [in the Dockerfile](https://github.com/freedomofpress/dangerzone/blob/d8adb2e9cc1bee17ffe21a006cfc7878e4a13bc3/container/Dockerfile#L94). As it stands, the unprivileged user is never actually used. [This has been fixed before](https://github.com/freedomofpress/dangerzone/issues/5), but the problem has been reinstated.

### Docker/kernel security options

-  Dangerzone [used to call](https://github.com/freedomofpress/dangerzone/commit/9278848adf3da7f7f972d3ec53cd27ad81d978b7) `docker run` with `--security-opt=no-new-privileges:true`, but that option [was removed during a refactor](https://github.com/freedomofpress/dangerzone/commit/488dca4a714507d589c03c46121c3e7ad94699c6#diff-c71b3544ff6d4565f66adbb800868b57681888dd909b190a94227a41500c9db3L150-L152) and is still missing. Can/should it be restored? [Podman has an option of the same name.](https://docs.podman.io/en/latest/markdown/podman-run.1.html#security-opt-option)

### Related discussions

- freedomofpress/dangerzone/issues/52
- firstlookmedia/dangerzone-converter/pull/7
- freedomofpress/dangerzone/issues/118
- freedomofpress/dangerzone/pull/119

deeplow:
Fixes #169

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -91,7 +91,8 @@ COPY dangerzone.py /usr/local/bin/
 RUN chmod +x /usr/local/bin/dangerzone.py
 
 # Add the unprivileged user
-RUN adduser -h /home/user -s /bin/sh -D user
+RUN adduser -s /bin/sh -D dangerzone
+USER dangerzone
 
 # /tmp/input_file is where the first convert expects the input file to be, and
 # /tmp where it will write the pixel files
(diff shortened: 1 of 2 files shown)


deeplow:
(rebased)

deeplow:
Tested on mac (intel), linux and windows.

PR #185, opened 2022-08-16
exclude python 3.10 minor verision mentions

deeplow:


deeplow:
Opened here by mistake

PR #166, opened 2022-06-05
Add type hints and refactor most *Common methods to be static or functional

gmarmstrong:
***This draft PR depends on a work-in-progress PR.***

*Builds upon commits from #164. This branch begins at https://github.com/freedomofpress/dangerzone/commit/616ba55db7c00f0c2fe148e5b54bfa4302788071. See [this diff](https://github.com/gmarmstrong/dangerzone/compare/qt6...gmarmstrong:typing) for comparison.*

Does some hefty refactoring:
* First adds type hinting (verified by static type checking with the help of [mypy](https://mypy.readthedocs.io)).
* Then refactors most methods of GlobalCommon and GuiCommon to be static.
* Those static methods are then moved to a dangerzone.util module (alias "dzutil").
* Ultimately, this allows classes in dangerzone.gui.main_window to avoid passing around objects except when necessary (i.e., in cases where window-specific settings need to share a state. The discovery here is that in most cases, classes could stand alone).
* After much more refactoring, GlobalCommon has been deleted. Settings are now held by GuiCommon (the CLI never uses settings.json).
* GuiCommon now subclasses Common.

Fixes #122 

If this design doesn't feel right to others, feel free to speak up about it! It feels more intuitive to me, but it's a big change.

Issue #193, opened 2022-08-22
Handle container install failure


deeplow:
The [`install_container()`](https://github.com/freedomofpress/dangerzone/blob/7bac3eb6b16c3ac2a2529105e2fa28f2bab3f7c3/dangerzone/global_common.py#L450) method returns false when it fails but none of its calls check the return value.

This should also be passed on to the user interface so the user can see the issue.

deeplow:
I am getting a similar issue to https://github.com/freedomofpress/dangerzone/issues/152 but I think that is a consequence of this lack of failure handling.

In short

<details><summary>$ dangerzone</summary>

```
[INFO] Installing Dangerzone container image...
Getting image source signatures
Copying blob 48a31ab7de6f [=====================>----------------] 900.8MiB / 1.5GiB
Copying blob b56b9d864269 done  
Copying blob 64a1a52c74ff done  
Copying blob 4965dc1ce878 done  
Copying blob 99f8e046e98a done  
Copying blob 994393dc58e7 skipped: already exists  
Error: payload does not match any of the supported image formats (oci, oci-archive, dir, docker-archive)
[ERROR] Failed to install the container image
[INFO] > /usr/bin/podman run --network none -u dangerzone --security-opt no-new-privileges --userns keep-id --cap-drop all --rm -v /home/user/dangerzone/test_docs/sample.doc:/tmp/input_file -v /home/user/.config/dangerzone/tmp/tmprkiz5sfc/pixels:/dangerzone dangerzone.rocks/dangerzone /usr/bin/python3 /usr/local/bin/dangerzone.py document-to-pixels
[ERROR] Invalid JSON returned from container: Trying to pull dangerzone.rocks/dangerzone:latest...

[ERROR] Invalid JSON returned from container: Error: initializing source docker://dangerzone.rocks/dangerzone:latest: pinging container registry dangerzone.rocks: invalid status code from registry 404 (Not Found)

[ERROR] documents-to-pixels failed
````

</details>

<details><summary>$ df -h</summary>

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda3      9.6G  6.6G  2.5G  73% /
none            9.6G  6.6G  2.5G  73% /usr/lib/modules
devtmpfs        4.0M     0  4.0M   0% /dev
tmpfs           1.0G   84K  1.0G   1% /dev/shm
tmpfs            58M  760K   57M   2% /run
[...]

PR #203, opened 2022-09-15
Reduce "global_common" coupling by moving methods that could be static onto "semantically-closer" py files

deeplow:
Based on work initially made by @gmarmstrong on https://github.com/freedomofpress/dangerzone/pull/166:
  - moves container-specific code out of `global_common.py` and into `container.py`
  - creates a `util.py` for static methods used through the whole app
  - move banner code from `global_common` onto `cli.py` given that it's only displayed there
  - updates tests to reflect these changes
  - move `ocr_languages` from `global_common` onto its own `json` file in `share/ocr-languages.json` to simplify `global_common` logic


deeplow:
@gmarmstrong let me know what you think. I have limited the scope of this PR to moving things out of global common. 
The refactor to make static the code in `gui_common` will be in a future PR.

dangerzone/container.py
```diff
@@ -29,6 +27,106 @@
...
+
+def get_runtime() -> str:
+    container_tech = get_runtime_name()
+    runtime = shutil.which(container_tech)
+    if runtime is None:
+        raise NoContainerTechException(container_tech)
+    return runtime
+
+
+def install() -> Optional[bool]:
```
gmarmstrong, line 52: I would either return `bool`, or keep the `Optional` but raise an exception (to be caught) upon failure to install.

dangerzone/gui/main_window.py
```diff
@@ -104,7 +104,7 @@ def __init__(self, global_common: GlobalCommon) -> None:
         self.global_common = global_common
 
     def run(self) -> None:
-        self.global_common.install_container()
+        container.install()
```
gmarmstrong, line 107: Check return value to ensure that the installation is successful.

dangerzone/cli.py
```diff
@@ -90,7 +92,7 @@ def cli_main(
             exit(1)
 
     # Ensure container is installed
-    global_common.install_container()
+    container.install()
```
gmarmstrong, line 95: Check return value to ensure that the installation is successful.

gmarmstrong: I see that I didn't do that either, so maybe I'm just forgetting something here. Up to you!

gmarmstrong: That may also just be outside the scope of this PR.

gmarmstrong: Oh, never mind all of this! You added that "container-tech-checking logic" elsewhere. In that case, I'd just change the comment now to say "install container".

gmarmstrong (approved):
Looks great!

dangerzone/container.py
deeplow: Changed to bool. I have an issue already for handling container installation failure https://github.com/freedomofpress/dangerzone/issues/193.

deeplow:
Thanks for the review @gmarmstrong!

Issue #205, opened 2022-09-15
0.3.1 release: parallel document conversion fails


deeplow:
If we open two Dangerzone windows via the tray menu's "new window" both windows get closed as soon as the first document finishes the conversion.

Release: 0.3.1

### Terminal log

From taking a look at the log, it seems it forcefully exits the whole application, leaving the conversion threat without being properly terminated or waited for.

```
$ poetry run dev_scripts/dangerzone

(python3:78763): Gdk-CRITICAL **: 16:08:27.352: gdk_atom_intern: assertion 'atom_name != NULL' failed

(python3:78763): Gdk-CRITICAL **: 16:08:27.353: gdk_atom_intern: assertion 'atom_name != NULL' failed
> /usr/bin/podman run --network none -v /home/user/dangerzone/test_docs/sample.pptx:/tmp/input_file -v /home/user/.config/dangerzone/tmp/tmpojlc23em/pixels:/dangerzone dangerzone.rocks/dangerzone /usr/bin/python3 /usr/local/bin/dangerzone.py document-to-pixels
0% Converting to PDF using LibreOffice
3% Separating document into pages
5% Converting page 1/3 to pixels
20% Converting page 2/3 to pixels
35% Converting page 3/3 to pixels
50% Converted document to pixels
> /usr/bin/podman run --network none -v /home/user/.config/dangerzone/tmp/tmpojlc23em/pixels:/dangerzone -v /home/user/.config/dangerzone/tmp/tmpojlc23em/safe:/safezone -e OCR=1 -e OCR_LANGUAGE=eng dangerzone.rocks/dangerzone /usr/bin/python3 /usr/local/bin/dangerzone.py pixels-to-pdf
50% Converting page 1/3 from pixels to searchable PDF
> /usr/bin/podman run --network none -v /home/user/dangerzone/test_docs/sample.odp:/tmp/input_file -v /home/user/.config/dangerzone/tmp/tmp_6uk9hvq/pixels:/dangerzone dangerzone.rocks/dangerzone /usr/bin/python3 /usr/local/bin/dangerzone.py document-to-pixels
0% Converting to PDF using LibreOffice
65% Converting page 2/3 from pixels to searchable PDF
80% Converting page 3/3 from pixels to searchable PDF
3% Separating document into pages
5% Converting page 1/3 to pixels
20% Converting page 2/3 to pixels
35% Converting page 3/3 to pixels
50% Converted document to pixels
[...]

Issue #204, opened 2022-09-15
"New Window" does not work in 0.3.2


deeplow:
When one opens a new window from the tray menu, on the `0.3.2 release` we get the following error:

```python
Traceback (most recent call last):
  File "/home/user/dangerzone/dangerzone/gui/__init__.py", line 91, in select_document
    and windows[list(windows.keys())[0]].common.input_filename == None
  File "/home/user/dangerzone/dangerzone/common.py", line 23, in input_filename
    raise RuntimeError("Input filename has not been set yet.")
RuntimeError: Input filename has not been set yet.
````

deeplow:
Even though this is a regression from the previous release, from my testing on `0.3.1` (previous release) an additional bug related to this had gone unnoticed. I have reported it [now](https://github.com/freedomofpress/dangerzone/issues/205).

So, in short, even though this is a regression in terms of UX (window buttons don't work), it is not a regression compared to the previous release in terms of functionality.

deeplow:
Oddly this doesn't seem to affect MacOS

eloquence:
The "New Window" systray feature no longer exists in the development version and has been replaced with multi-document support, closing.

PR #208, opened 2022-09-19
rename "common" to  "document" & simplify "new window" window logic

deeplow:
Simplify code logic for creating and handling documents through the renaming of `Common` class to `DocumentHolder` and associated simplifications.

When we support bulk document conversion we'll need to keep track of each document's state. This refactor opens way for that. A document holder does not represent a document in itself (e.g. can be instantiated without a document) but stores the input name, output name, and in the future possibly the conversion progress.

Clarification renames: some of the names were changed to better reflect the classes' or files' purpose.
- **renames:** `common.py::Common` -> `document.py::DocumentHandler`
- **renames:** `global_common.py::GlobalCommon` ->`logic.py::DangerzoneCore`
- **renames:** `gui/common.py::GuiCommon` -> `gui/common.py::DangerzoneGui`

Associated logic simplication:
- deduplicates input / output filename validation now done in `DocumentHandler`, when a new filename is added, throwing an `DocumentFilenameException` if that validation fails. (code was previously duplicated in `gui_main` and `cli_main`).
- simplifies "new window" creation logic (and associated `DocumentHandler` instantiation)
   now when the users click *new window*, it simply opens a new window without extra validation to see if any window is already open and doesn't have a document associated to it.

And lastly, it adds unit tests for `document.py` (100% coverage)

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -28,7 +28,6 @@ def cli_main(
 ) -> None:
     setup_logging()
     global_common = GlobalCommon()
-    document = Document()
 
     display_banner()
 
@@ -44,7 +43,7 @@ def cli_main(
         click.echo("Invalid filename")
         exit(1)
 
-    document.input_filename = os.path.abspath(filename)
+    document = Document(os.path.abspath(filename))
 
     # Validate safe PDF output filename
     if output_filename:
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -8,8 +8,8 @@ import click
 from colorama import Back, Fore, Style
 
 from . import container
-from .common import Common
 from .container import convert
+from .document import Document
 from .global_common import GlobalCommon
 from .util import get_version
 
@@ -28,7 +28,7 @@ def cli_main(
 ) -> None:
     setup_logging()
     global_common = GlobalCommon()
-    common = Common()
+    document = Document()
 
     display_banner()
 
@@ -44,7 +44,7 @@ def cli_main(
         click.echo("Invalid filename")
         exit(1)
 
-    common.input_filename = os.path.abspath(filename)
+    document.input_filename = os.path.abspath(filename)
 
     # Validate safe PDF output filename
     if output_filename:
@@ -63,18 +63,18 @@ def cli_main(
             click.echo("Safe PDF filename is not writable")
             exit(1)
 
-        common.output_filename = os.path.abspath(output_filename)
+        document.output_filename = os.path.abspath(output_filename)
 
     else:
-        common.output_filename = (
-            f"{os.path.splitext(common.input_filename)[0]}-safe.pdf"
+        document.output_filename = (
+            f"{os.path.splitext(document.input_filename)[0]}-safe.pdf"
         )
         try:
-            with open(common.output_filename, "wb"):
+            with open(document.output_filename, "wb"):
                 pass
         except:
             click.echo(
-                f"Output filename {common.output_filename} is not writable, use --output-filename"
+                f"Output filename {document.output_filename} is not writable, use --output-filename"
             )
             exit(1)
 
@@ -110,13 +110,13 @@ def cli_main(
             click.echo(f"Invalid JSON returned from container: {line}")
 
     if convert(
-        common.input_filename,
-        common.output_filename,
+        document.input_filename,
+        document.output_filename,
         ocr_lang,
         stdout_callback,
     ):
         print_header("Safe PDF created successfully")
-        click.echo(common.output_filename)
+        click.echo(document.output_filename)
         exit(0)
     else:
         print_header("Failed to convert document")
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -7,7 +7,7 @@ from typing import Optional
 import click
 from colorama import Back, Fore, Style
 
-from . import container, errors
+from . import args, container, errors
 from .container import convert
 from .document import Document
 from .global_common import GlobalCommon
@@ -20,9 +20,13 @@ def print_header(s: str) -> None:
 
 
 @click.command()
-@click.option("--output-filename", help="Default is filename ending with -safe.pdf")
+@click.option(
+    "--output-filename",
+    callback=args.validate_output_filename,
+    help="Default is filename ending with -safe.pdf",
+)
 @click.option("--ocr-lang", help="Language to OCR, defaults to none")
-@click.argument("filename", required=True)
+@click.argument("filename", required=True, callback=args.validate_input_filename)
 @errors.handle_document_errors
 def cli_main(
     output_filename: Optional[str], ocr_lang: Optional[str], filename: str
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -7,7 +7,7 @@ from typing import Optional
 import click
 from colorama import Back, Fore, Style
 
-from . import container
+from . import container, errors
 from .container import convert
 from .document import Document
 from .global_common import GlobalCommon
@@ -23,6 +23,7 @@ def print_header(s: str) -> None:
 @click.option("--output-filename", help="Default is filename ending with -safe.pdf")
 @click.option("--ocr-lang", help="Language to OCR, defaults to none")
 @click.argument("filename", required=True)
+@errors.handle_document_errors
 def cli_main(
     output_filename: Optional[str], ocr_lang: Optional[str], filename: str
 ) -> None:
@@ -31,51 +32,15 @@ def cli_main(
 
     display_banner()
 
-    # Validate filename
-    valid = True
-    try:
-        with open(os.path.abspath(filename), "rb") as f:
-            pass
-    except:
-        valid = False
-
-    if not valid:
-        click.echo("Invalid filename")
-        exit(1)
-
     document = Document(os.path.abspath(filename))
 
     # Validate safe PDF output filename
     if output_filename:
-        valid = True
-        if not output_filename.endswith(".pdf"):
-            click.echo("Safe PDF filename must end in '.pdf'")
-            exit(1)
-
-        try:
-            with open(os.path.abspath(output_filename), "wb"):
-                pass
-        except:
-            valid = False
-
-        if not valid:
-            click.echo("Safe PDF filename is not writable")
-            exit(1)
-
         document.output_filename = os.path.abspath(output_filename)
-
     else:
         document.output_filename = (
             f"{os.path.splitext(document.input_filename)[0]}-safe.pdf"
         )
-        try:
-            with open(document.output_filename, "wb"):
-                pass
-        except:
-            click.echo(
-                f"Output filename {document.output_filename} is not writable, use --output-filename"
-            )
-            exit(1)
 
     # Validate OCR language
     if ocr_lang:
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -10,7 +10,7 @@ from colorama import Back, Fore, Style
 from . import args, container, errors
 from .container import convert
 from .document import SAFE_EXTENSION, Document
-from .global_common import GlobalCommon
+from .logic import DangerzoneCore
 from .util import get_version
 
 
@@ -32,7 +32,7 @@ def cli_main(
     output_filename: Optional[str], ocr_lang: Optional[str], filename: str
 ) -> None:
     setup_logging()
-    global_common = GlobalCommon()
+    dangerzone = DangerzoneCore()
 
     display_banner()
 
@@ -47,14 +47,14 @@ def cli_main(
     # Validate OCR language
     if ocr_lang:
         valid = False
-        for lang in global_common.ocr_languages:
-            if global_common.ocr_languages[lang] == ocr_lang:
+        for lang in dangerzone.ocr_languages:
+            if dangerzone.ocr_languages[lang] == ocr_lang:
                 valid = True
                 break
         if not valid:
             click.echo("Invalid OCR language code. Valid language codes:")
-            for lang in global_common.ocr_languages:
-                click.echo(f"{global_common.ocr_languages[lang]}: {lang}")
+            for lang in dangerzone.ocr_languages:
+                click.echo(f"{dangerzone.ocr_languages[lang]}: {lang}")
             exit(1)
 
     # Ensure container is installed
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -9,7 +9,7 @@ from colorama import Back, Fore, Style
 
 from . import args, container, errors
 from .container import convert
-from .document import Document
+from .document import SAFE_EXTENSION, Document
 from .global_common import GlobalCommon
 from .util import get_version
 
@@ -23,7 +23,7 @@ def print_header(s: str) -> None:
 @click.option(
     "--output-filename",
     callback=args.validate_output_filename,
-    help="Default is filename ending with -safe.pdf",
+    help=f"Default is filename ending with {SAFE_EXTENSION}",
 )
 @click.option("--ocr-lang", help="Language to OCR, defaults to none")
 @click.argument("filename", required=True, callback=args.validate_input_filename)
@@ -38,13 +38,11 @@ def cli_main(
 
     document = Document(filename)
 
-    # Validate safe PDF output filename
+    # Set PDF output filename
     if output_filename:
         document.output_filename = output_filename
     else:
-        document.output_filename = (
-            f"{os.path.splitext(document.input_filename)[0]}-safe.pdf"
-        )
+        document.set_default_output_filename()
 
     # Validate OCR language
     if ocr_lang:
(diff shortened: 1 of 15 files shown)


deeplow:
I was getting errors with the last commit when we ran the cli without a full path. This [is now fixed](https://github.com/freedomofpress/dangerzone/compare/6c615a235f2aed0acb6780287a7d0f547742fc78..c20021b686b60ff5ffa34fddb5d29a90db702499).

```
$ poetry run dev_scripts/dangerzone-cli tests/test_docs/sample.pdf 
╭──────────────────────────╮
│           ▄██▄           │
│          ██████          │
│         ███▀▀▀██         │
│        ███   ████        │
│       ███   ██████       │
│      ███   ▀▀▀▀████      │
│     ███████  ▄██████     │
│    ███████ ▄█████████    │
│   ████████████████████   │
│    ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀    │
│                          │
│    Dangerzone v0.3.2     │
│ https://dangerzone.rocks │
╰──────────────────────────╯

> /usr/bin/podman run --network none -u dangerzone --security-opt no-new-privileges --userns keep-id --cap-drop all --rm -v tests/test_docs/sample.pdf:/tmp/input_file -v /home/user/.config/dangerzone/tmp/tmpzkabfaa4/pixels:/dangerzone dangerzone.rocks/dangerzone /usr/bin/python3 /usr/local/bin/dangerzone.py document-to-pixels
Invalid JSON returned from container: Error: error creating named volume "tests/test_docs/sample.pdf": error running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument

documents-to-pixels failed

Failed to convert document
```

tests/test_document.py
```diff
@@ -0,0 +1,67 @@
...
+    Attempts to read a document's filename when no doc has been set
+    """
+    d = DocumentHolder()
+    with pytest.raises(DocumentFilenameException):
+        d.input_filename
+
+
+def test_input_file_non_existing():
+    with pytest.raises(DocumentFilenameException):
+        DocumentHolder("fake-dir/non-existing-file.pdf")
```
deeplow, line 31: I need to fix the `/` and make it platform-independent. This won't work on Windows.

apyrgio:
I'm testing out this PR in order to see if #204 is resolved, and I have some quick questions:

1. What is the proper way to test this on Windows environments? In 0.3.2, when I click on New Window from the systray, I don't see any window being opened, nor an exception somewhere (I checked the Event Log Viewer as well). I suppose that the exception is logged somewhere, but I haven't found where yet. Any pointers would be helpful.
2. In Linux, I have reproduced the error, and I'm seeing that with your fix, two windows get opened. However, when you close either of them, both of them close, which is unexpected. Am I missing something here?

apyrgio:
> I think it would be logged on the terminal from which you launch dangerzone. But I've also had issues where exceptions are consumed somewhere along the codepath. It could be that.

It might be that I'm testing it out by running the `.exe`, not following the instructions on `BUILD.md`. I'll check that out.

> So two windows get open when only one should open? Or that part is working?

I meant that this part is working correctly, yes.

> Yes, that is https://github.com/freedomofpress/dangerzone/issues/205 which was present in linux in 0.3.1. This is fixed in the other PR, I think.

Oh, I thought that #205 was affecting windows that closed due to conversion completion, not due to a user action. Ok, then we'll see this later on.

dangerzone/gui/__init__.py
```diff
@@ -85,50 +86,42 @@ def delete_window(window_id: str) -> None:
         del windows[window_id]
 
     # Open a document in a window
-    def select_document(filename: Optional[str] = None) -> bool:
-        if (
-            len(windows) == 1
-            and windows[list(windows.keys())[0]].document.input_filename == None
-        ):
-            window = windows[list(windows.keys())[0]]
```
apyrgio, line 93: If I understand correctly, we now remove this logic whereby only one window could get created. This should be documented in the commit message that introduces this removal, since currently we are just saying we are simplifying the new window logic.

dangerzone/document.py
```diff
@@ -12,9 +12,9 @@ class Document(object):
     The Common class is a singleton of shared functionality throughout an open dangerzone window
     """
 
-    def __init__(self) -> None:
+    def __init__(self, input_file_path: str) -> None:
```
apyrgio: The "simplify 'new window' logic" commit tries to do three things at once:

1. Change the way we instantiate the document class.
2. Moves the filename checks out of the new window function.
3. Simplifies the window logic a bit.

Specifically for (2), we move this logic again under the `DocumentHolder()` class in a subsequent commit. So, all in all, it's better to break these three things in different commits, and merge them with any amendments we do in subsequent commits. This way, each commit has a clear purpose, and it doesn't introduce logic that will be altered.

dangerzone/cli.py
```diff
@@ -31,51 +31,23 @@ def cli_main(
...
+        if output_filename:
+            document.output_filename = os.path.abspath(output_filename)
+        else:
+            document.output_filename = (
+                f"{os.path.splitext(document.input_filename)[0]}-safe.pdf"
             )
-            exit(1)
+    except DocumentFilenameException as e:
+        click.echo(str(e))
+        exit(1)
```
apyrgio, line 49: Why do we need to catch `DocumentFilenameException` in various places in the code? Can we catch them in the main function instead?

dangerzone/document.py
```diff
@@ -12,29 +12,57 @@ class DocumentHolder(object):
...
     def output_filename(self, filename: str) -> None:
+        # validate output filename
+        if not filename.endswith(".pdf"):
+            raise DocumentFilenameException("Safe PDF filename must end in '.pdf'")
+        try:
+            with open(os.path.abspath(filename), "wb"):
+                pass
+        except PermissionError:
+            raise DocumentFilenameException("Safe PDF filename is not writable")
+
```
apyrgio, line 66: I fully agree with moving these checks closer to the `DocumentHolder` class. However, this way we can't reuse them from other contexts. For instance, it would make the CLI logic much simpler if we used Click validators: https://click.palletsprojects.com/en/8.1.x/options/#callbacks-for-validation. If we moved this logic into class/static methods, it would be trivial to create those.

dangerzone/document.py
```diff
@@ -12,29 +12,57 @@ class DocumentHolder(object):
...
         self._output_filename = filename
+
+
+class DocumentFilenameException(Exception):
+    def __init__(self, message):
+        super().__init__(message)
```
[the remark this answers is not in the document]
apyrgio: Also, the `DocumentFilenameException()` class is introduced in 'simplify "new window" logic", and is patched in "add unit tests for document.py". It's best to keep thing clean and merge the patch in the commit that introduces this class.

tests/test_document.py
```diff
@@ -0,0 +1,67 @@
...
+    d = DocumentHolder()
+    d.input_filename = sample_doc
+
+
+def test_input_file_none():
+    """
+    Attempts to read a document's filename when no doc has been set
+    """
+    d = DocumentHolder()
+    with pytest.raises(DocumentFilenameException):
```
apyrgio, line 25: When testing for exceptions, it's good to validate the exception message as well, especially for classes like `DocumentFilenameException` that can have multiple messages, to verify that the expected check was triggered. See `match=` in https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions13.

dangerzone/logic.py
```diff
@@ -18,9 +18,9 @@
 log = logging.getLogger(__name__)
 
 
-class GlobalCommon(object):
+class DangerzoneCore(object):
     """
-    The GlobalCommon class is a singleton of shared functionality throughout the app
+    Singleton of shared state / functionality throughout the app
```
apyrgio, line 23: `Dangerzone{Core,Gui}` is basically a collection of common functions and state for:

* Graphics (terminal colors, fonts, window icon)
* Settings (app path, OCR languages)
* Qt app
* Docker status

The fact that it's used for so many different stuff makes it candidate for becoming bloated, and not easily shareable between contexts. I understand that this class (and the common ones before it) act as an easy way to share the above logic between different contexts, but I believe we would greatly benefit if we moved the above logic into their own modules. We would be able to unit test them, and use them from any context, without having to pass around the Dangerzone class.

Note that we definitely need a "context-like" class, which is used to store the state for the conversion operation as a whole, but that's a separate thing. Also, in this PR I don't see this class holding any of that state, which is why I'm still not sure why it's needed. I believe it's used in subsequent PRs, so in this case, it would be more clear if we explain why we want it in the commit message.

apyrgio (changes requested):

apyrgio:
Also, I see the following commits:

```
Common -> Document                                                                             
Document -> DocumentHolder                                                                     
rename (Global)Common & related to "dangerzone"                                                
rename global_common.py -> logic.py
```

Might if I make their names a bit more consistent, such as "Rename X module/class to Y"?

dangerzone/document.py
```diff
@@ -7,9 +7,9 @@
 import appdirs
 
 
-class Document(object):
+class DocumentHolder(object):
```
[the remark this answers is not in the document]
deeplow: 1. is now done (the `Document` -> `DocumentHolder`) commit was deleted (in a parallel branch)
2. this is is better left for the multi-document PR. I tried that as well here, but there are situations in which the Document class isn't set in advance with the document name.

dangerzone/gui/__init__.py
deeplow: I don't think so. I think here the goal was really just to simplify the logic. It can be quite convoluted to understand but from my understanding, before it was checking if there was already a window open without a selected document. If there was, it would not create a window. If there wasn't then it would create a new window.

This check didn't seem important as what is the consequence of the user starting multiple windows without.

dangerzone/cli.py
deeplow: This is done in the child PR #216. [These lines](https://github.com/freedomofpress/dangerzone/pull/216/commits/3be0b852efeaf784eae826a0e17aa92ce6e1c1ad#diff-1818f923cb4b1458022430fcaea85123d0a5d8fd15b270705b5ba7ba085938deL33-L46) specifically.

I'd prefer avoiding further complicating that rebase.

dangerzone/document.py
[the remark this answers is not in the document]
deeplow: > Also, the DocumentFilenameException() class is introduced in 'simplify "new window" logic", and is patched in "add unit tests for document.py". 

Now it's introduced in `validate I/O filenames in Document class` and is no longer patched there.

> I'm not sure about the `DocumentFilenameException` as is,

This a UX thing. Front ends (closer to the user) should present more actionable information. Otherwise, on every front end we'll have to catch the original exception and show the user-understandable translation. But using an intermediary exception
  * Instead of showing the error: `[Errno 2] No such file or directory: document.pdf` which is not user-actionable
  * instead we show: `Input file not found: make sure you typed it correctly.`


> because we lose traceback info and the original message. We could retain these with raise ... from ..., but is it worth the extra complexity?

I have added the `raise ... from ...` syntax. I was not aware this was a thing.

dangerzone/document.py
deeplow: I tried making this a static method, but it will look weird because click's validators have three parameters (`ctx`, `param`, `value`) and not just `filename`. And we'd have to duplicate this validation code because at this moment we click for the cli and another for the gui.

tests/test_document.py
deeplow: Done.

deeplow:
@apyrgio I have addressed all comments in the parallel branch [simplify-common-2](https://github.com/freedomofpress/dangerzone/tree/simplify-common-2)

dangerzone/gui/__init__.py
apyrgio: I see. I didn't realize there could be a case where you have a window open without a document. If it's not important then sure, it can go.

dangerzone/document.py
apyrgio: Great, thanks for the explanation.

dangerzone/document.py
apyrgio: I believe this could be solved via:

* An `args.py` module that is Click-aware,
* Some `validate_{input,output}_filename()` functions, that implement the callback interface for Click (`ctx`, `param`, `value`) and wrap the `Document.validate_{input,output}_filename()` methods.
* Calling the `args.validate_{input,output}_filename()` functions from the CLI and GUI code.

dangerzone/logic.py
apyrgio: Understandable. I'm bringing it up now so that we don't overload this class in subsequent PRs. But we can see that later on.

apyrgio:
@deeplow Thanks a lot for the changes. I'll take a look and report back.

dangerzone/document.py
deeplow: Sounds good. Yes, if you could write it, that would work for me.

dangerzone/cli.py
apyrgio: Got it, thanks.

dangerzone/document.py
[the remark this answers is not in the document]
apyrgio: Cool, thanks for the changes. As for the original traceback, now that we retain it with `raise ... from ...`, it would be nice logging it as well. This requires handling the `DocumentFilenameException` from a single place. I'll give it a try.

apyrgio:
I have done some changes of my own, based on the above discussion, which you can find on the `simplify-common-3` branch. I have tested it on Linux and it works. I need to iron out one small detail, and then it will be good to merge.

deeplow:
Whoops. I have accidentally deleted this while trying to force push.

deeplow:
- Rebased w/ newly included windows tests
- added type hints to `test_documents.py` (because tests now have type hinting requirements)
- fixed `args.py` and `errors.py` type hints

@apyrgio I think the last thing missing is you changing your author email, if you so desire.

deeplow:
Interesting. Mypy is throwing some errors but it doesn't throw them on my system.

````
dangerzone/errors.py:4: error: Module "typing" has no attribute "ParamSpec"; maybe "_ParamSpec"?  [attr-defined]
dangerzone/errors.py:18: error: The first argument to Callable must be a list of types, parameter specification, or "..."  [misc]
dangerzone/errors.py:18: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas
dangerzone/errors.py:22: error: Name "P.args" is not defined  [name-defined]
dangerzone/errors.py:22: error: Name "P.kwargs" is not defined  [name-defined]
```

If we take a look at the docs, it is well documented there: https://mypy.readthedocs.io/en/stable/generics.html?highlight=ParamSpec#declaring-decorators

deeplow:
It looks like this happening because some of CI doesn't have python3.10 version running. We might have moved too soon with https://github.com/freedomofpress/dangerzone/pull/161.

deeplow:
Rebased and ready to be merged. Can I merge it @apyrgio?

apyrgio:
Gimme a bit to take a last look, and running the unit tests on Windows. Also, we need to squash "FIXUP" commits, since they pollute the history.

deeplow:
> Also, we need to squash "FIXUP" commits, since they pollute the history.

Done

apyrgio:
I'm doing some final changes (proper author attribution to some of the commits), and I'm testing on a Windows machine. I'll ping you when you can rebase, sign, and merge.

tests/test_document.py
```diff
@@ -0,0 +1,80 @@
...
+    with pytest.raises(DocumentFilenameException) as e:
+        Document("non-existing-file.pdf")
+    assert "Input file not found" in str(e.value)
+
+
+# XXX: This is not easy to test on Windows, as the file owner can always read it.
+# See also:
+# https://stackoverflow.com/questions/72528318/what-file-permissions-make-a-file-unreadable-by-owner-in-windows
+@pytest.mark.skipif(platform.system() == "Windows", reason="Unix-specific")
+def test_input_file_unreadable(unreadable_pdf):
```
deeplow, line 43: @apyrgio you're only missing here the type hints.

apyrgio: Yeap, I'm on it as we speak.

apyrgio: Done and force pushed.

apyrgio:
I've updated the branch with the following changes:

1. I properly attributed some commits to you. I had split some commits while reviewing and the author became myself, but I changed that back.
2. I fixed a Windows unit test error that I had actually fixed before on `simplify-common-4`. We probably didn't get the latest changes before force-pushing to `simplify-common`. That's ok, but in general it's good to have this in mind.
3. I checked that the Windows unit tests do pass. The only (minor) issue that we have is that `test_lang_eng` fails, because Tesseract times out. Since this test works on Linux, and Tesseract runs on a container, I'll attribute it to our timeouts configuration and my slow Windows VM.

apyrgio (approved):

Issue #214, opened 2022-09-22
Failed CLI execution produces an empty "safe" document


apyrgio:
## Description

Running Dangerzone CLI on a Linux environment that does not have Podman properly configured, throws an error during document conversion. A side effect of this error is that it produces an *empty* `*-safe.pdf` document.

## Steps to Reproduce

OS: Ubuntu 22.04
Release: 0.3.2

* Ensure you have access to the Dangerzone CLI (run `dangerzone-cli --help`).
* Make sure that `podman images ls` fails. You can temporarily move the `/usr/bin/podman` binary to somewhere else, for instance.
* Have an example PDF document ready.
* Run `dangerzone-cli test.pdf` within that container, where `test.pdf` is your example PDF.
* List the contents of your directory. You should see a `test-safe.pdf` there.

## Expected Behavior

Fail the execution without creating a safe document.

PR #216, opened 2022-09-27
Basic via-CLI mutli-document support

deeplow:
**Depends on PR https://github.com/freedomofpress/dangerzone/pull/208**

Adds basic bulk document support to the CLI and makes it so the GUI can be started with multiple files from the terminal.

Usage: `dangerzone-cli [OPTIONS] doc1.pdf doc2.pdf`

  * bulk conversion in CLI implemented via thread (similar to the GUI implementation) and number of threads is bound by the number of CPUs (2*CPU +1)
  *  **Security:** disallow interspersed args (to avoid having filenames be interpreted as options)
  * **Add test cases for bulk document conversion**
  * **Deduplicate arg parsing**
    Make cli and gui use the same arg parser - so that we didn't have to implement twice the multi-filename arg parsing.
  * **GUI basic multi-document support**
    When opening through the terminal (good for testing). Uses the code logic that was already there: opening a new dangerzone window for each document.
   * **Deduplicates `stdout_callback` code** (the one called when a container outputs a `json` line)
   * **Log document id** (based on arg order)
      So we can distinguish the output of each document:
   ```shell_session
        $ dangerzone-cli document-a.pdf document-b.pdf
    
        [doc 1] 3% Separating document into pages
        [doc 2] 0% Converting to PDF using LibreOffice
        [doc 1] 5% Converting page 1/4 to pixels
   ```


dangerzone/cli.py
```diff
@@ -19,76 +18,48 @@ def print_header(s: str) -> None:
     click.echo(Style.BRIGHT + s)
 
 
-@click.command()
-@click.option("--output-filename", help="Default is filename ending with -safe.pdf")
+@click.command(context_settings={"allow_interspersed_args": False})
```
deeplow, line 21: We should handle interspersed args the same way git does: if it detects potential args on the files list, it warns the user:
```
$ git log dangerzone/gui/systray.py -p
fatal: option '-p' must come before non-option arguments
```

deeplow:
The one missing thing is to increase the document conversion timeout proportional to the number of documents. But doing that will be much after all the timeout variables are centralized in one. This is already done with https://github.com/freedomofpress/dangerzone/pull/167. We just have to merge it.

deeplow:
Now that #208 is merged, this is ready for review @apyrgio

dangerzone/logic.py
```diff
@@ -36,3 +38,35 @@ def __init__(self) -> None:
...
+        self.documents: List[Document] = []
+
+    def add_document(
+        self, input_filename: str, output_filename: Optional[str] = None
+    ) -> None:
+        doc = Document(input_filename)
+        if output_filename:
+            doc.output_filename = output_filename
+        else:
+            doc.set_default_output_filename()
```
apyrgio, line 51: I think it's simpler to just:

1. Pass the `Document` instances here. No need for this class to perform the initialization.
2. Change the `Document` class to always have a default for output filenames (`the `self.set_default_output_filename()`).

This will help in the case we add more arguments to `Document.__init__`, or if we need subclasses of `Document` at some point.

dangerzone/cli.py
```diff
@@ -19,30 +18,36 @@ def print_header(s: str) -> None:
     click.echo(Style.BRIGHT + s)
 
 
-@click.command()
+@click.command(context_settings={"allow_interspersed_args": False})
```
apyrgio, line 21: I understand the sentiment behind this change, but I don't get how we can protect a user that does `dangerzone-cli *` from an attacker that has crafted the filenames. The fact that arguments need to be defined before the the input filenames doesn't help much, since `-` as a character has higher precedence than other characters.

Notre that this is a common issue in Unix environments (see some examples in https://www.hackingarticles.in/exploiting-wildcard-for-privilege-escalation/). I'd like to have a solution to this as well, but I'm not sure if there is any :-/.

dangerzone/document.py
```diff
@@ -18,13 +22,26 @@ class Document:
     document, and validating its info.
     """
 
+    doc_counter = 1
```
apyrgio, line 25: Mutable state at class level is usually prone to bugs. It's much better to assign an alphanumerical ID (can be just a few characters) that will uniquely track the document instance.

dangerzone/logic.py
```diff
@@ -36,3 +38,35 @@ def __init__(self) -> None:
...
+        def convert_doc(document: Document) -> None:
+            success = container.convert(
+                document,
+                ocr_lang,
+                stdout_callback,
+            )
+
+        max_jobs = container.get_max_parallel_conversions()
+        with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
+            executor.map(convert_doc, self.documents)
```
apyrgio, line 66: The fact that the blocking calls that this function does are basically two calls to `subprocess`, which can be replaced by their async counterpart (https://docs.python.org/3/library/asyncio-subprocess.html), makes me more confident that we can go down that route in the future

Also, it makes me wanna dip my toes just a tiny bit :grimacing:. Will post an update once I know more.

dangerzone/document.py
```diff
@@ -82,3 +99,27 @@ def set_default_output_filename(self) -> None:
         self.output_filename = (
             f"{os.path.splitext(self.input_filename)[0]}{SAFE_EXTENSION}"
         )
+
+    def is_unsafe(self) -> bool:
+        return self.state is Document.STATE_UNSAFE
```
apyrgio, line 104: I would suggest changing this check to `self.state is not Document.STATE_SAFE`. This way, if we have a document in any other state than strictly `STATE_UNSAFE`, we won't handle it as if it's safe (i.e., double negation does not work here).

apyrgio:
One more suggestion. I think we should update the Changelog, to reflect that we now have support for multiple doc conversion via the CLI.

apyrgio:
@deeplow I just finished my first round of comments. The code looks fine, tests pass, and I gave it a whirl as well. Hit me up with your feedback on the comments, once you have time.

dangerzone/cli.py
deeplow: Good point. Then I shall remove all code related to did. I search a bit about mitgations for this and found none.

So I came up with a rather simple one: checking if any parameter is also a file and aborting if that's the case. I already have a working POC. Let me know what you think of this approach.

apyrgio: Nice and clean, I agree.

dangerzone/document.py
deeplow: But then what do you print in the output? I am printing something like this at the moment:

```
[doc 11] 0% Converting to PDF using LibreOffice
[doc 5] 68% Converting page 3/5 from pixels to PDF
[doc 5] 77% Converting page 4/5 from pixels to PDF
[doc 12] 0% Converting to PDF using LibreOffice
[doc 5] 86% Converting page 5/5 from pixels to PDF
[doc 5] 95% Merging 5 pages into a single PDF
[doc 6] 100% Safe PDF created
```

With alphanumeric values it's going to look nonsensical, no?

dangerzone/document.py
deeplow: I understand the reasoning but we need some way to distinguish unsafe (because unconverted documents) from unsafe (failed to convert) ones. This is important in the GUI multi-document support so we know which image to show:

https://github.com/freedomofpress/dangerzone/blob/c81fde6670193f3c07453cbc87a4760f71f4eb4d/dangerzone/gui/main_window.py#L665-L673

dangerzone/document.py
apyrgio: I guess something like this:

```
[doc b1am3] Assigning ID 'b1ame' to doc C:\Users\dz\doc5.pdf
[doc 0rgs3] Assigning ID 'orgs3' to doc C:\Users\dz\doc6.pdf
[doc 9ein9] Assigning ID '9ein9' to doc C:\Users\dz\doc11.pdf
[doc c1trus] Assigning ID 'c1trus' to doc C:\Users\dz\doc12.pdf
...
[doc 9ein9] 0% Converting to PDF using LibreOffice
[doc b1am3] 68% Converting page 3/5 from pixels to PDF
[doc b1am3] 77% Converting page 4/5 from pixels to PDF
[doc c1trus] 0% Converting to PDF using LibreOffice
[doc b1am3] 86% Converting page 5/5 from pixels to PDF
[doc b1ame] 95% Merging 5 pages into a single PDF
[doc 0rgs3] 100% Safe PDF created
```

_(I'm adding the association between IDs and filenames, which we need to do regardless)_

As for the fact that they are alphanumerical, I think that's a bonus, as you can easily grep for them in the logs.

dangerzone/logic.py
```diff
@@ -36,3 +38,35 @@ def __init__(self) -> None:
...
+        self, ocr_lang: Optional[str], stdout_callback: Optional[Callable] = None
+    ) -> None:
+        def convert_doc(document: Document) -> None:
+            success = container.convert(
+                document,
+                ocr_lang,
+                stdout_callback,
+            )
+
+        max_jobs = container.get_max_parallel_conversions()
```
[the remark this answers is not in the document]
apyrgio: > I don't think we need to expose it at this point in time. I'd prefer a UX review first over which functionalities should be exposed to the user. Otherwise we may end up [like pied piper](https://featureisnotabug.files.wordpress.com/2018/02/interface_large.jpg?w=840). 

Right. We would need some sort of "advanced tab", if we wanted to implement this.

> The OS already should do this kind of management anyways, no?

I think that VMs have higher CPU priority than common processes, which is why the OS is not so great at scheduling those.

> Or maybe better is contextual help. Implement a prompt or some sort of warning that the conversion is consuming too many resources and give the user a prompt to reduce the number of processes with a simple click of a button.

It will be difficult to find an adequate heuristic for that. There are cases where using 100% of the CPU is acceptable, and others when it's not. It's similar to the CPU knob of Docker Desktop. Instinctively, most would prefer assigning all the CPU cores to it, but there are cases where this is not desired.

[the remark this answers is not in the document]
apyrgio: >  But that's for another PR.

Ok, I agree. Worst case scenario, users can limit the CPUs on the Docker Desktop VM, if system performance is suffering.

dangerzone/document.py
apyrgio: This shouldn't be a problem:

```python
 def update_status_image(self) -> None: 
     if self.document.is_safe(): 
         self.status_image.setPixmap(self.img_status_safe) 
     elif self.document.is_converting(): 
         self.status_image.setPixmap(self.img_status_converting) 
     elif self.document.is_failed(): 
         self.status_image.setPixmap(self.img_status_failed)
     elif self.document.is_unsafe(): 
         self.status_image.setPixmap(self.img_status_unsafe)
     else:
         raise Exception("Could not handle document status: ...")
```

If we deem that "unsafe" is a more general status, and safe/converting/failed are more specific ones, then we can simply change the order.

dangerzone/logic.py
apyrgio: ... and dip I did.

I have a POC (`209-bulk-doc-cli-support-async`) where I turned all necessary functions for conversion to async functions, mainly utilizing `asyncio.subprocess` to run Podman. I tested it out, and as far as I can see, conversions happen concurrently. I'm pretty sure there are bugs hiding, especially on Windows/MacOS. Also, I didn't invest time in making pooling work, so there are no limits to the number of concurrent jobs.

I don't know where we go from there though. If we want to seriously consider it, we need to see if it plays nice with your work on the GUI part. Else, it will have to stay on the shelf for a little while.

deeplow:
I spotted an extra issue. We were creating 0-byte length `-safe.pdf` for every converted file. This wasn't an issue before since it was a file at a time, but now it's much more noticeable. I added a fix for this in [01c2eca](https://github.com/freedomofpress/dangerzone/pull/216/commits/01c2eca58453d2800a9fbe382b478a0b87f0c528).

dangerzone/document.py
deeplow: I like the alignment in the log that this will bring and the searchability bonus. Some users might be confused and think that's like a commit hash. But I think it's OK.

apyrgio:
> I spotted an extra issue. We were creating 0-byte length -safe.pdf for every converted file. This wasn't an issue before since it was a file at a time, but now it's much more noticeable. I added a fix for this in [01c2eca](https://github.com/freedomofpress/dangerzone/pull/216/commits/01c2eca58453d2800a9fbe382b478a0b87f0c528).

So, you basically fixed https://github.com/freedomofpress/dangerzone/issues/214? If that's the case, can you close this issue in your commit message?

dangerzone/document.py
apyrgio: Ah, I see. Yes, then `unconverted` makes more sense, and is less easy to abuse it by mistake.

deeplow:
> > 
> 
> So, you basically fixed #214? If that's the case, can you close this issue in your commit message?

Basically. Thanks for the pointer. I added it there and also in the changelog.

dangerzone/document.py
deeplow: OK

deeplow:
I am done with addressing your comments. I am only really missing the async stuff. But I would say if not complete at this time, that can be a separate PR or maybe for a future release. https://github.com/freedomofpress/dangerzone/pull/216#discussion_r1009403650

apyrgio:
Let's start considering the async route once we have the GUI code as well, since this is where the most complexity will appear. Other than that, is there a branch that I can take a look?

deeplow:
> Other than that, is there a branch that I can take a look?

I don't think so. After this I will open the third PR for the GUI and that's it. Just let me know if you're OK with my changes.

deeplow:
When merging on top of #241 we need to be careful to create rename and create a wrapper for `validate_input_filenames()` (notice the plural). Otherwise we run the risk of this not being noticed on an unbutu focal machine.

apyrgio:
Thanks for the fixes, I'm talking a look at them right now.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## Dangerzone 0.4.0
+- Bug fix: Failed execution no longer produces an empty "safe" documents ([issue #214](https://github.com/freedomofpress/dangerzone/issues/214))
+- Feature: support multiple input documents in the CLI-version
```
apyrgio, line 5: It's best if we use the word "unreleased" for these sections, so that users don't get confused. This is already merged in the `main` branch, so it will be fixed once we rebase.

dangerzone/args.py
```diff
@@ -17,9 +18,46 @@ def validate_input_filename(
     return filename
 
 
+def sanitize_parameters(original_param: str) -> str:
+    """
+    Mitigates command injection vulnerabilities. Checks if an argument exists
+    could have originated from a file
```
apyrgio, line 24:
```suggestion
    Mitigates command injection vulnerabilities. Checks if a passed argument
    could have originated from a filename.
```

dangerzone/args.py
```diff
@@ -17,9 +18,46 @@ def validate_input_filename(
...
+
+    Example without mitigation:
+
+        $ dangerzone *
+
+    One file is maliciously named "--ocr-lang" or some other parameter would be
+    interpreted as a parameter and not as a file.
+    """
+    parameter = f"--{original_param}"
```
apyrgio, line 33: I think this doesn't take into account single dash arguments (`-v`), concatenated single dash arguments (`-aux`), or double dash arguments with an equal sign (`--output-filename=sample.pdf`). I think it's best to consult `sys.argv` instead:

1. List the files in the current working directory, and create a `set`.
2. Filter the args in `sys.argv` that start with `-`, and create a `set`.
3. Ensure that there is no intersection between these sets. Else, report the offending arguments.

This can take place after Click has parsed the arguments, we don't need to hook into its internals I believe.

dangerzone/document.py
```diff
@@ -18,13 +24,27 @@ class Document:
...
 
-    def __init__(self, input_filename: str = None) -> None:
+    # document conversion state
+    STATE_UNCONVERTED = enum.auto()
+    STATE_CONVERTING = enum.auto()
+    STATE_SAFE = enum.auto()
+    STATE_FAILED = enum.auto()
+
+    def __init__(self, input_filename: str = None, output_filename: str = None) -> None:
+        self.id = uuid.uuid4()
```
apyrgio, line 34: I was about to suggest using [`secrets.token_urlsafe(6)`](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe), which encodes random values as base64 strings, instead of hex strings which UUID4 does. This would encode more entropy within the same number of characters (36 bits of entropy vs 24 bits of entropy, for 6 characters).

However, I ran it through a simulation (you can play with it [here](https://www.wolframalpha.com/input?i2d=true&i=plot%5C%2840%29%5C%2840%291+-+Power%5Be%2CDivide%5B-k+*+%5C%2840%29k+-+1%5C%2841%29%2C2+*+Power%5B2%2C24%5D%5D%5D%5C%2841%29+*+100%5C%2844%29+0+%3C+k+%3C+1000%5C%2841%29)) based on [this article](https://preshing.com/20110504/hash-collision-probabilities/), and it seems that the chance of collision with 24 bits of entropy is low (0.03% for 100 documents, 3% for 1000 documents).

collision

Of course, with 36 bits of entropy, the chance of collision is virtually non-existent, less than 0.0007% for 1000 documents.

collision2

Anyway, the math say that we don't risk that much with UUID4, so we can go with them, if we prefer them to base64. I leave that to you, as I don't have a strong opinion, and this is not a critical property.

dangerzone/document.py
```diff
@@ -65,10 +79,15 @@ def input_filename(self, filename: str) -> None:
         self.validate_input_filename(filename)
         self._input_filename = filename
 
+        # set the default output filename as soon as we know the input filename
+        self.output_filename = (
+            f"{os.path.splitext(self.input_filename)[0]}{SAFE_EXTENSION}"
+        )
```
apyrgio, line 85: This scares me a bit because if we do `self.input_filename = ...` after `self.output_filename = ...`, we will overwrite the output filename. It's best to have this as part of `def output_filename()` instead:

```python
      @property
      def output_filename(self) -> str:
          if self._output_filename is None:
              if self._input_filename is not None:
                  # Basically repurpose set_default_output_filename()
                  return self._default_output_filename()
              else:
                  raise errors.NotSetOutputFilenameException()
          else:
              return self._output_filename
```

dangerzone/document.py
```diff
@@ -82,3 +101,33 @@ def set_default_output_filename(self) -> None:
...
+
+    def mark_as_safe(self) -> None:
+        log.debug(f"Marking doc {self} as 'safe'")
+        self.state = Document.STATE_SAFE
+
+    def get_short_id(self) -> str:
+        return str(self.id)[0:6]
+
+    def __str__(self) -> str:
+        return self.get_short_id()
```
apyrgio, line 133: I wouldn't suggest overriding `__str__()` to print a random value, as this will make debugging weirder. I would leave `__str__` as is, turn `self.get_short_id()` into a property (`self.short_id`) and use it where necessary.

apyrgio:
@deeplow: Done with the second round of review. I have some minor to moderate review comments, which I can try implementing myself, if you don't have the time. Else, I'll wait for you to respond.

apyrgio:
Actually, I'll just go ahead and implement some of the comments in a separate, and you can cherry-pick what you prefer.

apyrgio:
> Actually, I'll just go ahead and implement some of the comments in a separate branch, and you can cherry-pick what you prefer.

I did some crude fixes in `209-bulk-doc-cli-support-2`. Cherry-pick to your branch whatever your prefer.

CHANGELOG.md
deeplow: Makes sense.

dangerzone/args.py
deeplow: Whoops.

dangerzone/args.py
deeplow: Good catch.

I am struggling with some interesting friction with the `click` module. It's not trivial do to stuff outside of click:
  - **doing the parameter check in the `cli_main`** means `click` already did its own parsing and could have interpreted arguments like `--help`.
  - **doing it wrapping the `cli_main`** (before the click wrappers) makes the whole function a non-click command, which then leads to issues when testing using click's `CliRunner`. For example, when running `pytest tests/test_cli.py::TestSecurity` we get:
    ```python
    [...]
    E       assert 'Security: one file is maliciously named' in ''
    E        +  where '' = <CLIResult AttributeError("'function' object has no attribute 'main'")>.stdout
    ```

Furthermore I found that we could use a click function to detect parameters from `sys.argv`. A simple function but not worth reinventing the wheel if we are already depending on click. The function is `prefix, parameter = click.parser.split_opt(arg)`.

So it is a catch-22 situation. I have a working code to do what you proposed, but I don't have a way execute it before click without breaking the tests. If I do it like I was, with `token_normalize_func` trick the code run over every argument. So getting it from `sys.argv` every time doesn't make sense.

Way too much time spent on this already. I will put a pause on this one for now.

dangerzone/document.py
deeplow: I had assumed such a thing could happen (though I didn't looking into the probability) and I kept the full UUID as the full ID, but I guess from the user's perspective all they will see is the short version. So, for practical reasons we should avoid the collision there. So I have now done as you suggested.

dangerzone/args.py
apyrgio: Yeah, I get your pain. I hit against the same wall, and I had to override a public Click method to achieve CLI and test parity. It's in the `209-bulk-doc-cli-support-2` branch.

tests/test_cli.py
```diff
@@ -120,7 +122,17 @@ def run_cli(self, args: Sequence[str] | str = ()) -> CLIResult:
...
+        with runner.isolated_filesystem(temp_dir=tmp_path) as new_path:
+            if tmp_path:
+                # isolated_filesystem() creates a subdirectory (new_path)
+                # this "merges" the two directories so that helper files
+                # are in the same dir as the one where the cli is called
+                for f in os.listdir(tmp_path):
+                    src_path = os.path.join(tmp_path, f)
+                    dst_path = os.path.join(new_path, f)
+                    os.symlink(src_path, dst_path)
+            result = runner.invoke(cli_main, args)
```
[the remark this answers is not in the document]
deeplow: No, you're not missing anything. I was the one over-complicating things.

dangerzone/args.py
deeplow: I see now that you've found a solution to this https://github.com/freedomofpress/dangerzone/commit/44e32bd239778e3461d0f5f3a265a21f22242190#diff-1818f923cb4b1458022430fcaea85123d0a5d8fd15b270705b5ba7ba085938deR89

deeplow:
Pushed your changes here from the `209-bulk-doc-cli-support-2` branch. Everything is mostly fine and I just wanted to add a small comment or two.

tests/test_cli.py
```diff
@@ -197,3 +213,60 @@ def test_filenames(self, filename: str, tmp_path: Path) -> None:
...
+    def test_file_pretending_as_parameter(self, tmp_path: Path) -> None:
+        """
+        Protection against "dangeronze-cli *" where a file is maliciously named
+        as an argument.
+        """
+
+        msg = "Security: Detected CLI options that are also"
+
+        @contextlib.contextmanager
+        def temp_file(file):
```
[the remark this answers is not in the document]
apyrgio: The idea here is that we create some files in a directory, we run the CLI command, and we expect it to fail with a "Security: ..." message. There are various cases that we want to cover, so we do this multiple times.

If we don't remove the temp files before the next `run_cli()` invocation, the next command will fail, but we can't be sure if the files from the previous invocation triggered the failure.

I agree that the context manager is a bit heavy on the eyes, so we could just create a subdir for each `run_cli()` invocation instead.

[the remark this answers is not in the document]
apyrgio: I'll push a fixup commit for that, so you can evaluate it.

dangerzone/cli.py
```diff
@@ -63,30 +69,42 @@ def cli_main(
...
+    _parse_fn = cli_main.parse_args
+    check_fn = args.check_suspicious_options
+
+    @functools.wraps(_parse_fn)
+    def custom_parse_args(ctx, args) -> List[str]:
+        check_fn(args)
+        return _parse_fn(ctx, args)
+
+    cli_main.parse_args = custom_parse_args
+
```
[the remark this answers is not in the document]
apyrgio: Sure thing. I'll send a fixup commit for that as well.

tests/test_cli.py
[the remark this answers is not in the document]
deeplow: How about splitting these these into separate tests?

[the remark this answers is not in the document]
deeplow: Or maybe a pytest fixture that returns a pair: a tempdir and a file within it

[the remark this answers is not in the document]
apyrgio: It would add a bit more boilerplate stuff, but I don't mind much.

dangerzone/cli.py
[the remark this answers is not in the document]
apyrgio: Just did.

tests/test_cli.py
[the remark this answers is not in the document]
apyrgio: I sent a fixup for that as well.

deeplow:
All seems ready. Shall I merge it?

apyrgio:
Sure, go ahead!

deeplow:
Rebased and squashed FIXUP commits. Merging now with main.

PR #235, opened 2022-10-20
Fix Windows unit tests

### LINKED ISSUE #234: Fix unit tests on Windows
It seems that unit tests fail on the `main` branch on Windows:

windows_pytest_main

The  underlying cause seems to be unit test -related, and doesn't affect the actual Dangerzone program. Still, we should fix them as they impact the development on Windows.

apyrgio:
Fix some failing Windows unit tests. Also, add extra functionality that will make debugging CLI tests easier:
* Capture the provided arguments to a command.
* Capture the command's output and exception, and print it in case of an assertion error.

Closes #234

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
+import copy
 import os
 import shutil
+import sys
 import tempfile
+import traceback
+from typing import Sequence
 
 import pytest
 from click.testing import CliRunner, Result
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -21,24 +24,113 @@
 # TODO simulate container runtime missing
 # TODO simulate container connection error
 # TODO simulate container connection loss
-# FIXME "/" path separator is platform-dependent, use pathlib instead
+
+
+class CLIResult:
```
deeplow, line 29: We may want to check if something is in the output of the run command with `Result.stdout` or some other `Result` method. The way `CliResult` was is implemented here it wraps around the `click.Result` object but prevented calling its methods. I'd propose making this a subclass of Result.  I have a proposal at `origin/234-windows-unit-tests-2`. It looks a bit hacky but it works.

deeplow: This approach has the extra advantage of not having to call `self.click_result` but instead `self`.

deeplow (changes requested):
Generally I don't think we need to have lint checking in the tests, but if you want to add it we should also add it to the CI so these things don't go unnoticed.

I also have a proposal to make `CliResult` a subclass of `click.Result` for its methods callable directly. Other than that, all is good.

apyrgio:
Ah, and here I was, thinking that I got the MyPy lints right from the start. I'll fix them, thanks.

tests/test_cli.py
apyrgio: Ok, that's interesting. The way to go in these cases is usually to change the `__class__` attribute, instead of copying `__dict__`. There are various reasons for this; you need to initialize the new class, and you lose the previous arguments you had set (here we lose `downcast_result.args`, check it out).

Originally, I assumed that the caller would do `result.click_result.<method>`, instead of `result.<method>`. It's a bit more verbose, but less magic as well. It's tests though, we can have magic code there, so we can go with your suggestion.

apyrgio:
I pushed two commits on top of the `234-windows-unit-tests-2` branch. First commits improves the reclassing of `Result -> CLIResult`, and the second commit makes our tests pass the Mypy checks. Give 'em a look and tell me your opinion.

deeplow (approved):

deeplow:
Feel free to merge it @apyrgio.

PR #237, opened 2022-10-27
Bump changelog

apyrgio:


apyrgio:
I just noticed something interesting. Once we bump the changelog, it's probably good to bump the install notes as well (https://github.com/freedomofpress/dangerzone/wiki/Installing-Dangerzone), since the releases should show up there as well.

PR #248, opened 2022-11-10
WIP: Implement Linux User Namespaces support

### LINKED ISSUE #221: Defense in Depth
# Primer

Dangerzone currently uses two containers (Docker containers on MacOS/Windows, Podman containers on Linux) for the conversion process:

- The first container accepts a suspicious document/image as input, and produces a list of RGB files (corresponding to the pages of the document).
- The second container accepts the list of RGB files, and produces the safe PDF document.

On this issue, we'll focus on protecting the users against an attacker who tries to take control of the first container.

# Requirements

From the point where a conversion process reads an attacker payload, to the point where the attacker manages to escape to the host, there are several hurdles that they must jump. In order of importance, from the innermost hurdle to the outermost, these are the requirements from Dangerzone, to keep a user safe.

1. The attacker must not gain control of a process within the container.
2. The attacker must not access host data from within the container.   
3. The attacker must not access the network from within the container.           
4. The attacker must not become root within the container.                       
5. The attacker must not escape the container.                                                  
6. The attacker must not access host data outside the container.
7. The attacker must not become user/root outside the container.
8. The attacker must not escape the VM.

# Current situation

The protective measures that Dangerzone has in place are:

* (MacOS/Windows only) Containers run in VMs.
   - This is necessary, as this the only way that Docker Desktop can run Linux containers in the rest of the OSes.
* The conversion process in the container is not / cannot become root.
  - The Dangerzone container image has a regular user in it, and the conversion process runs with their UID.
  - No process can elevate themselves to root, since no new privileges can be acquired.
  - See the relevant discussion: https://github.com/freedomofpress/dangerzone/issues/169
* Containers have no capabilities (see [`capabilities(7)`](https://man7.org/linux/man-pages/man7/capabilities.7.html)).
  - We drop all capabilities when we create a new container, as we don't need them.
  - See the relevant PR: https://github.com/freedomofpress/dangerzone/pull/183
* Containers have no access to the network.

The question here is: can we improve on this situation even more?

# Previous issues

* https://github.com/freedomofpress/dangerzone/issues/52
* https://github.com/freedomofpress/dangerzone/issues/169

apyrgio:
Implement support for Linux User Namespaces in Dangerzone, as described in #221.

Issue #221 has moved to the 0.5.0 milestone, but since it's a complex subject matter, I wanted to push the work I have done so far on this subject, before switching to other things that have higher priority. The current state is that the proposed implementation steps seem to work, and the tests pass. The code needs a bit more polishing, but other than that, it should be simple to review and merge.

Fixes #221

### DIFF
--- a/dangerzone/container.py
+++ b/dangerzone/container.py
@@ -1,4 +1,5 @@
 import gzip
+import json
 import logging
 import os
 import pipes
@@ -47,6 +48,19 @@ def get_runtime() -> str:
     if runtime is None:
         raise NoContainerTechException(container_tech)
     return runtime
+
+
+def podman_get_subids():
+    info = subprocess.run(
+        ["podman", "info", "-f", "json"], check=True, stdout=subprocess.PIPE
+    )
+    info_dict = json.loads(info.stdout)
+    for id_type in ["uid", "gid"]:
+        mapping = info_dict["host"]["idMappings"][f"{id_type}map"]
+        count = 0
+        for m in mapping[1:]:
+            count += m["size"]
+        yield count
 
 
 def install() -> bool:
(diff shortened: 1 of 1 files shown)


deeplow (commented):
Thanks for working on this! It took me a while to understand how this works, but I think I got it and it makes sense.

This is already better than we had. However, if we run multiple containers at the same time, aren't we allocating the same namespace to all of them? [this can go on a future PR and it isn't as concerning as host-container separation. [This blog post](https://projectatomic.io/blog/2018/05/podman-userns/) had a great summary.

apyrgio:
This PR has been superceded by the gVisor integration PR (#590), which puts the Linux user that runs the rasterization logic of Dangerzone into a different user namespace. This approach was Linux-only, and hence more limited in scope, so we can safely discard it.

Issue #251, opened 2022-11-16
Option to move untrusted files into subdirectory after conversion


deeplow:
The way I implemented the bulk doc conversion (loosely following [Nina's proposal](https://github.com/freedomofpress/dangerzone/issues/117#issuecomment-861691174)), by default it leaves all the -safe.pdf files in the same directory. This can lead the user to mistakenly opening an untrusted file.

My proposal would be to change the behavior to make it a bit more like the Qubes implementation, where it leaves the trusted files in the same directory, but creating a subdirectory for untrusted files. I have summed these observations and their consequences in an image below.

![untitled_diagram drawio_3_](https://user-images.githubusercontent.com/47065258/202153032-e370165a-4506-485d-a91b-0ded200184d2.png)

PR #255, opened 2022-11-21
Option to move untrusted files into subdirectory after conversion 

deeplow:
Follows what's been outlined in #251:
opt3

The only thing missing would be a way to detect a case where an archive directory cannot be created where the original documents are located. This could show in the GUI by disabling the option to archive, or the CLI by giving out an error.

### DIFF
--- a/dangerzone/document.py
+++ b/dangerzone/document.py
@@ -13,6 +13,7 @@ import appdirs
 from . import errors
 
 SAFE_EXTENSION = "-safe.pdf"
+ARCHIVE_SUBDIR = "unsafe"
 
 log = logging.getLogger(__name__)
 
@@ -35,12 +36,14 @@ class Document:
         input_filename: str = None,
         output_filename: str = None,
         suffix: str = SAFE_EXTENSION,
+        archive: bool = False,
     ) -> None:
         # NOTE: See https://github.com/freedomofpress/dangerzone/pull/216#discussion_r1015449418
         self.id = secrets.token_urlsafe(6)[0:6]
 
         self._input_filename: Optional[str] = None
         self._output_filename: Optional[str] = None
+        self._archive = False
         self._suffix = suffix
 
         if input_filename:
@@ -50,6 +53,8 @@ class Document:
                 self.output_filename = output_filename
 
         self.state = Document.STATE_UNCONVERTED
+
+        self.archive_after_conversion = archive
 
     @staticmethod
     def normalize_filename(filename: str) -> str:
@@ -112,6 +117,33 @@ class Document:
             self._suffix = suf
         else:
             raise errors.SuffixNotApplicableException()
+
+    @property
+    def archive_after_conversion(self) -> bool:
+        return self._archive
+
+    @archive_after_conversion.setter
+    def archive_after_conversion(self, enabled: bool) -> None:
+        if enabled:
+            self._archive = True
+        else:
+            self._archive = False
+
+    def archive(self) -> None:
+        """
+        Moves the original document to a subdirectory. Prevents the user from
+        mistakenly opening the unsafe (original) document.
+        """
+        archive_dir = self.default_archive_dir
+        old_file_path = Path(self.input_filename)
+        new_file_path = archive_dir / old_file_path.name
+        log.debug(f"Archiving doc {self.id} to {new_file_path}")
+        Path.mkdir(archive_dir, exist_ok=True)
+        old_file_path.rename(new_file_path)
+
+    @property
+    def default_archive_dir(self) -> Path:
+        return Path(self.input_filename).parent / ARCHIVE_SUBDIR
 
     @property
     def default_output_filename(self) -> str:
--- a/dangerzone/document.py
+++ b/dangerzone/document.py
@@ -77,6 +77,11 @@ class Document:
             # in unwriteable directory
             raise errors.UnwriteableOutputDirException()
 
+    def validate_default_archive_dir(self) -> None:
+        """Checks if archive dir can be created"""
+        if not os.access(self.default_archive_dir.parent, os.W_OK):
+            raise errors.UnwriteableArchiveDirException()
+
     @property
     def input_filename(self) -> str:
         if self._input_filename is None:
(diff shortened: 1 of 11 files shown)


deeplow:
Dangerzone has refered to untrusted files in the past as suspicious. Should we name the directory and the option as suspicious?

apyrgio:
On the other hand, the non-suspicious documents are suffixed as "-safe", so having an "unsafe" directory makes the difference clearer. I'd vote to keep it as "unsafe" personally.

dangerzone/document.py
```diff
@@ -72,6 +78,11 @@ def validate_output_filename(filename: str) -> None:
             # in unwriteable directory
             raise errors.UnwriteableOutputDirException()
 
+    def validate_default_archive_dir(self) -> None:
+        """Checks if archive dir can be created"""
+        if not os.access(self.default_archive_dir.parent, os.W_OK):
+            raise errors.UnwriteableArchiveDirException()
```
apyrgio, line 84: There's something that worries me here. I tried to test this scenario out with the GUI version of Dangerzone. I created a directory, copy-pasted a document there, and then made the directory read-only.

After that, clicking on "Convert to Safe Document" did not do anything. In the terminal (since I started it via `dev_scripts/dangerzone`), I could see the `UnwriteableArchiveDirException` that is logged, but I'm afraid that the user won't see it. 

Triggered by that, I tried to see what would happen if we attempted to save a document in an output directory that is also not writable. In that case, we rely also on Python exceptions. What I saw is that the operation succeeded, but:

* the document was never written to the output directory (expected)
* no error was reported, either in the UI or the terminal logs (unexpected)

which is false, and the operation shouldn't start in the first place.

dangerzone/document.py
```diff
@@ -157,3 +188,6 @@ def mark_as_failed(self) -> None:
     def mark_as_safe(self) -> None:
         log.debug(f"Marking doc {self.id} as 'safe'")
         self.state = Document.STATE_SAFE
+
+        if self._archive:
+            self.archive()
```
apyrgio, line 193: I don't think `mark_as_safe()` should have side-effects. Like the rest of the `mark_as_*` functions, it should just update the status of the document.

I would suggest moving this to a separate method, maybe in the same place where we have the logic that copies the final document to the output directory.

dangerzone/gui/main_window.py
```diff
@@ -501,17 +526,24 @@ def start_button_clicked(self) -> None:
...
+                    document.suffix = self.safe_extension.text()
 
-                if self.output_dir:
-                    document.set_output_dir(self.output_dir)  # type: ignore [unreachable]
+            elif self.radio_save_to.isChecked():
+                for document in self.dangerzone.get_unconverted_documents():
+                    document.suffix = self.safe_extension.text()
+
+                    if self.output_dir:
+                        document.set_output_dir(self.output_dir)  # type: ignore [unreachable]
```
apyrgio, line 534: There are two things that bug me here:

1. Why is this operation a "this or that" thing. Can't we both archive the original document, and move the safe document to another directory? I don't get what's the conflict here.
2. There are some copy-pasted for-loops and function calls, so I've pushed a proposal that attempts to eliminate them.

dangerzone/document.py
deeplow: Oh, right. We're missing the GUI detection for this. I will do that now.

dangerzone/gui/main_window.py
deeplow: > There are some copy-pasted for-loops and function calls, so I've pushed a proposal that attempts to eliminate them.

Yes. There were some annoying duplicated code. I opted for having non-optimal iterations over all documents and duplicating the code. But your approach is better in the long run probably.

>  Why is this operation a "this or that" thing. Can't we both archive the original document, and move the safe document to another directory? I don't get what's the conflict here.

It doesn't have to be mutually exclusive. But it's two different ways of using the program. What would be the case if you archive all documents under the "unsafe" directory and at the same time move the safe files elsewhere? 

If there is a strong case for using it that way, then it makes sense. Otherwise, I'd push for a simpler usage, given that we're already increasing the UX (cognitive) complexity. Having 4 options instead of 2 might makes it more complex to the user. Maybe I'm exaggerating, but overall my point is the more use-case driven the UX is, the better (that won't make power users happy, though).

dangerzone/document.py
deeplow: Done in 0ab5996

apyrgio: Just took a look. Two things:

1. I see that the tooltip appears on hover everywhere on the window. Is this intentional?
2. Does this solve the problem of an unwriteable output directory for the safe files? Granted, this existed before, but I'd like to know.

dangerzone/gui/main_window.py
apyrgio: Ok, makes sense then.

dangerzone/document.py
deeplow: 1.  Fixed in https://github.com/freedomofpress/dangerzone/pull/255/commits/fe0cd2771a278e331515a1d158707e338993d2f9
2. Good point. It doesn't and that will be problematic. Apparently it fails silently, saying it succeeded but didn't actually save the -safe.pdf file. It's probably a problem that already existed in previous release. We'd need to add logic to tell the user to change directory. Should be easy, should I do that? Or drop both?

dangerzone/document.py
deeplow: That's a much better place. Thank you. Done in https://github.com/freedomofpress/dangerzone/pull/255/commits/a8fe7afdb652c99de883574536699864580d5712

dangerzone/document.py
apyrgio: 1. Nice, it works.
2. Hm, maybe I can give it a shot in a separate PR.

dangerzone/document.py
apyrgio: Yay, I like it!

dangerzone/document.py
deeplow: Not sure if you did that, but I have already some working code. Did you end up working on it?

apyrgio: Nope, I didn't. If you have some working code, then shoot, but I'd suggest in a different PR.

deeplow:
Had to make some final layout adjustments.

PR #256, opened 2022-11-23
Inform user # of selected docs when in settings

deeplow:
Reminds the users of the number of documents selected when they are in the settings.

docs_selection_new

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -296,6 +296,12 @@ class SettingsWidget(QtWidgets.QWidget):
         super(SettingsWidget, self).__init__()
         self.dangerzone = dangerzone
 
+        # Num Docs Selected
+        self.docs_selected_label = QtWidgets.QLabel("No documents selected")
+        self.docs_selected_label.setAlignment(QtCore.Qt.AlignCenter)
+        self.docs_selected_label.setContentsMargins(0, 0, 0, 20)
+        self.docs_selected_label.setProperty("class", "docs-selection")  # type: ignore [arg-type]
+
         # Save safe version
         self.save_checkbox = QtWidgets.QCheckBox("Save safe PDF to")
         self.save_checkbox.clicked.connect(self.update_ui)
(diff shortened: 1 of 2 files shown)


dangerzone/gui/main_window.py
```diff
@@ -296,6 +296,12 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
         super(SettingsWidget, self).__init__()
         self.dangerzone = dangerzone
 
+        # Num Docs Selected
+        self.docs_selected_label = QtWidgets.QLabel("No documents selected")
+        self.docs_selected_label.setAlignment(QtCore.Qt.AlignCenter)
+        self.docs_selected_label.setContentsMargins(0, 0, 0, 20)
+        self.docs_selected_label.setProperty(b"class", "docs-selection")
```
apyrgio, line 303: I see something weird here. If I run it as is, I get the following errors:

```
FIXME typing.Any cannot be used with isinstance()
...
TypeError: 'PySide2.QtCore.QObject.setProperty' called with wrong argument types:
  PySide2.QtCore.QObject.setProperty(bytes, str)
Supported signatures:
  PySide2.QtCore.QObject.setProperty(bytes, typing.Any)
QThread: Destroyed while thread is still running
Aborted (core dumped)
```

If I replace `b"class"` with `"class"`, then I don't get any error. Any particular reason you opted to use bytes there? I'm using Python 3.10.6, if it helps.

deeplow: Oh. That was a typo. I must have changed to `b"class"` last minute to see if it would fix the typing, but I forgot about it. 

It will have to be string and with `# type: ignore` comment. We've done it somwhere else in the code. I couldn't find a way to satisfy the linter.

deeplow: I added a FIXUP commit to address this.

deeplow:
Should we make it bold to make it consistent with the `0.3.2` release -- where it said "Suspicious: file-blah-blah.doc`?

deeplow:
Squashed FIXUP commits and rebased from main.

Issue #257, opened 2022-11-23
Support parallel conversions


deeplow:
Parallel conversion is currently implemented [for the CLI](https://github.com/freedomofpress/dangerzone/blob/aba699/dangerzone/logic.py#L64) and [the graphical](https://github.com/freedomofpress/dangerzone/blob/aba699a2388f83e13723b55a1dbe18fb385842ef/dangerzone/gui/main_window.py#L592) versions of Dangerzone.

However, it increased increased the likelihood of timeout. We should figure out a better strategy to handle timeouts and only then allow parallel document conversions.

Kol9yN:
here would be a good option for set how many cpu cores and memory program will consume

apyrgio:
That would be a nice knob for advanced users.

Quick note on the status of this feature: parallel conversions can be tricky in case of OCR, since it can use up all of the host's CPU. Ideally, we need a processor pool that will be used both for the doc-to-pixel stage, and the pixel-to-doc stage, while taking into account that these jobs are dissimilar.

Kol9yN:
what about use more containers then one for convert?
So DZ can convert more files (torrent like) at same time

apyrgio:
Oh yeah, we already have a different container per document. But as I said, it's doable, just not in our top priorities. Especially if you use OCR, Dangerzone 0.11.0 will probably saturate half of your CPUs, so the parallel conversion may not be that much needed. Let us know once it's out!

PR #258, opened 2022-11-23
Disable parallel conversions

deeplow:
Temporarily limit conversions to one at a time until timeout limitations
are resolved: https://github.com/freedomofpress/dangerzone/issues/257


apyrgio:
I would add a *FIXME* in that comment of yours. Otherwise, feel free to merge.

PR #269, opened 2022-11-25
Increase App Size due to trimmed widgets on macOS

deeplow:
The save group box would get partially trimmed when running in macOS this appears to be due to differences in rendering fonts and spacings. **Should this warrant setting different window heights for different platforms?**

> **Note:** I tried expanding the `save_group_box` `minimumHeight` and then adding some spacing or a stretch. But in all situations, the group box would always be on top of other widgets. In the end the solution that worked was increasing the window's height.

### macOS

|Screenshot 2022-11-25 at 10 52 38  |Screenshot 2022-11-25 at 12 49 33    |
|--|--|

### Linux
Side-effects on Linux include a greater space between the "convert" button and the bottom of the window.
| dz-before |after | 
|--|--|

### Windows

| image | image(2) |
|--|--|

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -29,7 +29,12 @@ class MainWindow(QtWidgets.QMainWindow):
         self.setWindowIcon(self.dangerzone.get_window_icon())
 
         self.setMinimumWidth(600)
-        self.setMinimumHeight(430)
+        if platform.system() == "Darwin":
+            # FIXME have a different height for macOS due to font-size inconsistencies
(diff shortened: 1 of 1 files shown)


deeplow:
Related: https://stackoverflow.com/questions/17320263/qt-fonts-have-different-sizes-on-different-systems

deeplow:
@apyrgio this is ready.

apyrgio:
Looks good to me. Might be worth having a "Refs #270" in the commit message as well, to link the PR with the issue. Other than that, feel free to merge.

PR #272, opened 2022-11-29
Fix 'open with Dangerzone' on macOS

### LINKED ISSUE #268: Unable to "Open with" on macOS
If we right-click a file in finder and choose <kbd>Open With Dangerzone</kbd>. This feature stopped working with the 0.3.2 release.

Screenshot 2022-11-25 at 12 15 08(1)

deeplow:
(Fixes #268) This PR adds again the support for 'open with Dangerzone', specifically on macOS (may work on other systems). 

Without the 'open with Dangerzone', the user could only really add documents once. But with this functionality, it opens up many other possibilities. So this PR handing extra edge-cases.

### 1. Prevent adding duplicate documents

image(4)

### 2. Prevent adding documents after the conversion has started
image(5)

### 3. Update settings page after more documents are added
Scenario:
1. User starts Dangerzone with one document
2. [Dangerzone open in settings saying '1 document selected']
3. User right-clicks another file and 'open with... Dangerzone'
4. [Dangerzone adds the doc and settings now show '2 documents selected']

### Adding files from different directories
With the 'open with Dangerzone', it would now possible for someone to add files from multiple directories. This would lead to complexity in regards to the options of where the file is saved. To make things simple, I have disabled this option for now in the GUI:

image(6)

### DIFF
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -19,7 +19,7 @@ from .main_window import MainWindow
 
 
 class Application(QtWidgets.QApplication):
-    document_selected = QtCore.Signal(str)
+    document_selected = QtCore.Signal(list)
     application_activated = QtCore.Signal()
 
     def __init__(self) -> None:
@@ -36,7 +36,7 @@ class Application(QtWidgets.QApplication):
             if isinstance(event, QtGui.QFileOpenEvent):
                 # Skip file open events in dev mode
                 if not hasattr(sys, "dangerzone_dev"):
-                    self.document_selected.emit(event.file())
+                    self.document_selected.emit([event.file()])
                     return True
             elif event.type() == QtCore.QEvent.ApplicationActivate:
                 self.application_activated.emit()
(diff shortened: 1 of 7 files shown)


dangerzone/gui/__init__.py
```diff
@@ -46,6 +54,26 @@ def monkeypatch_event(arg__1: QtCore.QEvent) -> bool:
 
         self.event = monkeypatch_event  # type: ignore [assignment]
 
+    def accumulate_file_open(self, filename: str) -> None:
```
apyrgio, line 57: I'm not sure I understand why accumulating these events is necessary, if the application can handle adding extra documents at any time. Can you elaborate a bit?

apyrgio:
I think there's a typo in this commit message, but I'm not sure what the end result should be.

"Make it so adding the process of adding documents converges"

dangerzone/gui/main_window.py
```diff
@@ -237,11 +236,53 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
...
+                self.dangerzone.output_dir = output_dir
+            elif self.dangerzone.output_dir != output_dir:
+                Alert(
+                    self.dangerzone,
+                    message="Dangerzone does not support adding documents from multiple locations.\n\n The newly added documents were ignored.",
+                    has_cancel=False,
+                ).exec_()
+                return
+            else:
+                self.dangerzone.output_dir = output_dir
```
apyrgio, line 255: During my tests, I saw that users can add documents from various directories with:

```
$ dangerzone dir1/doc.pdf dir2/doc.pdf
$ dangerzone dir1/doc.pdf dir2/doc.pdf
```

With the GUI, I can also save the files to a third directory. So, what do we do there? Should we have parity across platforms and catch this case as well, in the GUI and CLI, or let it be an allowed, but undefined behavior.

Personally, I'd lean on taking the less complicated approach, as we're nearing to the 0.4.0 release. But we can discuss.

PR #277, opened 2022-12-01
Update the QA section in RELEASE.md

apyrgio:
Update the QA section in the RELEASE.md, based on the latest changes on our main branch.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -85,22 +85,27 @@ showing the last error message.
 #### 6. Dangerzone succeeds in converting multiple documents
 
 Run Dangerzone against a list of documents, and tick all options. Ensure that:
-* Conversions start in parallel.
+* Conversions take place sequentially.
+* Attempting to close the window while converting asks the user if they want to
+  abort the conversions.
 * Conversions are completed successfully.
 * Conversions show individual progress.
 * _(Only for Linux)_ The resulting files open with the PDF viewer of our choice.
 * OCR seems to have detected characters in the PDF files.
 * The resulting files have been saved with the proper suffix, in the proper
   location.
+* The original files have been saved in the `unsafe/` directory.
 
 #### 7. Dangerzone CLI succeeds in converting multiple documents
 
-Run Dangerzone CLI against a list of documents. Ensure that conversions start in
-parallel, are completed successfully, and we see their progress.
+_(Only for Windows and Linux)_
+
+Run Dangerzone CLI against a list of documents. Ensure that conversions happen
+sequentially, are completed successfully, and we see their progress.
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -9,6 +9,8 @@ and newer platforms, we have to do the following:
 
 - [ ] In `.circleci/config.yml`, add new platforms and remove obsolete platforms
 - [ ] Make sure that the tip of the `main` branch passes the CI tests.
+- [ ] Ensure that README.md reflects the current state of the project
+  (description and screenshots)
```
apyrgio, line 13: We actually cover this in the same file, but in a different section: https://github.com/freedomofpress/dangerzone/blob/49b7736cb49d8b03d44c66a3903ba4b9ba1f9171/RELEASE.md#changelog-version-and-signed-git-tag

So, I suggest we drop this line.

deeplow: Actually, I think it would make more sense to keep it in it's original location. We don't know if during the QA procedure we will find visual issue or just need to do some other change that has visual consequences. Doing it in the end might mitigate that risk.

RELEASE.md
```diff
@@ -85,18 +87,19 @@ showing the last error message.
 #### 6. Dangerzone succeeds in converting multiple documents
 
 Run Dangerzone against a list of documents, and tick all options. Ensure that:
-* Conversions start in parallel.
+* Conversions take place sequentially.
 * Conversions are completed successfully.
 * Conversions show individual progress.
 * _(Only for Linux)_ The resulting files open with the PDF viewer of our choice.
```
deeplow, line 93: I suggest adding the an extra line:
```diff
+* Closing window while converting asks the user if they want to abort the conversions.
```

RELEASE.md
```diff
@@ -85,18 +87,19 @@ showing the last error message.
...
 * OCR seems to have detected characters in the PDF files.
 * The resulting files have been saved with the proper suffix, in the proper
   location.
+* The original files have been saved in the `unsafe/` directory.
 
 #### 7. Dangerzone CLI succeeds in converting multiple documents
```
deeplow, line 100:
```diff
+_(Except for MacOS)_
+
```

Cli isn't available on macOS.

RELEASE.md
```diff
@@ -85,18 +87,19 @@ showing the last error message.
...
 
-Run Dangerzone CLI against a list of documents. Ensure that conversions start in
-parallel, are completed successfully, and we see their progress.
+Run Dangerzone CLI against a list of documents. Ensure that conversions happen
+sequentially, are completed successfully, and we see their progress.
 
 #### 8. Dangerzone can open a document for conversion via right-click -> "Open With"
```
deeplow, line 105: 8. can also be tested on windows.

PR #278, opened 2022-12-01
Fix 0.4.0 windows builder

deeplow:
The windows builder was not working. This fixes it.

In the future we can make more permanent fixes like changing the build [directory name](https://cx-freeze.readthedocs.io/en/latest/setup_script.html#cmdoption-arg-build_exe) to something independent of the python version.


apyrgio:
Looks good. One minor nitpick is changing "Cx-Freeze 6.12.0" to "Cx-Freeze 6.13.0" in the last commit message.

PR #279, opened 2022-12-01
Various fixes for Windows 0.4.0

deeplow:
Fixes two things that were misbehaving while doing QA on windows for Dangerzone 0.4.0
-  replace `exit()` with `sys.exit()` to work on Windows
-  save / load 'open after conversion' setting

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -95,9 +95,9 @@ def cli_main(
         print_header("Failed to convert document(s)")
         for document in documents_failed:
             click.echo(document.input_filename)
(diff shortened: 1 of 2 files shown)


apyrgio:
Code-wise looks good, and I did a quick test on Fedora as well. I only have a suggestion on the first commit message:

> windows: Persist "Open safe documents after converting" setting 
>     
> Now that safe PDFs can open on Windows right after conversion
> (implemented in commit 5b2fefd15), we need to save/load the
> "Open safe documents after converting" setting.

PR #288, opened 2022-12-12
Fix test instability: pytest in seq. podman<4.3.0

### LINKED ISSUE #217: tests fail non-deterministically (Error: error retrieving size of image)
See example in: https://app.circleci.com/pipelines/github/freedomofpress/dangerzone/443

This happens sometimes when I run `make test`. It only started happening when I started doing tests in parallel (`pytest -n 4`).

Podman with the following:

```
E       AssertionError: assert 1 == 0
E        +  where 1 = <Result CalledProcessError(125, ['/usr/bin/podman', 'image', 'list', '--format', '{{.ID}}', 'dangerzone.rocks/dangerzone'])>.exit_code

tests/test_cli.py:65: AssertionError
----------------------------- Captured stderr call -----------------------------
Error: error retrieving size of image "a19b6c87d9a7e435b0ee80a86453ea9d9adbb211d151404db8baffe43dd531ea": you may need to remove the image to resolve the error: size/digest of layer with ID "652e014cf40878b225ca19dec247b7b1f82aea56f3cfb63a98058fd0dd8b1741" could not be calculated: lgetxattr /home/circleci/.local/share/containers/storage/overlay/d54f9e57784c8d6f6c3e15926b3460e5cbb24bdcd527c4ac13310f05e3dc6c9e/merged/usr/share/mime/application/vnd.coffeescript.xml: no such file or directory
```

deeplow:
Our automated tests sometimes would fail when running `podman images --format {{.ID}}`. It turns out that in versions prior to podman 4.3.0, podman volumes (stored in `~/.local/share/contaiers`) would get corrupted when multiple tests were run in parallel.

The current solution is to wrap the test command to run sequentially in versions prior to the fix and in parallel for versions after that.

Fixes #217

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,8 @@ lint: lint-black lint-isort mypy ## check the code with various linters
 lint-apply: lint-black-apply lint-isort-apply ## apply all the linter's suggestions
 
 .PHONY: test
-test:  ## run tests in parallel
-	pytest -v -n 4 --cov
+test:
+	./dev_scripts/pytest-wrapper.py -v --cov
 
 # Makefile self-help borrowed from the securedrop-client project
 # Explaination of the below shell command should it ever break.
--- a/dev_scripts/pytest-wrapper.py
+++ b/dev_scripts/pytest-wrapper.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# PyTest Wrapper script - temporary solution to tests failing non-deterministically [1]
+# This is only fixed in podman v4.3.0. The wrapper essentially runs the tests in sequence
+# when the podman version is lower.
+#
+# [1]: https://github.com/freedomofpress/dangerzone/issues/217
+
+# FIXME this whole script should be removed and the respective Makefile calling line
+# replaced once all supported platforms have a podman version >= v4.3.0.
+
+import re
+import subprocess
+import sys
+
+import pytest
+from pkg_resources import parse_version
+
+from dangerzone.container import get_runtime_name
+
+PODMAN_MIN_VERSION = "4.3.0"
+
+
+def get_podman_version():
+    result = subprocess.run(
+        ["podman", "version", "--format", "'{{.Client.Version}}'"], capture_output=True
+    )
+    version = result.stdout.decode()[:-1]  # trim trailing \n
+    return version.split("-dev")[0]  # exclude "-dev" suffix from version
+
+
+def run_tests_in_parallel(pytest_args):
+    args = pytest_args + ["-n", "4"]
+    exit_code = pytest.main(args)
+
+
+def run_tests_in_sequence(pytest_args):
+    print("running tests sequentially")
+    exit_code = pytest.main(pytest_args)
+
+
+if __name__ == "__main__":
+
+    pytest_args = sys.argv[1:]  # exclude program names
+
+    if get_runtime_name() == "docker":
+        run_tests_in_parallel(pytest_args)
+    else:
+        podman_ver_minimum_parallel = parse_version(PODMAN_MIN_VERSION)
+        podman_ver_current = parse_version(get_podman_version())
+        if podman_ver_current >= podman_ver_minimum_parallel:
+            run_tests_in_parallel(pytest_args)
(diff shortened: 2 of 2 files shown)


deeplow:
The tests are failing but it appears to be unrelated to the PR.

Makefile
```diff
@@ -40,7 +40,7 @@ lint-apply: lint-black-apply lint-isort-apply ## apply all the linter's suggesti
 
 .PHONY: test
 test:  ## run tests in parallel
```
apyrgio, line 42: Let's remove the "run test in parallel" comment.

dev_scripts/pytest-wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+
+from dangerzone.container import get_runtime_name
+
+PODMAN_MIN_VERSION = "4.3.0"
+
+
+def get_podman_version():
+    result = subprocess.run(
+        "podman version --format '{{.Client.Version}}'", shell=True, capture_output=True
+    )
```
apyrgio, line 28: Not critical, but it's best not to use `shell=True` unless necessary. In this setting, it's not a problem, but it builds muscle memory.

(For a reason why, see https://docs.python.org/3/library/subprocess.html#security-considerations)

dev_scripts/pytest-wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+    version = result.stdout.decode()[:-1]  # trim trailing \n
+    return version.split("-dev")[0]  # exclude "-dev" suffix from version
```
apyrgio, line 30: Fun fact, Python 3.9 added [`str.removesuffix()`](https://docs.python.org/3/library/stdtypes.html#str.removesuffix). However, we can't use it yet, because Ubuntu Focal has Python 3.8 :-/. Anw, not actionable, just wanted to share this fact.

dev_scripts/pytest-wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def run_tests_parallel(pytest_args):
+    args = pytest_args + ["-n", "4"]
+    exit_code = pytest.main(args)
+
+
+def run_tests_sequence(pytest_args):
```
apyrgio, line 38: Minor, but I think you're looking for "serial" (or "sequentially" as an adverb).

apyrgio (approved):

dev_scripts/pytest-wrapper.py
deeplow: That's quite useful. Too bad that we can't use it :/

Makefile
deeplow: Sure, if you prefer it. I have changed it in [495d736](https://github.com/freedomofpress/dangerzone/pull/288/commits/495d73671ab7ea36b4b7f70a1050aa2d1f5f0169).

`run_tests_parallel()` would also need to follow the same logic so I changed it to `run_tests_in_parallel` and `run_tests_in_sequence()`, respectively. Does that sound good?

dev_scripts/pytest-wrapper.py
deeplow: Yes, I'm aware. In this case in particular, the string was a constant and it's more readable. But we can also do it as you suggest. I have changed it in [5a61fff](https://github.com/freedomofpress/dangerzone/pull/288/commits/5a61fff68ae10b0c8fa84e2cf84b04853dc74f7b).

dev_scripts/pytest-wrapper.py
deeplow: Sure, if you prefer it. I have changed it in [495d736](https://github.com/freedomofpress/dangerzone/pull/288/commits/495d73671ab7ea36b4b7f70a1050aa2d1f5f0169).

`run_tests_parallel()` would also need to follow the same logic so I changed it to `run_tests_in_parallel` and `run_tests_in_sequence()`, respectively. Does that sound good?

deeplow:
I had to rebase on top of main so the CI fixes from https://github.com/freedomofpress/dangerzone/pull/293/ would apply.

apyrgio:
I took a look at your comments, and they look good. Feel free to merge this.

Issue #270, opened 2022-11-25
Inconsistent Fonts across different Operating Systems


deeplow:
As can be seen in the comparison bellow fonts, widgets sizes, spacing differs across operating systems. Is there a way to reduce these differences such that we don't end up with issues like https://github.com/freedomofpress/dangerzone/pull/269 ?

| Mac | Fedora | Windows |
|--|--|--|
|<img src="https://user-images.githubusercontent.com/47065258/203991560-0dfa6ee5-aeeb-4985-8b57-668a8f212034.png" width=900> | <img src="https://user-images.githubusercontent.com/47065258/203991879-b9b12bf5-1c19-4a07-917f-98b2749bce61.png" width=900> | <img src="https://user-images.githubusercontent.com/47065258/203994786-466ae855-c101-4853-b375-13fb4b94bc3d.png" width=900> |

deeplow:
https://github.com/freedomofpress/dangerzone/pull/487 is another report of font inconsistencies, but this time color-related.

zenmonkeykstop:
FWIW I'm seeing this in 0.9.0 on Ubuntu Budgie 24.04:

![Image](https://github.com/user-attachments/assets/d0a86966-f70a-42f4-8e98-a890670e7cce)

apyrgio:
And is that stock Ubuntu 24.04? I've added this to the 0.10.0 milestone, because as is the text is not readable.

apyrgio:
Just an FYI, this is something that we have tried to tackle in this release for various reasons, but we just can't fix it 100% of the time. One change we've made is to use font hints in order to adjust our widgets accordingly. Our tests on Linux are very promising, but the situation on macOS is still buggy, and we don't why.

For the time being, we will increase the default height of the application to 550px, just to sidestep some immediate issues, but we need a better fix here.

almet:
Removing this from the `0.10.0` release, as we now have a way to circumvent the other issue we were facing (see https://github.com/freedomofpress/dangerzone/pull/1225).

This still needs to be solved!

PR #295, opened 2022-12-23
219 Version

deeplow:
Adds a `--version` parameter to the CLI and a the version number right next to header bar.

Due to https://github.com/freedomofpress/dangerzone/issues/270 on windows it looks a bit funny and out of place:

| OS | What it loooks like |
|------|----------------------------|
| Windows | image(8)
| Debian | dz-debian
| Fedora |dz-fedora
| MacOS | dz-macos


dangerzone/cli.py
```diff
@@ -38,6 +38,7 @@ def print_header(s: str) -> None:
     type=click.UNPROCESSED,
     callback=args.validate_input_filenames,
 )
+@click.version_option(version=get_version())
```
apyrgio, line 41: Suggestion: if we add `... , message="%(version)s)`, we don't have to see the prog name in the output, nor do any special parsing.

apyrgio (approved):

dangerzone/cli.py
deeplow: I missed this one. Actually, we don't even need to specify anything. `@click.version_option()` works out of the box. And it's function is a bit smarter than our `get_version()`.

It would be nice if we could use click's method for determining the version in runtime instead of having `share/version.txt`. But for now I'd say we can leave that since many things depend on that method (`setup.py`, `build-{deb,rpm}.py, etc.)

So I think I'll go ahead and leave it as it was (since you approved the PR already). This way getting the version from `share/version.txt` is the canonical place.

Is that OK?

apyrgio: Yeah, my comment was not directed to `get_version()`, but making `--version` print just the version number, instead of the program name as well, which makes it difficult to parse (I'm referring to the code in the `tests/`)

Anw, that's just a suggestion. Feel free to merge it as is.

deeplow: Ah. Then I agree. I have added `... , message="%(version)s)` and simplified the test `tests/test_cli.py::TestCliBasic::test_version`.

apyrgio: Nice, feel free to merge.

deeplow:
Squashed FIXUPs and rebased from main.

Issue #211, opened 2022-09-21
Migrate to Qt6 before Qt5 end-of-life


deeplow:
Mirroring https://github.com/freedomofpress/securedrop-client/issues/1562:

> According to [Qt](https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders), Qt 5.15 will reach EOL on 2023-05-26 for non-subscribers. We should ensure we migrate to a supported version before then.

We have a [PR](https://github.com/freedomofpress/dangerzone/pull/164) that already does this change but it is blocked on distros lacking Pyside6 (Qt6) packages.

deeplow:
I checked on Dec. 12th 2022 the current availability of the `python3-pyside6` package which we need for Qt6 in python. I checked these for the currently supported systems. It is not looking promising:

| Distro | version | Qt6 | PySide6 | 
|-----------|-----------|:----------------------:|:---------------------------:|
| Ubuntu | 22.10 (kinetic) | :heavy_check_mark: |[:x:](https://packages.ubuntu.com/kinetic/allpackages) |
| Ubuntu | 22.04 (jammy) |:heavy_check_mark: | [:x:](https://packages.ubuntu.com/jammy/allpackages) | 
| Ubuntu |  20.04 (focal)  |:x:  | [:x:](https://packages.ubuntu.com/focal/allpackages) |
| Debian |  12 (bookworm) | [:heavy_check_mark:](https://packages.debian.org/bookworm/qt6-base-dev-tools) | [:x:](https://packages.debian.org/search?suite=bookworm&section=all&arch=any&searchon=names&keywords=python3-pyside6) |
| Debian    | 11 (bullseye)  | [:heavy_check_mark:](https://packages.debian.org/bullseye-backports/qt6-base-dev-tools)   | [:x:](https://packages.debian.org/search?suite=bullseye&section=all&arch=any&searchon=names&keywords=python3-pyside6) | 
| Fedora | 37 | [:heavy_check_mark:](https://packages.fedoraproject.org/pkgs/qt6-qtbase/qt6-qtbase/)  | [:x:](https://packages.fedoraproject.org/search?query=python3-pyside6&releases=Fedora+37&start=0)
| Fedora|  36 |  [:heavy_check_mark:](https://packages.fedoraproject.org/pkgs/qt6-qtbase/qt6-qtbase/) | [:x:](https://packages.fedoraproject.org/search?query=python3-pyside6&releases=Fedora+36&start=0)
[...]

PR #296, opened 2022-12-23
Adds PySide 6 Support For Windows and MacOS

### LINKED ISSUE #117: Improve Overall UX
I have volunteered to develop a plausible (to reasonably implement) and intuitive ux solution to guide implementation of #77, in the GUI.

Internews is funding some of my work on the [SecureDrop Worksation](https://workstation.securedrop.org/) project, through its [Usable.Tools](https://usable.tools/) fund. _DangerZone_ has emerged as a relevant tool in Worksation users' workflows, and we (the SecureDrop team) would like to contribute a few of my hours from the usable.tools award, towards making _DangerZone_ more intuitive—and extending that functionality to support bulk file sanitization. 

I am creating this issue to track and collect feedback against work towards improving _Dangerzone's_ _general_ usability, in a light design sprint with occasional follow-up. This work is likely to include my asking some of questions, and my submission of wireframes and interactive proof-of-concept video(s) for review.

### LINKED ISSUE #177: Support building on M1 macs
The `shiboken2` package fails to install on m1s because it has no builds for ARM architectures on macbooks ([no available wheels](https://pypi.org/simple/shiboken2/)).

---

 When running `poetry install` on an m1 macbook:

error_shiboken

_Originally posted by @deeplow in https://github.com/freedomofpress/dangerzone/issues/172#issuecomment-1194050403_

deeplow:
I tested:
- **MacOS**: dev env + building and running a `.app`
- **Windows**: dev env + building and running `.exe`
-  **Linux**: dev env (still on PySide2) -- see [reasoning here](https://github.com/freedomofpress/dangerzone/issues/211)

Fixes #177

> **Note**: When building on Windows, CXFreeze apparently copies the entire poetry virtualenv into the `.exe` and when you run `poetry install` it only adds the new software -- it doesn't remove old one. So I moved from `PySide2` to `PySide6` and `PySide2` was still installed but no longer the dependencies. However, I saw in the build log that it was getting included.
> 
> To ensure we don't ship `PySide2` accidentally it's recommended to uninstall the virtualenv with `poetry env remove [virtualenv]`

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -5,11 +5,20 @@ import platform
 import shutil
 import subprocess
 import tempfile
+import typing
 from multiprocessing.pool import ThreadPool
 from typing import List, Optional
 
 from colorama import Fore, Style
-from PySide2 import QtCore, QtGui, QtWidgets
+
+# FIXME: See https://github.com/freedomofpress/dangerzone/issues/320 for more details.
+if typing.TYPE_CHECKING:
+    from PySide2 import QtCore, QtGui, QtWidgets
+else:
+    try:
+        from PySide6 import QtCore, QtGui, QtWidgets
+    except ImportError:
+        from PySide2 import QtCore, QtGui, QtWidgets
 
 from .. import errors
 from ..document import SAFE_EXTENSION, Document
@@ -385,8 +394,10 @@ class SettingsWidget(QtWidgets.QWidget):
         self.safe_extension_name_layout.addWidget(self.safe_extension_filename)
         self.safe_extension_name_layout.addWidget(self.safe_extension)
 
-        dot_pdf_regex = QtCore.QRegExp(r".*\.[Pp][Dd][Ff]")
-        self.safe_extension.setValidator(QtGui.QRegExpValidator(dot_pdf_regex))
+        dot_pdf_regex = QtCore.QRegularExpression(r".*\.[Pp][Dd][Ff]")
+        self.safe_extension.setValidator(
+            QtGui.QRegularExpressionValidator(dot_pdf_regex)
+        )
         self.safe_extension_layout = QtWidgets.QHBoxLayout()
         self.safe_extension_layout.addWidget(self.save_checkbox)
         self.safe_extension_layout.addWidget(self.safe_extension_label)
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -46,7 +46,7 @@ class MainWindow(QtWidgets.QMainWindow):
         header_label.setFont(self.dangerzone.fixed_font)
         header_label.setStyleSheet("QLabel { font-weight: bold; font-size: 50px; }")
         header_version_label = QtWidgets.QLabel(get_version())
-        header_version_label.setProperty("class", "version")  # type: ignore [arg-type]
+        header_version_label.setProperty("class", "version")
         header_version_label.setAlignment(QtCore.Qt.AlignBottom)
 
         header_layout = QtWidgets.QHBoxLayout()
@@ -363,7 +363,7 @@ class SettingsWidget(QtWidgets.QWidget):
         self.docs_selected_label = QtWidgets.QLabel("No documents selected")
         self.docs_selected_label.setAlignment(QtCore.Qt.AlignCenter)
         self.docs_selected_label.setContentsMargins(0, 0, 0, 20)
-        self.docs_selected_label.setProperty("class", "docs-selection")  # type: ignore [arg-type]
+        self.docs_selected_label.setProperty("class", "docs-selection")
 
         # Save safe version
         self.save_checkbox = QtWidgets.QCheckBox()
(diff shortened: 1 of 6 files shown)


dangerzone/gui/main_window.py
```diff
@@ -9,7 +9,15 @@
 from typing import List, Optional
 
 from colorama import Fore, Style
-from PySide2 import QtCore, QtGui, QtWidgets
+
+try:
+    from PySide6 import QtCore, QtGui, QtWidgets
+
+    # Alias renamed Qt6 methods for backwards compatibility
+    QtCore.QRegExp = QtCore.QRegularExpression
```
apyrgio, line 17: In their [docs](https://doc-snapshots.qt.io/qt6-dev/qtcore-changes-qt6.html#regular-expression-classes), they propose using `QRegularExpression` in Qt5, as a migration path towards Qt6. I propose we do the same.

From a cursory glance on the sole regular expression that we use, I think we will not be affected by this change.

apyrgio:
Some minor comments:

1. In your first commit message, you write:

    > in Linux distro's packages, whereas with Linux and macOS
    
    it should be "whereas on Windows and macOS".

2. I think we don't need this line anymore:

    https://github.com/freedomofpress/dangerzone/blob/84b8212e5d85de7a1daf3d522d4f48402a1cbe00/install/windows/build-wxs.py#L25-L27
    
3. Have you tried out running `mypy` against our codebase on Windows / macOS, with this PR? I'm asking because we no longer use the `Pyside2-stubs`, and I see reports of mypy failing on PySide6 ([example](https://github.com/python/mypy/issues/10209)).

dangerzone/gui/main_window.py
deeplow: Good catch

deeplow:
> Have you tried out running mypy against our codebase on Windows / macOS, with this PR? I'm asking because we no longer use the Pyside2-stubs, and I see reports of mypy failing on PySide6 (https://github.com/python/mypy/issues/10209).

Even on linux it shows multiple errors. How come CI didn't run on this branch?

I came across mypy and pyside2 issues [before](https://github.com/freedomofpress/dangerzone/commit/bc7188eb4dfc6fafc082f6b8fc1d341f3b1f3e5d). And apparently this is still [an upstream issue](https://bugreports.qt.io/browse/PYSIDE-1675). I wonder if we'll need a PySide6-stubs or similar to fix it...

> it should be "whereas on Windows and macOS".

Thanks! I'll fix this when I rebase.

> I think we don't need this line anymore:

Done. I checked on windows and there is no `examples` lib either. So we don't have to exclude the pyside6 equivalent.

deeplow:
> > Have you tried out running mypy against our codebase on Windows / macOS, with this PR? I'm asking because we no longer use the Pyside2-stubs, and I see reports of mypy failing on PySide6 ([python/mypy#10209](https://github.com/python/mypy/issues/10209)).

With the new fix it runs on Linux and macOS.

apyrgio (approved):

deeplow:
(rebased on main)

apyrgio:
## Proposal

The best way forward would be to use `if sys.platform == "linux" ... else: ...`, and move the imports there. However, we can't do this until PySide6 fixes its Mypy issues.

The next best thing is this:

```python
if typing.TYPE_CHECKING:
    from PySide2 import QtCore, QtGui, QtWidgets
else:
    try:
        from PySide6 import QtCore, QtGui, QtWidgets
    except ImportError:
        from PySide2 import QtCore, QtGui, QtWidgets
```

and install PySide2 stubs in all platforms. The rationale is the following:

1. The PySide2 and PySide6 interfaces we are using happen to be the same, so the checks we do for PySide2 will apply to PySide6 as well.
2. Environments only with PySide2 remain unaffected.
3. The `--warn-unused-ignores` flag can be reinstated.

deeplow:
Wow. Thanks for digging deeper into this. Something did in fact feel old about my "hack" but since it was just a linter, it didn't seem much consequential. I totally agree with your assessment. I will attempt to implement it now.

deeplow:
> 3\. 
> 
> 
> Check this out in the `117-apple-silicon-mypy` branch ([8dfec99](https://github.com/freedomofpress/dangerzone/commit/8dfec99d9ed4349a712a1aaff02bc62c0c68f88d)).

This may work on windows but I it leads to the initial problem on macOS (https://github.com/freedomofpress/dangerzone/issues/177). And on a fresh virtualenv on fedora-36, it's getting stuck on installing `shiboken2` as well:

```
  • Installing shiboken2 (5.15.2.1): Pending...
```

It just hangs there...

apyrgio:
Current state of this PR:

I had made some recent comments and research regarding Mypy, so we decided to take over this part, and implement the fixes myself. Here's what I've done in general on this PR:

1. I squashed all the fixes we have agreed on and tidied up the commits.
2. I switched `PySide2-stubs` with `types-PySide2`. These stubs are more complete (we even remove some `type: ignore` statements) and don't bring `PySide2` as a dependency. This means that we can use them in all platforms, even M1 MacOS.
3. I've updated the original issue with some decisions we took along the way on this PR (see https://github.com/freedomofpress/dangerzone/issues/177#issuecomment-1404059314).
4. I've pushed the above to `117-apple-silicon-mypy`. I couldn't push to the original branch, because it's in your forked repo.

My suggestion is to:
1. Change the author in the commits of this branch, and sign the commits with your key.
2. Force push the `*-mypy` branch to the original branch, and then merge it.
    * This way, we will retain a pointer to the topics we discussed here.

deeplow:
I like how the classify the PySide original stubs as abysmal :rofl: on the types-PySide2 [project page](https://pypi.org/project/types-PySide2/):

abysmal

deeplow:
The changes LGTM. And I tested on Mac and Windows and it checks out.

> My suggestion is to:
>     1. Change the author in the commits of this branch, and sign the commits with your key.
>     2. Force push the `*-mypy` branch to the original branch, and then merge it.

I have done so now and I've rebased it on top of main.

PR #301, opened 2023-01-02
Fix CLI logging in non-dev enviornments

deeplow:
In non-development environments, the logging was mistakenly logged at the `DEBUG` level. Only a few messages were at this level, though.

This PR enables the right level of logging and fixes its formatting.

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -105,11 +105,28 @@ args.override_parser_and_check_suspicious_options(cli_main)
 
 
 def setup_logging() -> None:
+    class EndUserLoggingFormatter(logging.Formatter):
+        """Prefixes any non-INFO log line with the log level"""
+
+        def format(self, record: logging.LogRecord) -> str:
+            if record.levelno == logging.INFO:
+                # Bypass formatter: print line directly
+                return record.getMessage()
+            else:
+                return super().format(record)
+
     if getattr(sys, "dangerzone_dev", False):
         fmt = "[%(levelname)-5s] %(message)s"
         logging.basicConfig(level=logging.DEBUG, format=fmt)
     else:
-        logging.basicConfig(level=logging.INFO, format="%(message)s")
+        # prefix non-INFO log lines with the respective log type
+        fmt = "%(levelname)s %(message)s"
+        formatter = EndUserLoggingFormatter(fmt=fmt)
+        ch = logging.StreamHandler()
+        ch.setFormatter(formatter)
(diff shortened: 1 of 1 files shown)


dangerzone/cli.py
```diff
@@ -104,11 +104,11 @@ def cli_main(
...
 
 def setup_logging() -> None:
-    if getattr(sys, "dangerzone_dev", True):
-        fmt = "%(message)s"
+    if getattr(sys, "dangerzone_dev", False):
+        fmt = "[%(levelname)-5s] %(message)s"
         logging.basicConfig(level=logging.DEBUG, format=fmt)
     else:
-        logging.basicConfig(level=logging.ERROR, format=fmt)
+        logging.basicConfig(level=logging.INFO, format="%(message)s")
```
apyrgio, line 111: Looks good. One minor question though: why don't we print the error level in the basic configuration, and we print just the message?

deeplow: This is the "logging" that end-users are supposed to see when they run the CLI version of dangerzone. So I don't think we should show them 'INFO before all the messages.

But you're right. Probably for `WARN` and `ERROR`, we want to label them as such. I'm looking at way of doing that.

deeplow: I have fixed this now in https://github.com/freedomofpress/dangerzone/pull/301/commits/565ad300ceea509ee7f6a93099a3e9cdf5b5cc79. You can test it out by setting `logger.setLevel()` to `logging.DEBUG` and seeing that it prints a `DEBUG` line:

```
[...]
DEBUG Marking doc 4LAzqc as 'converting'
> /usr/bin/podman run --network none -u dangerzone  [...]
```

deeplow: What do you think about this solution?

apyrgio: Hm, I think the types of users that will see these logs will probably be the users who won't mind seeing the `INFO` log level. So, I wouldn't add this extra complexity. If it's an eyesore for you, sure, feel free to merge it.

apyrgio: Okey dokey!

apyrgio (approved):

dangerzone/cli.py
deeplow: Whoops. My mistake: I accidentally implemented this logic for a dev environment and not the final version (the one users run). Maybe this was what confused you.

deeplow:
Squashed commits and rebased from main.

Issue #229, opened 2022-10-19
Automated Testing in Windows & Mac


deeplow:
Apparently with CircleCI we can have our [own runners](https://circleci.com/docs/runner-overview/) in our machines. We could use this to add Mac & Windows to our CI without having to pay the [exaggerated](https://circleci.com/docs/dedicated-hosts-macos/#pricing-and-specs) Circle CI Baremetal Mac pricing.

deeplow:
I started working on this recently.

PR #302, opened 2023-01-04
Abstract Isolation Providers and Add Dummy One

deeplow:
Encapsulates isolation providers (will be useful for adding Qubes support later) and adds a dummy one. It can be run on both the CLI and GUI with the `--unsafe-dummy-conversion` flag (hidden and available only in the testing environment). This is a step for [adding Windows and MacOS](https://github.com/freedomofpress/dangerzone/issues/229) to the CI testing environment, given that this way (at the cost of some fidelity) we gain the ability to run automated tests without the need for nested virtualization.

```
$ ./dev_scripts/dangerzone-cli --unsafe-dummy-conversion tests/test_docs/sample.gif
```

It runs as expected but doesn't actually do any conversion. Instead it shows preset output conversion progress messages and always "succeeds" in the conversion.

```
╭──────────────────────────╮
│           ▄██▄           │
│          ██████          │
│         ███▀▀▀██         │
│        ███   ████        │
│       ███   ██████       │
│      ███   ▀▀▀▀████      │
│     ███████  ▄██████     │
│    ███████ ▄█████████    │
│   ████████████████████   │
│    ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀    │
│                          │
│    Dangerzone v0.4.0     │
│ https://dangerzone.rocks │
╰──────────────────────────╯
Assigning ID 'Mw9Eqs' to doc '/home/user/dangerzone/tests/test_docs/sample.gif'

Marking doc Mw9Eqs as 'converting'
Dummy converter started:
  - document: sample.gif (Mw9Eqs)
  - ocr     : None

(simulating conversion)
[doc Mw9Eqs] 0.0% Converting to PDF using GraphicsMagick
[doc Mw9Eqs] 3.0% Separating document into pages
[doc Mw9Eqs] 5.0% Converting page 1/1 to pixels
[doc Mw9Eqs] 50.0% Converted document to pixels
[doc Mw9Eqs] 50.0% Converting page 1/1 from pixels to PDF
[doc Mw9Eqs] 95.0% Merging 1 pages into a single PDF
[doc Mw9Eqs] 97.0% Compressing PDF
[doc Mw9Eqs] 100.0% Safe PDF created
Marking doc Mw9Eqs as 'safe'

Safe PDF(s) created successfully
/home/user/dangerzone/tests/test_docs/sample-safe.pdf
```

> **Note**: No actual 'safe' PDF is created (as opposted to what the last line says).


apyrgio:
Minor commit message issues:

1. In `Isolation-provider-specific methods in _convert()`, you're missing a verb ("All isolation providers will some similar steps ...")

dangerzone/isolation_provider.py
```diff
@@ -0,0 +1,501 @@
...
+            self._print_progress(document, error, text, percentage)  # type: ignore [arg-type]
+            if stdout_callback:
+                stdout_callback(error, text, percentage)
+            if error:
+                success = False
+            time.sleep(0.2)
+
+        return success
+
+    def _print_progress(
```
apyrgio, line 416: We can move this method to `IsolationProvider`, and use it both from the `Container` and `Dummy` classes. Even though the class is an abstract one, it's legal to have working methods in it.

dangerzone/gui/main_window.py
```diff
@@ -51,9 +50,15 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
...
-        self.waiting_widget.finished.connect(self.waiting_finished)
+        if isinstance(self.dangerzone.isolation_provider, isolation_provider.Container):
+            # Waiting widget replaces content widget while container runtime isn't available
+            self.waiting_widget: WaitingWidget = WaitingWidgetContainer(self.dangerzone)
+            self.waiting_widget.finished.connect(self.waiting_finished)
+
+        elif isinstance(self.dangerzone.isolation_provider, isolation_provider.Dummy):
+            # Don't wait with dummy converter
+            self.waiting_widget = WaitingWidget()
+            self.dangerzone.is_waiting_finished = True
```
apyrgio, line 61: Why is this distinction necessary? It would be nice if we could check the code's behavior if `IsolationProvider.install()` fails.

For instance, I just realized that we have this code:

https://github.com/freedomofpress/dangerzone/blob/84b8212e5d85de7a1daf3d522d4f48402a1cbe00/dangerzone/container.py#L83-L85

but we don't check for `False` either in `cli.py` or `main_window.py`. Having an easy way to trigger this code path in our test via the `Dummy` provider I believe would help.

dangerzone/isolation_provider.py
```diff
@@ -0,0 +1,501 @@
...
+            + prevent_leakage_args
+            + extra_args
+            + [self.CONTAINER_NAME]
+            + command
+        )
+
+        args = [container_runtime] + args
+        return self._exec(document, args, stdout_callback)
+
+    def _convert(
```
apyrgio, line 259: Hot take: everything aside from the command creation and execution should be factored out of this class, so that the `Dummy` provider can use it as well. This will improve the testing surface a lot, as we will be able to set a GUI option (e.g., change the prefix), and see it actually executed.

apyrgio (commented):
@deeplow I just finished my first pass of the PR. Feel free to share your thoughts.

dangerzone/isolation_provider.py
deeplow: Good point. Done in [df6231d](https://github.com/freedomofpress/dangerzone/pull/302/commits/df6231d4464c02eb28a1cff22ff4e63a1cf0aa08).

dangerzone/gui/main_window.py
deeplow: Yes. I'm aware: https://github.com/freedomofpress/dangerzone/issues/193. But I feel it may be out of scope for this PR. What do you think?

dangerzone/isolation_provider.py
deeplow: Why would the following make sense in `Dummy`? `is_container_installed()`, `get_runtime_name()` and `get_runtime()`. Theoretically, wouldn't we have tests for each isolation provider? In that sense we could test coverage specifically for `Container`.

Other than that, I think `parse_progress()` could make sense to get factored out.

dangerzone/gui/main_window.py
apyrgio: ... oh. I think at some point we should though, since we have this code path fresh. I'll add this issue in 0.4.1.

apyrgio: And now I see that you have added it in 0.5.0. Ok, let's leave it there.

dangerzone/isolation_provider.py
apyrgio: Actually, I'm referring to the code in the `_convert()` method, not the methods outside of it.

deeplow: I agree. I didn't do it originally because passing the extra variables to `_convert()` would make it look like a train of arguments. But actually it's just two extra. So let's do that. 

Let me know what you think: [cb325ca](https://github.com/freedomofpress/dangerzone/pull/302/commits/cb325ca386d2c12f6872a319a27ca1a06ef986b6).

apyrgio: It's in a right direction, but I'd like to press a bit more. If we also factor out the last part, where we move the converted file to the safe dir, then we can improve our test coverage as well (see this [thread](https://github.com/freedomofpress/dangerzone/pull/303#discussion_r1071362787)).

deeplow: Oh. Right, I mean to include, but forgot about it. Thanks!

deeplow: Now thinking a bit more about this, I'm not sure it makes much sense.

I think it's clearer if the `_convert()` does the conversion, returns a success bool and saves the final file in `document.output_filename`. How it does it should be implementation-specific: it may create a `/safe`, for example and put the intermediate file there, but all that matters is that in the end if `success is True`, then the output file must be in `document.output_filename` and we can test if that happened.

This extra flexibility would allow a future implementation (e.g. Qubes) where it reads directly from the `pixels_dir` and converts to the output PDF without needing the `safe_dir`. It could move the file directly to `document.output_filename`. 

Does this make sense? 

> [..] as we will be able to set a GUI option (e.g., change the prefix), and see it actually executed.

This thing, for example, won't be a problem since if there is a prefix change, that will be reflected in `document.output_filename` and the conversion will know to move the final file there.

So I propose we leave it as it originally was. Thoughts?

deeplow: Per verbal discussion, lets leave this in its original form.

deeplow: I have now revered the changes and save a dummy PDF so all tests pass on platforms where the dummy converter will be enabled.

deeplow:
I have now revered the changes and save a dummy PDF so all tests pass on platforms where the dummy converter will be enabled.

apyrgio:
I checked your fixups and they look good. I also tested the code a bit locally. Feel free to merge!

deeplow:
Rebased on top of main and added one last FIXUP commit which made it to [wrong branch](https://github.com/freedomofpress/dangerzone/pull/303#issuecomment-1397160953).

apyrgio:
I saw the FIXUP commit. I'm ok with the static method, but just wanted to mention that you are free to use `self.` for static methods. You don't need to refer to the class by name.

apyrgio (approved):

dangerzone/isolation_provider/container.py
```diff
@@ -0,0 +1,317 @@
...
+import shutil
+import subprocess
+import tempfile
+from typing import Callable, List, Optional, Tuple
+
+import appdirs
+
+from ..document import Document
+from ..util import get_resource_path, get_subprocess_startupinfo
+from . import IsolationProvider
```
apyrgio, line 16: Oh, there's one thing I realized:

It's weird practice to have a submodule do an import from `__init__.py`. In general, it's preferable to have a `base.py` module, and import that. The reason is that you may want to re-export the isolation provider classes in `__init__.py`, and doing so would lead to cyclic references.

deeplow: Just added that in [fd9bd37](https://github.com/freedomofpress/dangerzone/pull/302/commits/fd9bd37cb394b774da664c285cdd1c67d0018aa2).

apyrgio: Niceeee... but! The CI tests are failing due to this move. Once you fix them, feel free to merge.

deeplow:
> just wanted to mention that you are free to use `self.` for static methods. You don't need to refer to the class by name.

I wish I could, but because I am calling this already from a static context, I don't have access to self.

deeplow:
Rebased from main and squashed commits.

deeplow:
Fedora-36 failed but I'll merge it anyways since it appears that the issue is not related to the PR.

PR #303, opened 2023-01-04
Run CLI tests in Windows and macOS via Github Actions

deeplow:
Depends on https://github.com/freedomofpress/dangerzone/pull/302.

Configures GitHub Actions to run tests with the dummy converter and disables certain tests that fail because of said converter.

Github Actions was chosen for this as opposed to our usual Cicle CI due to the lack of cost and the issues the folks at OnionShare experienced.

>**Note**: the windows evironment appears to be Windows Server and not a end-user version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,8 @@ on:
   push:
   pull_request:
     branches: [ main ]
+  schedule:
+    - cron: '0 0 * * *' # Run every day at 00:00 UTC.
 
 jobs:
   windows:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: Tests
+on:
+  push:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  windows:
+    runs-on: windows-latest
+    env:
+      DUMMY_CONVERSION: True
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: pip install poetry
+      - run: poetry install
+      - name: Run CLI tests
+        run: poetry run make test
+
+  macOS:
+    runs-on: macos-latest
+    env:
+      DUMMY_CONVERSION: True
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: pip install poetry
+      - run: poetry install
+      - name: Run CLI tests
+        run: poetry run make test
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -126,6 +126,9 @@ class TestCli(TestBase):
             # to tokenize it.
             args = (args,)
 
+        if os.environ.get("DUMMY_CONVERSION", False):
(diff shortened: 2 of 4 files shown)


deeplow:
Should we pin the windows and macOS versions like we do with others?

deeplow:
We may only want to run this on `pull_request` instead of also `push` because this leads to duplicate results:
Screenshot 2023-01-05 at 10-33-57 Pull requests · freedomofpress_dangerzone

tests/test_cli.py
```diff
@@ -208,13 +211,15 @@ def test_lang_eng(self) -> None:
             "spaces test.pdf",
         ],
     )
+    @pytest.mark.skipif(dummy_conversion(), reason="Won't run with dummy converter")
```
apyrgio, line 214: I believe that skipping these tests will not be necessary once we fix https://github.com/freedomofpress/dangerzone/pull/302#discussion_r1068369476. But let's discuss more on that thread.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+      DUMMY_CONVERSION: True
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: pip install poetry
+      - run: poetry install
+      - name: Run CLI tests
+        run: poetry run pytest -v
```
apyrgio, line 20: Quick question, why don't we use `make test` here? I think it should be available in the CI runner.

apyrgio (changes requested):
I made the first pass of the PR. Feel free to comment on my suggestions.

apyrgio:
> We may only want to run this on `pull_request` instead of also `push` because this leads to duplicate results:
> 
> Screenshot 2023-01-05 at 10-33-57 Pull requests · freedomofpress_dangerzone

I see that there's already a question for that, with no satisfying answer: https://github.com/orgs/community/discussions/26276. Personally, I like the fact that we are able to run tests on dev branches, so I'd be ok with having duplicate tests on PRs. If it bothers us too much, we can limit them to PRs, and open a draft PR when we want to perform a test.

apyrgio:
> Should we pin the windows and macOS versions like we do with others?

Regarding pinning the Windows and macOS versions, I see that the best we can do is pin them to `windows-2022` or `macos-12` (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on).

I'm fine with keeping it as is though, since these versions change very slowly, and it's generally best to test on the bleeding edge. If you prefer not to, do change it.

.github/workflows/ci.yml
deeplow: By default windows doesn't have `make`. But I can try and see if it works.

deeplow: You're right! Make is available! https://github.com/freedomofpress/dangerzone/actions/runs/3948344493/jobs/6758222698

Closing this.

deeplow:
Whoops. Should be fixed now.

apyrgio:
Just saw the changes in the branch. They look good, so feel free to merge.

apyrgio (approved):

deeplow:
Rebased, squashed commits and merged.

PR #310, opened 2023-01-17
Show Exceptions Raised During Document Conversion

### LINKED ISSUE #309: Exception in Conversion Does Not Lead to Dangerzone Failing
To reproduce this, I raised an exception at the end of the conversion process (in `container.py`). However, the program would run fine and not report the exception at all (exit code 0):

```shell_session
$ ./dev_scripts/dangerzone-cli ./tests/test_docs/sample.pdf 
[doc 1i6ll0] 0.0% Converting to PDF using GraphicsMagick
[doc 1i6ll0] 3.0% Separating document into pages
[doc 1i6ll0] 5.0% Converting page 1/1 to pixels
[doc 1i6ll0] 50.0% Converted document to pixels
[doc 1i6ll0] 50.0% Converting page 1/1 from pixels to PDF
[doc 1i6ll0] 95.0% Merging 1 pages into a single PDF
[doc 1i6ll0] 97.0% Compressing PDF
[doc 1i6ll0] 100.0% Safe PDF created
```

This can lead to false positives in our testing environment (which currently relies primarily in checking the exit code -- which will need to be revisited soon).

deeplow:
Exceptions raised during the document conversion process would be silently hidden. This was because ThreadPoolExecuter created various threads and hid any exceptions raised.

Fixes #309

### DIFF
--- a/dangerzone/document.py
+++ b/dangerzone/document.py
@@ -200,3 +200,6 @@ class Document:
         if not isinstance(other, Document):
             return False
         return self.input_filename == other.input_filename
+
+    def __hash__(self) -> int:
+        return hash(self.id)
--- a/dangerzone/logic.py
+++ b/dangerzone/logic.py
@@ -1,4 +1,3 @@
-import concurrent.futures
 import gzip
 import json
 import logging
@@ -7,7 +6,9 @@ import platform
 import shutil
 import subprocess
 import sys
-from typing import Callable, List, Optional
+import traceback
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import Callable, Dict, List, Optional
 
 import appdirs
 import colorama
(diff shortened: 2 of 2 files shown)


apyrgio:
I tested this PR a bit, and it works, but I see some issues. Let me explain:

The main problem is that if an exception occurs during conversion, it's never logged, since it happens in a different execution context. The solution in this PR is to:

* Submit conversion jobs for each document.
* Grab the result of each conversion in the main thread.
* If it's an exception, log an error and print a traceback. Else, continue with the next conversion result.

The thing is, we don't need to involved the main thread in this. We can effectively do the same by printing the exception within the worker thread:

```patch
-        def convert_doc(document: Document) -> None:
-            success = container.convert(
-                document,
-                ocr_lang,
-                stdout_callback,
-            )
+        def convert_doc(document: Document) -> None:
+            try:
+                success = container.convert(
+                    document,
+                    ocr_lang,
+                    stdout_callback,
+                )
+            except Exception:
+                log.exception(
+                    f"An exception occurred while converting document '{document.id}'"
+                )
```
---

So, suppose we simplify the error reporting bit. This still misses some other problems that are tangent to the main one:

* We don't mark the document as failed, if an exception occurs.
* We don't mark the document as failed, if `success == False`.
* We don't report exceptions when we run the Dangerzone GUI from the CLI.

Therefore, my suggestion is to rebase this PR on top of the dummy providers one, and:

1. Move the `try ... except` block in the `IsolationProvider.convert()` method. This way, both the GUI and CLI will report errors.
2. Mark the document as failed in `IsolationProvider.convert()`, if `success == False` or an exception occurs.

What do you think?

deeplow:
I agree with that.

deeplow:
Closed in favor of https://github.com/freedomofpress/dangerzone/pull/313

PR #312, opened 2023-01-20
Fix Linter Failure in CI

deeplow:
- unpins poetry version by using alternative fix to upstream bug in poetry (following @eloquence's [suggestion](https://github.com/freedomofpress/redmine_docs/blame/switch-to-poetry/.circleci/config.yml#L10))
- Fixes out of sync `BUILD.md` instructions and `qa.py`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ jobs:
           # Debian bug [1]. Once this bug is fixed, revert to installing
           # Poetry from the Debian repos instead, which is more stable.
           #
-          # Also, we pin the Poetry version to 1.2.2, to sidestep a Poetry bug
+          # Also, we run poetry with --no-ansi, to sidestep a Poetry bug
           # [2] that currently exists in 1.3.
           #
           # [1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029156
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -90,16 +90,16 @@ jobs:
...
           #
           # [1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029156
           # [2]: https://github.com/freedomofpress/dangerzone/issues/292#issuecomment-1351368122
           command: |
             apt-get update
             apt-get install -y make python3 python3-pip --no-install-recommends
-            pip install poetry==1.2.2
-            poetry install --only lint
+            pip install poetry
+            poetry install --no-ansi --only lint
```
apyrgio, line 102: If we plan to use Poetry 1.3, then it may be worth checking if the [original Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029156) still applies, so that we can ditch installing Poetry from PyPI.

apyrgio (approved):

.circleci/config.yml
deeplow: In theory it should be fixed, but I tested without the fix and [it still broken](https://app.circleci.com/pipelines/github/freedomofpress/dangerzone/862/workflows/dbeb452b-286f-4825-9f14-c58b98a37a5c/jobs/5956). It's some issue again where poetry runs fine but exits non-zero.

So I'll merge this.

PR #313, opened 2023-01-20
Show Exceptions Raised During Document Conversion

deeplow:
Depends on #302; Replaces #310. 

Exceptions raised during the document conversion process would be silently hidden. This was because ThreadPoolExecuter created various threads and hid any exceptions raised.

Fixes https://github.com/freedomofpress/dangerzone/issues/309

### DIFF
--- a/dangerzone/isolation_provider/base.py
+++ b/dangerzone/isolation_provider/base.py
@@ -26,7 +26,13 @@ class IsolationProvider(ABC):
         stdout_callback: Optional[Callable] = None,
     ) -> None:
         document.mark_as_converting()
-        success = self._convert(document, ocr_lang, stdout_callback)
+        try:
+            success = self._convert(document, ocr_lang, stdout_callback)
+        except Exception:
+            success = False
+            log.exception(
+                f"An exception occurred while converting document '{document.id}'"
+            )
         if success:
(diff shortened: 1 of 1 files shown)


dangerzone/isolation_provider/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        stdout_callback: Optional[Callable] = None,
+    ) -> None:
+        document.mark_as_converting()
+        try:
+            success = self._convert(document, ocr_lang, stdout_callback)
+        except Exception:
+            success = False
+            log.exception(
+                f"An exception occurred while converting document '{document.id}'"
+            )
```
apyrgio, line 35: We also need to do `success = False` at the end of this block, since it's not defined. Other than that, looks good.

deeplow: Sorry, I don't follow. I thin it should be set either in `try` or via `except`. Is there any other way this could go?

apyrgio: What tha, I thought that there was no `success = False` line in the except block. My eyes must have fooled me, sorry.

apyrgio (approved):

deeplow:
Rebased on top of main.

PR #322, opened 2023-01-26
Add a required --distro option to build-deb.py

maeve-fpf:
Since building with a Debian toolchain and an Ubuntu toolchain produce slightly different `.deb`s with different checksums, we want to tag each with the distro type in order to give them different filenames. This will allow them to coexist in the `packages.freedom.press` pool. (This is just about how packages are compressed right now, but perhaps someday there might be more significant differences, like one distro changes something related to container runtimes first.)

I've added `--distro` as a command-line option to match some other scripts in the repo. This is a required option for `build-deb.py`, so references to it in docs have been updated. I don't know if YAML inclusion (for CircleCI config) can pass a parameter so I just duplicated that block with a different `--distro`.

For the stdeb command itself, we used to just call `bdist_deb`, which took care of running `sdist_src` itself. However, `bdist_deb` can't pass additional options/arguments on to the subcommands it calls. So, it looks like the thing you can do is append multiple commands in the same argv. I did this, and successfully build a deb, but this needs to run through CI and be checked by a dev.

### DIFF
--- a/install/linux/build-deb.py
+++ b/install/linux/build-deb.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+
+import argparse
 import inspect
 import os
 import shutil
@@ -21,6 +23,22 @@ def run(cmd):
 
 
 def main():
+    parser = argparse.ArgumentParser(
+        prog=sys.argv[0],
+        description="Dev script for building Dangerzone debs",
+    )
+    # FIXME: The name of the distro is important, as it can help users who are upgrading
+    # from a distro version to another. If we *do* need to provide a name at some point,
+    # here's a suggestion on how we should tackle naming:
+    #
+    # https://github.com/freedomofpress/dangerzone/pull/322#issuecomment-1428665162
+    parser.add_argument(
+        "--distro",
+        required=False,
+        help="The name of the Debian-based distro",
+    )
+    args = parser.parse_args()
+
     dist_path = os.path.join(root, "dist")
     deb_dist_path = os.path.join(root, "deb_dist")
 
@@ -31,13 +49,31 @@ def main():
         shutil.rmtree(deb_dist_path)
 
     print("* Building DEB package")
-    # This command also builds the Debian source package, and then creates the DEB
-    # package, meaning that we don't need to run `sdist_dsc` as well.
-    run(["python3", "setup.py", "--command-packages=stdeb.command", "bdist_deb"])
+    # NOTE: This command first builds the Debian source package, and then creates the
+    # final DEB package. We could simply call `bdist_deb`, which performs `sdist_dsc`
+    # implicitly, but we wouldn't be able to pass the Debian version argument. Because
+    # we do this in a single invocation though, there's no performance cost.
+    if args.distro is None:
+        deb_ver_args = ()
+        deb_ver = "1"
+    else:
+        deb_ver_args = ("--debian-version", args.distro)
+        deb_ver = args.distro
+
+    run(
+        [
+            "python3",
+            "setup.py",
+            "--command-packages=stdeb.command",
+            "sdist_dsc",
+            *deb_ver_args,
+            "bdist_deb",
+        ]
+    )
 
     print("")
     print("* To install run:")
-    print("sudo dpkg -i deb_dist/dangerzone_{}-1_all.deb".format(version))
+    print(f"sudo dpkg -i deb_dist/dangerzone_{version}-{deb_ver}_all.deb")
 
 
 if __name__ == "__main__":


apyrgio:
Hi Maeve, thanks a lot for this PR. I managed to test it locally and it works :tada:. I had to fix some minor Dangerzone issues, but they are orthogonal to this PR.

Now, digging deeper, I have some comments below:

1. The reason all Debian packages had the same checksums (same applies to Ubuntu), was because we actually built them **once**, for a random Debian / Ubuntu version. This was an error in our CI configuration, and we will fix it for the next release. This means that we will have more options than just `{debian,ubuntu}`, probably `{bookworm,bullseye,kinetic,focal,jammy}`.
   * My suggestion here is to let the system auto-detect the distribution version from the environment (`/etc/os-release`). We have a strict subset of platforms we support, so we don't need to deal with exotic release files.
   * I can offer a fixup for this, if you want.
2. I'm trying to predict situations that this may affect our end users. I think I have one, that is not uncommon: Suppose that a user on Ubuntu Jammy (22.04) upgrades to Ubuntu Kinetic (22.10). In that case, they should also change their APT sources to point to our Ubuntu Kinetic repo. They should therefore move from `0.4.1-1.jammy` to `0.4.1-1.kinetic`, which is a valid upgrade. The same does not apply to Debian though, as they will go from, e.g., `0.4.1-1.bullseye` to `0.4.1-1.bookworm`. This is considered as a downgrade by Debian. Any thoughts on whether this is an actual issue, or how we can tackle this?
    * One suggestion could be to append a number in the `--debian-version` of such distros.
3. We still need to find a solution for the 0.4.0 packages, where this fix won't be applied. I see two solutions there:
    a. Check whether we can build a single Dangerzone `.deb` and install it in all of our supported distros.
    b. Cherry-pick this commit to a branch of `v0.4.0`, and build packages from there.
    
    Any better idea?

maeve-fpf:
Thanks! Sorry for the delay.

1. We can expand the list of possible values for `--distro`, or even make it freeform; I don't think that's a problem. I'm not sure what the correct thing do in that case for the CircleCI config is, as I think we just want to make sure it runs successfully once on Debian and once on Ubuntu without caring about the particular distro.
2. I think this needs to be tested: if dist-upgrading a Debian system, the user will need to edit the sources.list.d entry for the repo, and I don't *think* an unavailable version will be replaced with an available-but-downgrade version, but it might be. Specifying the version numerically seems fine. However, since the contents of the packages are functionally identical, maybe it doesn't matter as long as the existing version still runs; the next release after the user changes their sources.list will be installed normally.
3. I'm in favor of rebuilding, as I don't think we can get around trying to include existing same-filename-but-different-checksum packages. For testing the LFS repos currently, I have removed previous revisions and only kept the most recent. Can we do that under normal circumstances, or is there some value to keeping old versions in the pool?

legoktm:
> They should therefore move from 0.4.1-1.jammy to 0.4.1-1.kinetic, which is a valid upgrade. The same does not apply to Debian though, as they will go from, e.g., 0.4.1-1.bullseye to 0.4.1-1.bookworm. This is considered as a downgrade by Debian. Any thoughts on whether this is an actual issue, or how we can tackle this?

I would recommend using the standard Ubuntu (PPA) and Debian version numbering format unless there's a good reason to diverge.

For Ubuntu that's something like 0.4.1-1\~jammy1. For Debian it's 0.4.1-1~deb10u1 (where 10 is the Debian version).

apyrgio:
Updating this PR based on our latest understanding:

I checked whether we can build a single .deb for Dangerzone, as we've been practically doing this whole time, and install it in all of our supported flavors. I've produced .debs for our most bleeding edge (Debian Bookworm) and most stable (Ubuntu Focal) distros, installed them interchangeably, and they work.

There are three reasons why they work:

1. The packages do not have any compiled files for a specific Python version, so the installed files are exactly the same.
2. The dependencies do not differ based on the flavor, so the control file is exactly the same.
3. The .deb format underwent a major change (zstd compression) in dpkg version 1.19, but all flavors have dpkg >= 1.19. So, the Debian package structure is also the same.

What's different between flavors? One thing, the entry points (`/usr/bin/dangerzone` / `/usr/bin/dangerzone-cli`). Due to different setuptools versions, they produce slightly different entrypoints. From what I've seen, they do so for compatibility between `easy_install` and `pip`, and they don't affect typical Debian installations.

Based on the above, it seems that we won't need the `--distro` option in the near future, which is why I propose making it optional. However, it's important to include this option in our code, in case the above assessment proves to be wrong and we **do** need separate packages for each distro.
 
> I would recommend using the standard Ubuntu (PPA) and Debian version numbering format unless there's a good reason to diverge.
> 
> For Ubuntu that's something like `0.4.1-1~jammy1`. For Debian it's `0.4.1-1~deb10u1` (where 10 is the Debian version).

Thanks for the tip Kunal, this helps a lot. I have also pointed to this discussion from the code, so that we can use your tip, if we choose to revisit this decision.

apyrgio (approved):

Issue #318, opened 2023-01-25
Non-zero Exit Status When at Least one Document Fails


deeplow:
The introduction of multi-document conversion introduced a bug where exceptions were suppressed. https://github.com/freedomofpress/dangerzone/pull/313 makes them visible, but we should also make the whole program exit at the end of all conversions with non-zero, indicating some failure.

This is important to ensure our tests fail when a document fails. Currently they pass, regardless of the success of the conversion.

PR #329, opened 2023-01-31
GUI: exit with 1 when some conversion failed

deeplow:
Fixes: #318

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -111,8 +111,13 @@ class MainWindow(QtWidgets.QMainWindow):
             ok_text="Abort conversions",
         )
         converting_docs = self.dangerzone.get_converting_documents()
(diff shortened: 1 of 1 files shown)


dangerzone/gui/main_window.py
```diff
@@ -112,7 +117,7 @@ def closeEvent(self, e: QtGui.QCloseEvent) -> None:
             else:
                 e.accept()
 
-        self.dangerzone.app.quit()
+        self.dangerzone.app.exit(0)
```
apyrgio, line 120: I'd personally change the return code to 2, since it's not a successful run. Other than that, LGTM.

apyrgio (approved):

dangerzone/gui/main_window.py
deeplow: Sounds good. I have now changed it.

PR #332, opened 2023-02-02
Introduce proportional timeouts

### LINKED ISSUE #306: Error in separating document into pages , timeout 120s
I am new to coding and stuff so i dont know how can i increase the timeout delay of 'danger-zone' process as i saw another thread discussing this issue saying that a 1000 page pdf would need about 500 seconds or smth like that , so kindly can someone help me with it?
btw os: Win11

### LINKED ISSUE #314: `pdfunite` times out on M1 Pro
I tried to use Dangerzone to convert a 800 page document to a PDF. Everything went fine up until the software tried to merge the PDF: it converts it to pixels and the pixels to PDF, but fails at merging the PDF. The error message I get is this: "error merging pages into document, pdfunite timed out after 120 seconds". I am running the latest ARM image of Docker and the latest version of Dangerzone. Converting a three page document works just fine. Any suggestions?

### LINKED ISSUE #327: Remove timeouts that are not proportional to the pages
The Dangerzone container currently has two types of timeouts:

https://github.com/freedomofpress/dangerzone/blob/56b5b98f1e8b9cb0c6c41d9caf594dd1eb350d79/container/dangerzone.py#L27-L31

* `DEFAULT_TIMEOUT` is a timeout that applies to any command that runs in a Dangerzone container.
* `COMPRESSION_TIMEOUT` is a timeout that applies to a specific command (compressing a PDF) that is *proportional* to the number of pages in the PDF.

The second type of timeout should be our end goal. The reason is that the first type of timeout is page-agnostic, meaning that operations on large PDFs may surpass it. There are several issues that exemplify this problem:

* #70 
* #314 
* #306

The reason we're still stuck with the first type of timeout is that (until recently) we didn't have a way to know the number of pages of a PDF beforehand. Now that we do, we can do the following:

1. Grab the number of pages in a document as soon as we can.
    * Generally, this means after we have a PDF document we can work with.
    * This means that operations on files that are not PDFs (e.g., `.docx`, `.xlsx`) will have to use a page-agnostic timeout.
    * For large images or non-PDF documents, maybe we can have a per-MB timeout.
2. Settle on a per-page timeout that is sufficient for all of the PDF operations that we use. Currently this seems to be 10 seconds.
3. Replace all global timeouts with a pages x per page timeout.
4. Add a CLI command that will allow users to change this proportional timeout to something bigger. This should act as an escape hatch, so that users won't need to re-build the image, if they need to change it.
    * Since it's an escape hatch, it will not be visible in the UI for now. The end goal after all is to have something sensible that all users can work with.
5. Add a CI test that converts a very large PDF (e.g, 1000 pages). The rationale is that we've missed these cases, because for the time being we didn't have such tests. Bonus points if it fails in the `main` branch but works on our new implementation.

apyrgio:
This is a PR that introduces proportional timeouts for the various Dangerzone commands that we use for doc conversions.

Taken from the respective commit:

> 
> Previously, we had a single timeout for each command (120 seconds),
> which didn't scale well either with the number of pages in a document,
> or with the size of the document.
> 
> In this commit, we look into each operation, and we're trying to figure
> out the following:
> 
> 1. What's the number of pages we will operate on?
> 2. How large is the document?
> 
> Knowing the above, we can break down a command into multiple operations,
> at least conceptually. Having a number of operations and a sane timeout
> value per operation (10 seconds), we can multiply those and reach to a
> timeout that fits the command better.

Fixes #327
Fixes #314
Fixes #306 

Note: The only thing missing for the PR is finding a large enough dataset in order to test this code against, and make it part of our CI tests.

### DIFF
--- a/container/dangerzone.py
+++ b/container/dangerzone.py
@@ -121,6 +121,9 @@ class DangerzoneConverter:
         * Documents with lots of pages, but small file size.
         * Single images with large file size.
         """
+        if not int(os.environ.get("ENABLE_TIMEOUTS", 1)):
+            return None
+
         # Do not have timeouts lower than 10 seconds, if the file size is small, since
         # we need to take into account the program's startup time as well.
         timeout = max(TIMEOUT_PER_MB * size, 10)
--- a/container/dangerzone.py
+++ b/container/dangerzone.py
@@ -25,11 +25,8 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import magic
 
-# timeout in seconds for any single subprocess
-DEFAULT_TIMEOUT: float = 120
-
-# timeout in seconds for compressing a single page of the final document
-COMPRESSION_TIMEOUT: float = 10
+TIMEOUT_PER_PAGE: float = 10  # (seconds)
+TIMEOUT_PER_MB: float = 10  # (seconds)
 
 
 async def read_stream(sr: asyncio.StreamReader, callback: Callable = None) -> bytes:
@@ -60,7 +57,7 @@ async def run_command(
     *,
     error_message: str,
     timeout_message: str,
-    timeout: float = DEFAULT_TIMEOUT,
+    timeout: Optional[float],
     stdout_callback: Callable = None,
     stderr_callback: Callable = None,
 ) -> Tuple[bytes, bytes]:
@@ -107,6 +104,29 @@ async def run_command(
 class DangerzoneConverter:
     def __init__(self) -> None:
         self.percentage: float = 0.0
+
+    def calculate_timeout(
+        self, size: float, pages: Optional[float] = None
+    ) -> Optional[float]:
+        """Calculate the timeout for a command.
+
+        The timeout calculation takes two factors in mind:
+
+        1. The size (in MiBs) of the dataset (document, multiple pages).
+        2. The number of pages in the dataset.
+
+        It then calculates proportional timeout values based on the above, and keeps the
+        large one.  This way, we can handle several corner cases:
+
+        * Documents with lots of pages, but small file size.
+        * Single images with large file size.
+        """
+        # Do not have timeouts lower than 10 seconds, if the file size is small, since
+        # we need to take into account the program's startup time as well.
+        timeout = max(TIMEOUT_PER_MB * size, 10)
+        if pages:
+            timeout = max(timeout, TIMEOUT_PER_PAGE * pages)
+        return timeout
 
     async def document_to_pixels(self) -> None:
 
@@ -187,6 +207,14 @@ class DangerzoneConverter:
         if mime_type not in conversions:
             raise ValueError("The document format is not supported")
 
+        # Get file size (in MiB)
+        size = os.path.getsize("/tmp/input_file") / 1024**2
+
+        # Calculate timeout for the first few file operations. The difference with the
+        # subsequent ones is that we don't know the number of pages, before we have a
+        # PDF at hand, so we rely on size heuristics.
+        timeout = self.calculate_timeout(size)
+
         # Convert input document to PDF
         conversion = conversions[mime_type]
         if conversion["type"] is None:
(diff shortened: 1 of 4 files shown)


container/dangerzone.py
```diff
@@ -24,83 +26,92 @@
 
 import magic
 
-# timeout in seconds for any single subprocess
-DEFAULT_TIMEOUT: float = 120
+# The size threshold (in MiB) over which we assume that it takes more than one
+# "operation" on a file.
+SIZE_THRESHOLD: float = 1.0
```
deeplow, line 31: I had to read this a couple of times to understand what it meant. I think it would be easier to understand if we made it `timeout_per_mb`:

```suggestion
# The size threshold (in MiB) over which we assume that it takes more than one
# "operation" on a file.
TIMEOUT_PER_MB: float = 1.0
```

But it's not too critical. We can leave it as it.

deeplow (commented):
Two minor things in terms of naming stuff. Other than that, LGTM.

container/dangerzone.py
apyrgio: I agree, it's a bad explanation. Regarding `TIMEOUT_PER_MB = 1.0`, I think that people who are reading it will expect this to be a 1 second timeout per 1 MiB, but this is not the case here. How about `OPS_PER_MB`, since we already use this term? This fits nicely with this calculation (`ops = math.ceil(size / OPS_PER_MB)`).

dangerzone/cli.py
```diff
@@ -36,6 +36,16 @@ def print_header(s: str) -> None:
 @click.option(
     "--unsafe-dummy-conversion", "dummy_conversion", flag_value=True, hidden=True
 )
+@click.option(
+    "--op-timeout",
+    default=Container.OP_TIMEOUT,
```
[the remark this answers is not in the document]
deeplow: I quite like that approach. It's a pressure-relief valve. If the timeouts are not working for people, then the issue with the timeouts is for everyone and that needs to be fixed in the next version.

And there's no need to over-complicate things.

container/dangerzone.py
apyrgio: I have this feeling that I won't remember what op means in 6 months from now as well. Ok, let's make it simpler then:

```suggestion
# The timeout (in seconds) for a single page. If a command spans a number of
# pages, this number will be multiplied by the number of pages.
TIMEOUT_PER_PAGE: float = 10
# The timeout (in seconds) per MiB. If a command ingests files that are larger
# than 1 MiB, this number will be multiplied by their MiBs.
TIMEOUT_PER_MB: float = 1.0
# NOTE: For each operation on a file (be it a single page, multiple pages or a
# whole document) we calculate these two timeout values, and we use the
# largest one.
```

Is it clearer now?

deeplow: > Is it clearer now?

Way clearer. Thank you! So clear that I think we don't even need the long comment. I'd remove most of the comments since they are self-explanatory

```diff
-# The timeout (in seconds) for a single page. If a command spans a number of
-# pages, this number will be multiplied by the number of pages.
-TIMEOUT_PER_PAGE: float = 10
+TIMEOUT_PER_PAGE: float = 10  # (seconds)
-# The timeout (in seconds) per MiB. If a command ingests files that are larger
-# than 1 MiB, this number will be multiplied by their MiBs.
-TIMEOUT_PER_MB: float = 1.0
+TIMEOUT_PER_MB: float = 1.0  # (seconds)
-# NOTE: For each operation on a file (be it a single page, multiple pages or a
-# whole document) we calculate these two timeout values, and we use the
-# largest one.
```

The last note could possibly also go, since it's likely to be out of sync with the code if we ever change that (we won't remember there's a comment referencing it). Maybe moving it closer to the `max()`? (if the code isn't clear enough already).

apyrgio:
I've updated the PR based on the things we discussed. Most of those couldn't be pushed as fixups, so I had to force-push the branch.

deeplow (approved):
The new code LGTM.

PR #333, opened 2023-02-06
Fix failing CI by using PySide6 in all of our dev environments

### LINKED ISSUE #294: Cannot create build environment in Fedora 37
As is, we can't currently create a build environment for Fedora 37 because:

* Fedora has Python 3.11 as default.
* PySide2 does not yet support Python 3.11, and may take some time to support it. See:
   - https://pypi.org/project/PySide2/
   - https://wiki.qt.io/Qt_for_Python_Development_Notes#04._August_2022:
     > 3.11 compatibility will not be backported to 5.15.x releases for x < 11

This means that we can't create a build environment, by running `poetry install` and installing PySide2 from PyPI.

Interestingly, the Fedora maintainers have managed to make PySide2 run on Fedora 37, meaning that end users who install Dangerzone will simply download PySide2 from the official repos, and not get affected.

Still, this impacts development, and its one more push for #164

_Originally posted by @apyrgio in https://github.com/freedomofpress/dangerzone/issues/263#issuecomment-1332834637_

### LINKED ISSUE #330: CI Broken: Debian Bookworm and Fedora-37 can't install PySide2 via Pypi
In both Fedora 37 and Debian Bookworm, doing [`pip install` fails](https://app.circleci.com/pipelines/github/freedomofpress/dangerzone/915/workflows/4c781be6-f871-47ea-be30-048246fe2c71/jobs/6643). The reason for this is because they both have python 3.11 which our project doesn't support via poetry.

apyrgio:
Our CI currently fails for several reasons, the main one being that more and more environments (Fedora 37, Debian Bookworm) use Python 3.11, which is incompatible with PySide2 from PyPI.

In this PR, we remove PySide2 as a dev dependency, and we use PySide6 for all of our environments. The Debian/Fedora packages for Dangerzone will still require PySide2, but that should be the last reference to PySide2 remaining, until we fuly complete the Qt6 transition.

Fixes #294
Fixes #330

_(This PR depends on top of #296)_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,10 +151,10 @@ jobs:
       - run:
           name: Install poetry dependencies
           command: |
-            # Remove this pin once the upstream Poetry issue is fixed:
+            sudo pip3 install poetry
+            # This flag is important, due to an open upstream Poetry issue:
             # https://github.com/python-poetry/poetry/issues/7184
-            sudo pip3 install poetry==1.2.2
-            poetry install
+            poetry install --no-ansi
       - run:
           name: Prepare cache directory
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,6 +156,10 @@ jobs:
             # https://github.com/python-poetry/poetry/issues/7184
             poetry install --no-ansi
       - run:
+          name: Install test dependencies
+          command: |
+            sudo apt-get install -y libqt5gui5 --no-install-recommends
+      - run:
           name: Prepare cache directory
           command: |
             sudo mkdir -p /caches
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -223,6 +223,32 @@ jobs:
             ./dev_scripts/env.py --distro ubuntu --version 22.04 run --dev \
                 bash -c 'cd dangerzone; poetry run make test'
 
+  ci-fedora-37:
+    machine:
+      image: ubuntu-2004:202111-01
+    steps:
+      - checkout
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -151,10 +151,14 @@ jobs:
...
+            sudo pip3 install poetry
+            # This flag is important, due to an open upstream Poetry issue:
             # https://github.com/python-poetry/poetry/issues/7184
-            sudo pip3 install poetry==1.2.2
-            poetry install
+            poetry install --no-ansi
+      - run:
+          name: Install test dependencies
+          command: |
+            sudo apt-get install -y libqt5gui5 --no-install-recommends
```
deeplow, line 161: Why is this one needed? Is this for testing the built `.deb` (which will use PySide2)? If so, isn't this already a depednency of the produced `.deb`?

deeplow:
Other than the above comment, this LGTM. :heavy_check_mark:

.circleci/config.yml
apyrgio: There is no built `.deb` in that step. We are in a development environment in this case, so we install dependencies via Poetry. The problem here is that PySide needs to load some `.so` libraries, which must be provided via a system package.

Feel free to check the commit message as well for this addition. If it doesn't adequately explain why, I can extend it, and maybe add a comment there.

deeplow: Sounds good. I should have checked it.

deeplow (approved):

deeplow (approved):

Issue #50, opened 2020-03-11
Provide arm64 Docker Image


kylerankin:
I am looking to provide dangerzone as a default application on the Librem 5 phone hardware (running PureOS, which is based off of Debian Buster). As the dangerzone package is architecture-independent I have been able to install the buster Debian package and run it on the phone (it doesn't quite fit on the default small screen resolution but I'm working on a PR to help with that).

Unfortunately the only docker image that is provided is for amd64. This causes docker to fail on platforms that don't support amd64. Please provide an alternate docker image built for arm64 platforms.

micahflee:
The containers are built automatically using CircleCI. Here's the state of CircleCI's arm64 support: https://ideas.circleci.com/ideas/CCI-I-900

kylerankin:
Hmm, looks like they've been sitting on the issue for a year. That doesn't make the prospect of arm64 images too promising.

andrewcrook:
Would be useful for when Apple Silicon devices with Big Sur come out and  Docker for Apple silicon ( although it might run in a VM with a guest ARM Linux like the current version for Intel Macs uses Linux)

gmarmstrong:
CircleCI now [supports ARM as resource class](https://circleci.com/docs/2.0/using-arm/), but not with their Docker executor. The issue is [tracked here](https://circleci.canny.io/cloud-feature-requests/p/support-arm-resource-class-on-docker-executor).

gmarmstrong:
Building a native image on an ARM machine (by removing the `--platform linux/amd64` argument from `container.py` and `build-image.sh` and then rebuilding the image) gives a remarkable improvement to document conversion speed (over 3 times faster):

**ARM image on an ARM machine (Apple Silicon/M1) (~5 seconds):**

```
(dangerzone-H9rmGDxH-py3.10) guthrie@macbook dangerzone % time ./dev_scripts/dangerzone-cli test_docs/sample.pdf                                                     typing
╭──────────────────────────╮
│           ▄██▄           │
│          ██████          │
│         ███▀▀▀██         │
│        ███   ████        │
│       ███   ██████       │
│      ███   ▀▀▀▀████      │
│     ███████  ▄██████     │
│    ███████ ▄█████████    │
│   ████████████████████   │
│    ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀    │
│                          │
│    Dangerzone v0.3.1     │
│ https://dangerzone.rocks │
[...]

PR #337, opened 2023-02-09
Build arm64 docker image for arm-based Macs

deeplow:
Partial fix for #50

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -188,11 +188,9 @@ class Container(IsolationProvider):
         container_runtime = self.get_runtime()
 
         if self.get_runtime_name() == "podman":
-            platform_args = []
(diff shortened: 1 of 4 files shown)


apyrgio:
I think we don't even need to specify the platform. By default, Docker should build the image natively, which is what we want. So, I'd propose completely removing the `--platform` argument. Am I missing something?

deeplow:
You're right. Not need for `--platform` at all.

apyrgio:
> You're right. Not need for --platform at all.

I still see it in `install/macos/build-image.sh`. Am I missing something?

deeplow:
> I still see it in install/macos/build-image.sh. Am I missing something?

Whoops! You're right!

apyrgio (approved):
Looks good, feel free to merge.

PR #338, opened 2023-02-13
Container: fix issues pdftoppm-related (timeout & errors)

deeplow:
After running the documents over a large test set some issues related to `pdftoppm` came to light. This provides fixes to some of those issues.

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -8,6 +8,7 @@ RUN apk -U upgrade && \
     libreoffice \
     openjdk8 \
     poppler-utils \
+    poppler-data \
     python3 \
     py3-magic \
     tesseract-ocr \
--- a/container/dangerzone.py
+++ b/container/dangerzone.py
@@ -305,7 +305,10 @@ class DangerzoneConverter:
                 num_pages = int(num_pages_str)
                 page = int(page_str)
             except ValueError as e:
-                raise RuntimeError("Conversion from PDF to PPM failed") from e
+                # Ignore all non-progress related output, since pdftoppm sends
+                # everything to stderr and thus, errors can't be distinguished
+                # easily. We rely instead on the exit code.
+                return
 
             percentage_per_page = 45.0 / num_pages
             self.percentage += percentage_per_page


deeplow:
I found `pdftoppm` displaying another error:

> Internal Error: xref num 4 not found but needed, try to reconstruct<0a>

I guess the best approach is to rely on its exit code, rather than any output it prints.

container/Dockerfile
```diff
@@ -74,7 +74,15 @@ RUN apk -U upgrade && \
...
-    tesseract-ocr-data-vie
+    tesseract-ocr-data-vie \
+    poppler-data
+
+# XXX waiting on fix upstream bug with poppler utils
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850163#10
+RUN ln -s /usr/share/poppler/cidToUnicode   /cMap && \
+    ln -s /usr/share/poppler/cidToUnicode   /cidToUnicode && \
+    ln -s /usr/share/poppler/nameToUnicode  /nameToUnicode && \
+    ln -s /usr/share/poppler/unicodeMap     /unicodeMap
```
apyrgio, line 85: I have a reservation for these lines. I see in [this issue](https://bitbucket.org/merarischroeder/poppler-for-windows/issues/5/missing-language-pack-for-adobe-japan1) that the solution to this problem is to install the `poppler-data` package. Also, I see that we erroneously symlink `/usr/share/poppler/cidToUnicode` to `/cMap`.

So, are these lines actually necessary, or are we covered by installing `poppler-data`?

container/dangerzone.py
```diff
@@ -261,7 +261,14 @@ def pdftoppm_progress_callback(line: bytes) -> None:
                 num_pages = int(num_pages_str)
                 page = int(page_str)
             except ValueError as e:
-                raise RuntimeError("Conversion from PDF to PPM failed") from e
+                if line.decode().startswith(
+                    "Syntax Warning"
+                ) or line.decode().startswith("Syntax Error"):
+                    # Ignore syntax warnings and errors.
+                    # pdftoppm produces a valid document regardless of these errors
+                    return
```
apyrgio, line 269: I think you've mentioned it somewhere, but I don't see the comment now. Since `pdftoppm` may print arbitrary error messages in stderr, we should just discard such lines (they will be included in the logs after all), and simply rely on its return code.

apyrgio (changes requested):
Minor, but in 0909a59d8974a5e782e219f01db1c839ddcca963, we should change "to have document" -> "to have a document".

container/dangerzone.py
deeplow: I agree.

container/Dockerfile
deeplow: Good catch. You're right. We don't need that. I tested with [this document](https://github.com/deeplow/dangerzone-test-set/blob/main/%25e4%25b8%258d%25e5%2590%258c%25e6%25b5%25b7%25e8%25a1%25a8%25e9%259d%25a2%25e6%25b8%25a9%25e5%25ba%25a6%25e5%25af%25b9%25e5%258d%2597%25e6%25b5%25b7%25e5%258f%25b0%25e9%25a3%258e%25e2%2580%259c%25e6%259d%259c%25e9%25b9%2583%25e2%2580%259d%25e7%259a%2584%25e5%25bd%25b1%25e5%2593%258d%25e8%25af%2595%25e9%25aa%258c.pdf_uncompressed.pdf) and it does not have issues like `Missing language pack for`.

container/dangerzone.py
deeplow: Resolved via [cf0c724](https://github.com/freedomofpress/dangerzone/pull/338/commits/cf0c724e6bab5e3cf48f22b52c12d0908df89aa8).

container/Dockerfile
deeplow: Resolved via [6a99bf0](https://github.com/freedomofpress/dangerzone/pull/338/commits/6a99bf0903cf6972455011a12fd6fa4d74ba4cf1).

apyrgio (approved):
I saw the updates and they look good to me. Feel free to merge, but make sure to drop the 803cbbe6f6c504407591e95df8f59506fa4c21e6 commit, as it will conflict with `main`.

PR #347, opened 2023-02-21
Fix Windows .msi building instructions

deeplow:
Building the installer on windows was failing. This fixes that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,13 @@ jobs:
       - run: poetry install
       - name: Run CLI tests
         run: poetry run make test
+      # Taken from: https://github.com/orgs/community/discussions/27149#discussioncomment-3254829
+      - name: Set path for candle and light
+        run: echo "C:\Program Files (x86)\WiX Toolset v3.11\bin" >> $GITHUB_PATH
+        shell: bash
+      - name: Build the MSI installer
+        # NOTE: This also builds the .exe internally.
+        run: poetry run .\install\windows\build-app.bat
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -20,6 +20,14 @@ jobs:
...
       - name: Run CLI tests
         run: poetry run make test
+      - name: Set path for candle and light
+        env:
+          WIX_PATH: ${{ env.wix }}
+        run: echo "$env:WIX_PATH\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Build .exe
+        run: poetry run python .\setup-windows.py build
+      - name: Build the MSI installer
+        run: poetry run .\install\windows\build-app.bat
```
deeplow, line 30: I believe this step already calls `python .\setup-windows.py build` but it probably won't run successfully because it expects some signing stuff. I had to comment out some things to make it work locally.

deeplow: Oh. Never mind. I see it's a WIP. Testing CI stuff is complicated.

apyrgio (approved):
Looks good. Since I don't have a Windows machine, I couldn't test it out locally. BUT! Why not let GitHub do it for me. So, I went ahead and added a commit in your branch that builds the MSI installer in the GitHub actions. If you approve this commit, feel free to resign it and merge it.

deeplow:
The few extra commits was just me playing to see if the msi failed to produce it would indeed fail the step. I reverted them in the end and rebased on top of main.

PR #348, opened 2023-02-21
Add instructions for publishing/installing Debian packages to/from packages.freedom.press

### LINKED ISSUE #291: Use packages.freedom.press for Debian packages
FPF has a public [Git-LFS repo](https://github.com/freedomofpress/apt-tools-prod) that is linked with an [APT repository](https://packages.freedom.press/apt-tools-prod) hosted by FPF. This repo can be used instead of the PackageCloud repo where the Debian packages currently live.

Subtasks:

- [x] Build the Debian packages for 0.4.0 locally, for all the Debian-like distributions that we support.
- [x] Update the Git-LFS repo with these packages, and sign them with the GPG developer key.
- [x] Update the Dangerzone release instructions to include this flow.
- [x] Update the Dangerzone install instructions for end-users to point to the FPF's APT repo.
- [x] Remove the CircleCI task that updates PackageCloud with new Debian packages.

apyrgio:
This is a draft PR for updating our Debian/Ubuntu instructions, in preparation for the PackageCloud -> packages.freedom.press transition. It's draft for the following reason:

1. It depends on bringing our APT repo to a state that users can use. This means having a proper GPG key, submitting it somewhere that users can download, and testing our instructions in all platforms.

Outdated reasons:

2. The instructions may change if we use [`extrepo`](https://grep.be/blog/en/computer/debian/Announcing_extrepo/) as an easy way for users to add our repo.
    * Update: We _can_ use `extrepo`, once we have a proper key, but it will not be a blocker for the release.
3. It is based on a PR (#346) that will be merged soon.
   * Update: Merged.

Fixes #291

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -38,10 +38,25 @@ Dangerzone is available for:
   ```
 </details>
 
-Add our repository following [these instructions](https://packagecloud.io/firstlookmedia/code/install#manual-deb), or by running this script:
+Add our repository following these instructions:
 
+Download the GPG key for the repo:
+
+```sh
+gpg --keyserver hkps://keys.openpgp.org \
+    --no-default-keyring --keyring ./fpf-apt-tools-archive-keyring.gpg \
+    --recv-keys "DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281"
+sudo mkdir -p /etc/apt/keyrings/
+sudo mv fpf-apt-tools-archive-keyring.gpg /etc/apt/keyrings
 ```
-curl -s https://packagecloud.io/install/repositories/firstlookmedia/code/script.deb.sh | sudo bash
+
+Add the URL of the repo in your APT sources:
+
+```sh
+source /etc/os-release
+echo deb [signed-by=/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg] \
+    https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main \
+    | sudo tee /etc/apt/sources.list.d/fpf-apt-tools.list
 ```
 
 Install Dangerzone:
@@ -50,6 +65,22 @@ Install Dangerzone:
 sudo apt update
 sudo apt install -y dangerzone
 ```
+
+<details>
+  <summary><i>:memo: Expand this section for a security notice on third-party Debian repos</i></summary>
+  </br>
+
+  This section follows the official instructions on configuring [third-party
+  Debian repos](https://wiki.debian.org/DebianRepository/UseThirdParty).
+
+  To mitigate a class of attacks against our APT repo (e.g., injecting packages
+  signed with an attacker key), we add an additional step in our instructions to
+  verify the downloaded GPG key against its fingerprint.
+
+  Aside from these protections, the user needs to be aware that Debian packages
+  run as `root` during the installation phase, so they need to place some trust
+  on our signed Debian packages. This holds for any third-party Debian repo.
+</details>
 
 ### Fedora
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -212,6 +212,40 @@ Rename `Dangerzone.msi` to `Dangerzone-$VERSION.msi`.
 
 ## Linux release
 
+### Debian/Ubuntu
+
+Because the Debian packages do not contain compiled Python code for a specific
+Python version, we can create a single Debian package and use it for all of our
+Debian-based distros.
+
+Create a Debian Bookworm development environment. You can [follow the
+instructions in our build section](https://github.com/freedomofpress/dangerzone/blob/main/BUILD.md#debianubuntu),
+or create your own locally with:
+
+```sh
+./dev_scripts/env.py --distro debian --version bookworm build-dev
+./dev_scripts/env.py --distro debian --version bookworm run --dev bash
+cd dangerzone
+```
+
+Build the latest container:
+
+```sh
+./install/linux/build-image.sh
+```
+
+Create a .deb:
+
+```sh
+./install/linux/build-deb.py
+```
+
+Publish the .deb under `./deb_dist` to the
+[`freedomofpress/apt-tools-prod`](https://github.com/freedomofpress/apt-tools-prod)
+repo, by sending a PR. Follow the instructions in that repo on how to do so.
+
+### Fedora
+
 Linux binaries are automatically built and deployed to repositories when a new tag is pushed.
 
 ### Fedora


apyrgio:
I've rebased this PR on top of `main`, and I've updated its description. The only thing pending is to find out the proper URL for the FPF signing key. Ideally, I'd like to add a GitHub Actions job that verifies that we can build a `.deb` from Debian Bookworm and install it in every other distro. I have tested it manually already, so that's more of a stretch goal.

In the meantime, @deeplow feel free to take a look at it, as it won't change that much.

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
 </details>
 
-Add our repository following [these instructions](https://packagecloud.io/firstlookmedia/code/install#manual-deb), or by running this script:
-
-```
-curl -s https://packagecloud.io/install/repositories/firstlookmedia/code/script.deb.sh | sudo bash
+Add our repository following these instructions:
+
+```sh
+sudo apt-get update
```
deeplow, line 44: I'd say we can assume the user has updated sources.
```suggestion
```

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
+sudo apt-get install -y curl
```
deeplow, line 45: Damn. So Debian has `curl` by default but not `wget` and Ubuntu is the other way around :/

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
+sudo mkdir -p /etc/apt/keyrings/
+curl -S https://packages.freedom.press/apt-tools-prod/fpf-archive-keyring.gpg \
```
deeplow, line 47: Won't we need to `gpg --dearmor` this?

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
+    | sudo tee >/dev/null /etc/apt/keyrings/fpf.gpg
```
deeplow, line 48: Tiny pet peeve of mine, but can we do the redirection at the end?
```suggestion
    | sudo tee /etc/apt/keyrings/fpf.gpg >/dev/null
```

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
+source /etc/os-release
+echo deb [signed-by=/etc/apt/keyrings/fpf.gpg] https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main \
+    | sudo tee /etc/apt/sources.list.d/fpf.list
+cat <<EOF | sudo tee /etc/apt/preferences.d/fpf.pref
+Package: *
+Pin: origin packages.freedom.press
+Pin-Priority: 100
+EOF
```
deeplow, line 56: can you explain a bit the logic behind the setting a priority? I was reading up a bit [on apt priorities](https://manpages.debian.org/buster/apt/apt_preferences.5.en.html) but I don't fully get yet what it mitigates here.

INSTALL.md
```diff
@@ -38,10 +38,22 @@ Dangerzone is available for:
...
+Add our repository following these instructions:
+
+```sh
+sudo apt-get update
+sudo apt-get install -y curl
+sudo mkdir -p /etc/apt/keyrings/
+curl -S https://packages.freedom.press/apt-tools-prod/fpf-archive-keyring.gpg \
+    | sudo tee >/dev/null /etc/apt/keyrings/fpf.gpg
+source /etc/os-release
+echo deb [signed-by=/etc/apt/keyrings/fpf.gpg] https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main \
```
deeplow, line 50: `${VERSION_CODENAME?}` since all debian versions are in fact the same, is it possible for us to do [like the signal folks](https://signal.org/download/#) and just hardcode here one version and avoid an extra line for the user to run? (the `source` line).

INSTALL.md
deeplow, line 1: We'll probably need to bump the `qa.py` due to changes in this file.

deeplow (changes requested):
Extrepo is the way to go installation instructions are way too complex. In an ideal world, users wouldn't even have to touch a command line to install software on Linux.

I made some proposals to reduce the install code where possible.

INSTALL.md
apyrgio: If you look at `apt-tools-prod`, the key should be in DER format, so no need to dearmor it.

INSTALL.md
apyrgio: Here's a bit more context: https://wiki.debian.org/DebianRepository/UseThirdParty#Standard_pinning

In a nutshell, you don't want an external repo to ship, e.g., their own `libc` and override the official one, because that's a huge security risk. With APT preferences, we can let external repos provide new packages, but have lower priority (100 < 500) for any other package.

INSTALL.md
apyrgio: We discussed this live. This may become an issue if we want to ship different package definitions for other repos (e.g., compiled code, extra dependencies, etc.). I suggest we start conservatively for now.

INSTALL.md
apyrgio: The `INSTALL.md` file is not tracked by our `qa.py`, only parts of `RELEASE.md` and `BUILD.md`.

INSTALL.md
deeplow: I understand. To be honest, I don't think this is a risk worth considering given that there already other ways an attacker controlling the repo can already do on the system. If my understanding is correct `post-install` scripts runs as root. So any attacker-controlled package could already do severe damage on the system without the need to ship a libc.

The line of defense that we have here is the fact that a webserver compromise won't result in malicious package installs since they won't control the signing keys.

Does this analysis make sense?

INSTALL.md
deeplow: Oh. Right. I got confused.

INSTALL.md
apyrgio: It does make sense, and it's a point that a Debian developer also makes: https://blog.jak-linux.org/2021/06/20/migrating-away-apt-key/#pretending-to-be-safer-by-using-signed-by

The Debian community does not have a definitive answer to this, but it [acknowledges this problem](https://wiki.debian.org/DebianRepository/UseThirdParty):

> Note that the procedures documented here aim to prevent a repository from shipping packages that the administrator does not expect that repository to ship. For example, a repo that ships a video game emulator and its mods shouldn't be able to override [libc6](https://packages.debian.org/libc6).
> 
> However, the installation of any single malicious package from a malicious repository can currently undo these protections, for example by running a [MaintainerScripts](https://wiki.debian.org/MaintainerScripts) command to override the configured preferences or by authorizing new OpenPGP certificates. For the purposes of this page, attacks by a package that belongs to a given repository are out of scope. To restrict what an installed package can do, see the larger [UntrustedDebs](https://wiki.debian.org/UntrustedDebs) problem, and particularly [Teams/Dpkg/Spec/DeclarativePackaging](https://wiki.debian.org/Teams/Dpkg/Spec/DeclarativePackaging) for a potential solution.

---

In any case, we can remove this extra instruction, since it doesn't offer much more protection than pinning the GPG key. We can also explain to the user that:

* our instructions follow the [official Debian guide on external repos](https://wiki.debian.org/DebianRepository/UseThirdParty), and
* packages from external repos are installed with admin privileges, which is why we take several precautions to make sure that a malicious third-party cannot infect our packages.freedom.press infrastructure.

deeplow: OK. Let's remove it, then.

apyrgio:
I've updated the instructions with a security notice, and a check for the GPG key.

INSTALL.md
```diff
@@ -38,10 +38,37 @@ Dangerzone is available for:
...
+Download the GPG key for the repo:
+
+```sh
+sudo mkdir -p /etc/apt/keyrings/
+curl -s https://packages.freedom.press/apt-tools-prod/fpf-apt-tools-archive-keyring.gpg \
+    | sudo tee /etc/apt/keyrings/fpf.gpg >/dev/null
 ```
-curl -s https://packagecloud.io/install/repositories/firstlookmedia/code/script.deb.sh | sudo bash
+
+Verify it against its SHA-256 checksum
```
legoktm, line 57: From a user perspective, I don't think providing a checksum here is that reassuring; versus having them run something like `gpg --keyring /etc/apt/keyrings/fpf.gpg --list-keys` and then verify that only a single key is printed, and provide the fingerprint + email here.

apyrgio:
I made two changes to the PR:

1. I improved the key download based on feedback by @deeplow and @legoktm (thanks!). What we now do is that we download the key from the keyserver by providing the full fingerprint, without importing it in the user's keyring. This solves the issue of using an easy-to-change SHA-256 checksum (e.g., due to a bump in the expiration date).
2. I changed the name of the keyring and the APT source to a fuller name (`fpf-apt-tools-archive-keyring.gpg` and `fpf-apt-tools.list` respectively).

INSTALL.md
```diff
@@ -40,35 +40,29 @@ Dangerzone is available for:
 
 Add our repository following these instructions:
 
-Install `curl`, if it's not already available:
+Install `gpg`, if it's not already available:
 
 ```sh
-sudo apt-get install -y curl
+sudo apt-get install -y gpg ca-certificates
```
deeplow, line 46: Nitpick, but can we use the `apt` instead of `apt-get`? It's supposedly higher level (according to the man pages).

deeplow: Also, I was trying to check if these packages weren't already default in debian, but apparently that's not an easy task without Debian itself. @legoktm do you know if these are default in debian?

INSTALL.md
```diff
@@ -40,35 +40,29 @@ Dangerzone is available for:
...
 ```
 
 Download the GPG key for the repo:
 
 ```sh
+gpg --keyserver hkps://keys.openpgp.org \
+    --no-default-keyring --keyring ./fpf-apt-tools-archive-keyring.gpg \
```
deeplow, line 53: Should we save it in `/tmp/fpf-apt-tools-archive-keyring.gpg` to avoid messing up the user's `$PWD` directory?

deeplow (commented):
Minor suggestions. But overall I like the shortness of the current version.

INSTALL.md
deeplow: Thanks! In that case I'd argue for removing this line. But can try to test it via docker as well, just to be sure.

apyrgio: I've originally added this line because, as Kunal mentioned, these two packages are not part of the base system packages. This means that on a stripped down container image, they do not exist (tested).

Worse, if just `gpg` exists, but not `ca-certificates`, this will lead to a cryptic error that took me a bit to debug, when we download keys via the `hkps` protocol.

However, I agree with the sentiment that Dangerzone is a desktop application, therefore almost all users will have `ca-certificates` installed, and `gpg` most probably. So I'll remove this line to shorten the instructions.

Thanks again for the feedback :-)

INSTALL.md
apyrgio: It's generally best to avoid saving sensitive files in a world-writeable directory with a predictable name. Chances are slim, but another low-privileged process can add its own keyring there.

Other ideas I had in mind:

1. Use `sudo gpg` and store the keyring directly under `/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg`. This is short and sweet but also dangerous. I wouldn't trust running `gpg` with `sudo` due to any bugs lurking there.
2. Generate a random filename, save it in an environment variable, and use that name in the `gpg` and `mv` commands. This makes things a bit more complicated to users reading these instructions.
3. Let `gpg` store the keyring under `~/.gnupg/`, by not specifying a path (i.e., remove `./`). See also the [documentation on `--keyring`](https://www.gnupg.org/gph/en/manual/r1574.html). I'm not sure though what are the repercussions of storing a keyring under `~/.gnupg`. I can say that the default keyring is not affected, so that's a good indication.

All in all, I think chances are slim that the user's `$PWD` will be affected, since we immediately move the file to the target path. If we ever have complaints though, we can revisit this decision.

INSTALL.md
apyrgio: I've pushed some changes that address Kunal's comment, so I'm resolving this discussion. Feel free to re-open it.

INSTALL.md
deeplow: Ah. Right. We're moving it anyways. Let's keep it then.

Issue #355, opened 2023-02-28
100 languages in Dangerzone OCR don't work


deeplow:
Version: 0.4.1 (pre-release)

![filipino](https://user-images.githubusercontent.com/47065258/221935277-ca8eca42-035f-48b5-a536-e4480c89c366.png)

```
[INFO] [doc cPaH6f] 50% Converting page 1/3 from pixels to searchable PDF
[ERROR] [doc cPaH6f] 50% Page 1/3 OCR failed
[ERROR] pixels-to-pdf failed
[DEBUG] Marking doc cPaH6f as 'failed'
```

And this is the container output:

```
Error opening data file /usr/share/tessdata/fil.traineddata
Please make sure the TESSDATA_PREFIX environment variable is set to your "tessdata" directory.
Failed loading language 'fil'
Tesseract couldn't load any languages!
Could not initialize tesseract.
```

deeplow:
Turns out this problem is way bigger. When the the container image was switched from ubuntu 18.04 to alpine it lost many languages since they were not available on alpine. It lost about 100 languages and gained 4, two of which are of ancient times. :cry:

deeplow:
Apparently we mistakenly remove Slovenian

PR #356, opened 2023-02-28
Removes Languages from List that were not actually supported

deeplow:
Context in #355.

**Missing:** upgrade path for those who have an invalid language in the settings.


dangerzone/logic.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, isolation_provider: IsolationProvider) -> None:
 
         # Languages supported by tesseract
         with open(get_resource_path("ocr-languages.json"), "r") as f:
-            self.ocr_languages = json.load(f)
+            unsorted_ocr_languages = json.load(f)
+            self.ocr_languages = dict(sorted(unsorted_ocr_languages.items()))
```
apyrgio, line 38: I come from a Python 2 background, so `dict(sorted(...))` seemed dangerous to me, as there are no guarantees that the `dict()` constructor would respect the provided order. Turns out that this is the case on Python 3.7+, so we're good: https://stackoverflow.com/a/613218. Just wanted to point that out.

deeplow: Yes, it also felt a bit odd to me since dictionaries have no order, supposedly. But it did sort by keys. Thanks for looking into it!

apyrgio (approved):

deeplow:
You're right. That would have been an invalid setup anyways.

PR #378, opened 2023-03-27
container: Allow converting more document formats

### LINKED ISSUE #369: Incomplete MIME type support
When Dangerzone first encounters a file, it needs to detect its MIME type, so that it can choose the proper converter. The list of supported mime types (and the associated converters) is the following:

https://github.com/freedomofpress/dangerzone/blob/a33dcfbb519ccea302d7f36cd5d37577f8b14417/container/dangerzone.py#L138-L203

Using Dangerzone on a large set of files, we discovered that there are two MIME types that very common but are not supported:

```
application/zip
application/octet-stream
```

For instance, this file currently fails on Dangerzone: https://github.com/freedomofpress/dangerzone-test-set/blob/4cbf14ac31ac986ced60e83867aac8a6d2d4a81b/all_documents/HTMLImage.odt. For an association between MIME types and file extensions, you can see the following, taken from a list of 200 documents:

```
02_doc_macros_signed_by_attacker_manipulated.odt: application/zip                                                      
02_doc_signed_by_attacker_manipulated2.odt: application/zip                                                            
02_doc_signed_by_attacker_manipulated.odt: application/zip 
02_doc_signed_by_attacker_manipulated_triple.odt: application/zip                                                      
02_doc_signed_by_trusted_person_manipulated.odt: application/zip                                                       
1_page.docx: application/octet-stream                      
82fff64a-0a21-4b09-bbdc-2914a5a150f0.odt: application/zip                                                              
BackgroundImageTest.odt: application/zip                                                                               
CUSTOM.odt: application/zip                                                                                                                                                                                                                    
CVE-2003-0820-1.doc: application/octet-stream                                                                          
CVE-2005-0941-1.doc: application/octet-stream                                                                          
CVE-2006-2389-1.doc: application/octet-stream                                                                          
CVE-2006-3059-1.xls: application/octet-stream                                                                          
CVE-2006-3086-1.xls: application/octet-stream                                                                          
CVE-2006-3493-1.doc: application/octet-stream                                                                          
CVE-2006-3655-1.ppt: application/octet-stream                                                                          
CVE-2006-3656-1.ppt: application/octet-stream                                                                          
CVE-2006-3660-1.ppt: application/octet-stream                                                                          
CVE-2006-5296-1.ppt: application/octet-stream                                                                          
CVE-2006-6561-1.doc: application/octet-stream                                                                                                                                                                                                  
CVE-2006-6628-1.doc: application/octet-stream                                                                          
CVE-2007-0031-1.xls: application/octet-stream                                                                          
CVE-2007-1347-1.doc: application/octet-stream                                                                          
CVE-2007-3490-1.xls: application/octet-stream                                                                          
CVE-2008-2752-1.doc: application/octet-stream                                                                                                                                                                                                  
CVE-2008-2752-2.doc: application/octet-stream                                                                          
CVE-2008-2752-3.doc: application/octet-stream                                                                          
CVE-2008-2752-4.doc: application/octet-stream                                                                          
CVE-2008-4841-1.doc: application/octet-stream                                                                          
CVE-2009-0200-1.doc: application/octet-stream                                                                          
CVE-2009-0201-1.doc: application/octet-stream              
CVE-2009-0259-1.doc: application/octet-stream                                                                          
CVE-2009-3129-1.xls: application/octet-stream                                                                          
CVE-2009-3301-1.doc: application/octet-stream              
CVE-2009-3302-1.doc: application/octet-stream              
CVE-2009-3302-2.doc: application/octet-stream              
CVE-2010-0033-1.ppt: application/octet-stream              
CVE-2010-1245-1.xls: application/octet-stream              
CVE-2010-1246-1.xls: application/octet-stream                                                                          
CVE-2010-1248-1.xls: application/octet-stream                                                                          
CVE-2010-3200-1.doc: application/octet-stream                                                                          
CVE-2011-0105-1.xls: application/octet-stream              
CVE-2011-0978-1.xls: application/octet-stream                                                                          
CVE-2012-4233-1.odt: application/octet-stream                                                                          
CVE-2012-4233-2.odg: application/octet-stream              
CVE-2014-6356-1.doc: application/octet-stream              
CVE-2014-6361.xls: application/octet-stream                
EDB-18952-1.doc: application/octet-stream                                                                                                                                                                                                      
HTMLImage.odt: application/zip  
```

From this list, it's evident that `application/octet-stream` can refer to many file types. `application/zip` refers just to `.odt`, but we can't be definitely sure about that. Ideally then, if a file does not have a known MIME type and instead uses one of those two, we should also check the file extension.

apyrgio:
Remove the association between MIME types and export filters, because
LibreOffice is able to auto-detect them on its own. Instead, ask
LibreOffice to simply convert the document to a .pdf.

This association was cumbersome for yet another reason; there are MIME
types that may be associated with more than one file type. That's why
it's better to let LibreOffice decide the proper filter for the
conversion.

Our current understanding is that this change won't widen our attack
surface for the following reasons:

* The output filters for PDF documents are pretty specific, and we don't
  affect the input filters somehow.
* The default behavior of LibreOffice on Alpine Linux is to disable
  macros.
* We preemptively run LibreOffice in safe mode, to remove hardware
  acceleration and make sure that macros are not invoked as well.

Closes #369

_(must be merged after https://github.com/freedomofpress/dangerzone/pull/377)_

### DIFF
--- a/container/dangerzone.py
+++ b/container/dangerzone.py
@@ -141,57 +141,66 @@ class DangerzoneConverter:
             # .docx
             "application/vnd.openxmlformats-officedocument.wordprocessingml.document": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "writer_pdf_Export",
             },
             # .doc
             "application/msword": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "writer_pdf_Export",
             },
             # .docm
             "application/vnd.ms-word.document.macroEnabled.12": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "writer_pdf_Export",
             },
             # .xlsx
             "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "calc_pdf_Export",
             },
             # .xls
             "application/vnd.ms-excel": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "calc_pdf_Export",
             },
             # .pptx
             "application/vnd.openxmlformats-officedocument.presentationml.presentation": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "impress_pdf_Export",
             },
             # .ppt
             "application/vnd.ms-powerpoint": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "impress_pdf_Export",
             },
             # .odt
             "application/vnd.oasis.opendocument.text": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "writer_pdf_Export",
             },
             # .odg
             "application/vnd.oasis.opendocument.graphics": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "impress_pdf_Export",
             },
             # .odp
             "application/vnd.oasis.opendocument.presentation": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "impress_pdf_Export",
             },
-            # .ops
+            # .ods
             "application/vnd.oasis.opendocument.spreadsheet": {
                 "type": "libreoffice",
-                "libreoffice_output_filter": "calc_pdf_Export",
+            },
+            # .ods / .ots
+            "application/vnd.oasis.opendocument.spreadsheet-template": {
+                "type": "libreoffice",
+            },
+            # .odt / .ott
+            "application/vnd.oasis.opendocument.text-template": {
+                "type": "libreoffice",
+            },
+            # At least .odt, .docx, .odg, .odp, .ods, and .pptx
+            "application/zip": {
+                "type": "libreoffice",
+            },
+            # At least .doc, .docx, .odg, .odp, .odt, .pdf, .ppt, .pptx, .xls, and .xlsx
+            "application/octet-stream": {
+                "type": "libreoffice",
+            },
+            # At least .doc, .ppt, and .xls
+            "application/x-ole-storage": {
+                "type": "libreoffice",
             },
             # .jpg
             "image/jpeg": {"type": "convert"},
@@ -230,7 +239,7 @@ class DangerzoneConverter:
                 "libreoffice",
                 "--headless",
                 "--convert-to",
-                f"pdf:{conversion['libreoffice_output_filter']}",
+                "pdf",
                 "--outdir",
                 "/tmp",
                 "/tmp/input_file",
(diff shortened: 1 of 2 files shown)


container/dangerzone.py
```diff
@@ -228,8 +238,9 @@ async def document_to_pixels(self) -> None:
             args = [
                 "libreoffice",
                 "--headless",
+                "--safe-mode",
```
deeplow, line 241: > * We can start `libreoffice` with the `--safe-mode` flag, which I think (again, in my system) opens documents with Macro Security Level High.

This is one of the cases where terminology can be confusing. Here they're talking about safety and not security. This option exists for [recovering from borked instalations](https://help.libreoffice.org/latest/en-US/text/shared/optionen/macrosecurity_sl.html). So it won't actually do what we want.

container/dangerzone.py
```diff
@@ -133,64 +135,72 @@ def calculate_timeout(
         return timeout
 
     async def document_to_pixels(self) -> None:
-
         conversions: Dict[str, Dict[str, Optional[str]]] = {
             # .pdf
             "application/pdf": {"type": None},
             # .docx
             "application/vnd.openxmlformats-officedocument.wordprocessingml.document": {
```
deeplow, line 142: Now that we have only a 1-1 mapping, maybe we should drop the nested dict.

container/dangerzone.py
deeplow: I'm OK with keeping it. But in that case can tweak a bit the last part of the commit? It says something like this:


> * We preemptively run LibreOffice in safe mode, to remove hardware
  acceleration and make sure **that macros are not invoked as well.**

deeplow: Maybe split out that part of the commit since it's not about supporting more doc formats?

apyrgio: Sure, I can split it and explain the bold part better.

container/dangerzone.py
apyrgio: Personally, I wanted to just have three sets: LibreOffice MIME types, `convert` MIME types, and no conversion MIME types.

However, I was thinking that changing the structure completely would prohibit including more info about a MIME type in the future, that's why I left it as is. If you hard-pressed on removing the nested dict though, I can do it.

container/dangerzone.py
apyrgio: Done in 58a8241844d6625667144b442ef95dc5b20e9e46.

container/dangerzone.py
deeplow: OK. Let's keep it as-is.

deeplow (approved):

PR #382, opened 2023-04-05
MacOS: Allow dettached signing of release and update developer ID

deeplow:
Redoes the build-app.py script to add an option to sign only an already-
produced app bundle and updates the apple developer ID.


deeplow:
Closing. Should be to the RC branch.

PR #384, opened 2023-04-05
MacOS: Allow dettached signing of release and update developer ID

deeplow:
Redoes the build-app.py script to add an option to sign only an already-produced app bundle and updates the apple developer ID.

### DIFF
--- a/install/macos/build-app.py
+++ b/install/macos/build-app.py
@@ -19,6 +19,28 @@ def run(cmd):
     subprocess.run(cmd, cwd=root, check=True)
 
 
+def build_app_bundle():
+    """
+    Builds the Dangerzone.app bundle and saves it in dist/Dangerzone.app
+    """
+    print("○ Deleting old build and dist")
+    if os.path.exists(build_path):
+        shutil.rmtree(build_path)
+    if os.path.exists(dist_path):
+        shutil.rmtree(dist_path)
+
+    print("○ Building app bundle")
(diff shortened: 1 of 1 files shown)


install/macos/build-app.py
```diff
@@ -37,96 +59,104 @@ def codesign(path, entitlements, identity):
...
+            "--app-drop-link",
+            "300",
+            "70",
+            dmg_path,
+            app_path,
+            "--identity",
+            identity_name_application,
+        ]
+    )
+    print(f"○ Finished building DMG: {dmg_path}")
```
apyrgio, line 125: Minor, but since this action is now separated from the building of the DMG, it would probably make sense to rename it to "Finished signing DMG".

apyrgio (approved):
Other than this minor comment, LGTM.

install/macos/build-app.py
deeplow: Actually, as far as I undestand, the DMG is only produced when the signing is enabled ([it was the case before](https://github.com/freedomofpress/dangerzone/pull/384/files#diff-9e3e8e5c0a05684d57b4ae2d4184a72ffc7ae648ee9627faa4a91a4024e728d7L74)). So I think we can keep it as is, but if you prefer I can also say that it's finished building and signing the DMG.

How would you prefer?

apyrgio: Oh, thanks for the explanation. Let's keep it then.

PR #385, opened 2023-04-06
 WIP: Add instructions for publishing/installing Fedora packages to/from packages.freedom.press

deeplow:
Adds install / deployment instructions for Fedora.

Install deployment suggests:
  - adding our key fingerprint to dangerzone.rocks and mastodon bio, potentially (for cross-checking), following [GrapheneOS's approach](https://grapheneos.org/install/cli#obtaining-factory-images) to fingerprint verification (as opposed to the rather complex PGP web of trust)
  - adding a repo definition to packages.freedom.press for easier installation

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,17 +53,52 @@ sudo apt install -y dangerzone
 
 ### Fedora
 
-Add our repository following [these instructions](https://packagecloud.io/firstlookmedia/code/install#manual-rpm), or by running this script:
+### 1. Add our repository
+
+Start by installing our repository. This allows you to
+
+Open a terminal window and type the following commands:
 
 ```
-curl -s https://packagecloud.io/install/repositories/firstlookmedia/code/script.rpm.sh | sudo bash
+gpg --keyserver hkps://keys.openpgp.org --recv-key "DE28 AB24 1FA4 8260 FAC9  B8BA A7C9 B385 2260 4281"
+gpg --export --armor A7C9B38522604281 | sudo tee /etc/pki/rpm-gpg/RPM-GPG-dangerzone.pub > /dev/null
+sudo yum-config-manager --add-repo=https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo
 ```
 
-Install Dangerzone:
+
+### 2. Install Dangerzone
+
 
 ```
-sudo dnf install -y dangerzone
+sudo dnf install dangerzone
 ```
+
+<details>
+<summary>Importing GPG key 0x22604281: ... Is this ok [y/N]:</summary>
+
+After some minutes of running the above command (depending on your internet speed) you'll be asked to confirm the fingerprint of our signing key. This is to make sure that in the case our servers are compromized your computer stays safe. It should look like this:
+
+```console
+--------------------------------------------------------------------------------
+Total                                           389 kB/s | 732 MB     32:07
+Dangerzone repository                           3.8 MB/s | 3.8 kB     00:00
+Importing GPG key 0x22604281:
+ Userid     : "Dangerzone Release Key <dangerzone-release-key@freedom.press>"
+ Fingerprint: DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281
+ From       : /etc/pki/rpm-gpg/RPM-GPG-dangerzone.pub
+Is this ok [y/N]:
+```
+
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -212,6 +212,42 @@ Rename `Dangerzone.msi` to `Dangerzone-$VERSION.msi`.
 
 Linux binaries are automatically built and deployed to repositories when a new tag is pushed.
 
+### Fedora
+
+Because the Fedora packages do not contain compiled Python code for a specific
+Python version, we can create a single Debian package and use it for all of our
+Debian-based distros.
```
apyrgio, line 219: This section needs to be reworded.

INSTALL.md
```diff
@@ -53,18 +53,53 @@ sudo apt install -y dangerzone
...
+
+Start by installing our repository. This allows you to
+
+Open a terminal window and type the following commands:
 
 ```
-curl -s https://packagecloud.io/install/repositories/firstlookmedia/code/script.rpm.sh | sudo bash
+gpg --keyserver hkps://keys.openpgp.org --recv-key "DE28 AB24 1FA4 8260 FAC9  B8BA A7C9 B385 2260 4281"
+gpg --export --armor A7C9B38522604281 | sudo tee /etc/pki/rpm-gpg/RPM-GPG-dangerzone.pub > /dev/null
+sudo yum-config-manager --add-repo=https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo
```
apyrgio, line 65: Do we also have to `yum-config-manager --enable` the repo as well. If so, we can do it only if the downloaded repo file matches a known checksum, so that we are sure we add the proper repo.

INSTALL.md
```diff
@@ -53,18 +53,53 @@ sudo apt install -y dangerzone
...
+Total                                           389 kB/s | 732 MB     32:07
+Dangerzone repository                           3.8 MB/s | 3.8 kB     00:00
+Importing GPG key 0x22604281:
+ Userid     : "Dangerzone Release Key <dangerzone-release-key@freedom.press>"
+ Fingerprint: DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281
+ From       : /etc/pki/rpm-gpg/RPM-GPG-dangerzone.pub
+Is this ok [y/N]:
 ```
 
+> **Note**: If it does not show this fingerprint confirmation or the fingerprint does not match, it is possible that our servers were compromized. Be distrustful and reach our to us.
```
rocodes, line 92: [nit, sorry] -- reach _out_, rather than 'our'

RELEASE.md
deeplow: Whoops. This was some copy-pasta mistake.

INSTALL.md
deeplow: Whoops. Thanks for taking a look and spotting this one!

RELEASE.md
deeplow: I guess the note that I added bellow was supposed to replace this, but I forgot to remove it in the end, I guess. Removing it now.

INSTALL.md
```diff
@@ -53,18 +53,53 @@ sudo apt install -y dangerzone
 
 ### Fedora
 
-Add our repository following [these instructions](https://packagecloud.io/firstlookmedia/code/install#manual-rpm), or by running this script:
+### 1. Add our repository
+
+Start by installing our repository. This allows you to
```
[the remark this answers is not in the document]
deeplow: Sentence removed.

INSTALL.md
deeplow: We may go with that approach. It's a long command, that I'd rather avoid, but it seems like the safest alternative (that doesn't need to user to do the fingerprint verification step -- but we're trusting github).

deeplow: I'll look at an alternative for importing directly to with `rpm --import` to import the fingerprint directly into Fedora's keyring.

deeplow: After discussion, we decided to go with the approach of [docker](https://docs.docker.com/engine/install/fedora/) and Tor and downloading a `.repo` file which references a remote key. And then we inform the user to verify the fingerprint matches the one on `INSTALL.md`.

apyrgio:
I added a small fixup commit for some minor typos. Other than that, feel free to merge.

apyrgio (approved):

PR #387, opened 2023-04-12
Release 0.4.1 mac final touches

deeplow:
Solves two issues with the macOS `build-app.py`.

### DIFF
--- a/install/macos/build-app.py
+++ b/install/macos/build-app.py
@@ -88,7 +88,7 @@ def sign_app_bundle(build_path, dist_path, app_path):
     print(f"○ Signed app bundle: {app_path}")
 
     # Detect if create-dmg is installed
-    if not os.path.exists("/usr/local/bin/create-dmg"):
+    if not shutil.which("create-dmg"):
         print("create-dmg is not installed, skipping creating a DMG")
         return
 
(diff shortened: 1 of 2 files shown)


install/macos/build-app.py
```diff
@@ -75,9 +75,7 @@ def sign_app_bundle(build_path, dist_path, app_path):
     icon_path = os.path.join(root, "install", "macos", "dangerzone.icns")
 
     print("○ Code signing app bundle")
-    identity_name_application = (
-        "Apple Distribution: Freedom of the Press Foundation (94ZZGGGJ3W)"
-    )
+    identity_name_application = "Apple Development: Trevor Timm (TW4JM8ZCU)"
```
apyrgio, line 78: Let's also change a similar line in `RELEASE.md`:

https://github.com/freedomofpress/dangerzone/blob/1c0dfb45f5dc6cc36fbfa2983ad2e1ff45aaef91/RELEASE.md?plain=1#L169

apyrgio: Speaking of this change, it may be worth remove this line before it:

https://github.com/freedomofpress/dangerzone/blob/1c0dfb45f5dc6cc36fbfa2983ad2e1ff45aaef91/RELEASE.md?plain=1#L168

This version is 3 major versions behind the current one. At this point, I don't think any developer will have a problem.

apyrgio: Actually, I did the above already, since I found some other small issues as well.

apyrgio:
I sent a FIXUP in this branch, since I found a few more issues.

apyrgio (approved):

PR #392, opened 2023-04-18
Minor changelog tweaks

eloquence:
Remove stray parentheses; expand "FPF"; use Packagecloud's own capitalization

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Dangerzone 0.4.1]
+## [Unreleased]
+
+## Dangerzone 0.4.1
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Dangerzone 0.4.1]
+## Dangerzone 0.4.1
```
apyrgio, line 8: Good catch about the stray brackets. What I _should_ have done instead is the following, which is in line with the ["Keep a Changelog"](https://keepachangelog.com/en/1.1.0/) format.

```
## [Unreleased]

## [Dangerzone 0.4.1]

[... rest of file ...]

[Unreleased]: https://github.com/freedomofpress/dangerzone/compare/v0.4.1...HEAD
[Dangerzone 0.4.1]: https://github.com/freedomofpress/dangerzone/compare/v0.4.0...v0.4.1
```

Are you ok with adding this to the commit instead?

apyrgio (commented):
Just one comment on the changelog links. Other than that, we can merge it.

CHANGELOG.md
eloquence: I think it'd be good to standardize on that format throughout in later follow-up -- I've for now amended it to just add and bracket the "Unreleased"

apyrgio: We discussed about and we want to make as minimal changes as possible. So we will just add the "Unreleased" block, without the extra links.

PR #401, opened 2023-04-28
Add note for MacOS 11+ users blocked by SIP

### LINKED ISSUE #371: consistently on convert, `invalid json returned from container`
On attempting to convert one or more files, error of `invalid json returned from container`. This happens consistently, and no files are converted successfully.

# versions

- Dangerzone: 0.4.0
- Docker: 4.17.0
- OS: MacOS 13.2.1, Intel

# reproduce

1. select file(s) of type .pdf, etc. I've tried a few.
2. tap Convert to Safe Document.

# other context

This issue is a change from long-time use, where conversion was working. This error began on 2023.02.06. The first indication of something being broken was Dangerzone failing to detect Docker.

I removed both, and reinstalled. This resolved Dangerzone not detecting Docker. Then `invalid json returned from container` began appearing.

Removing, and reinstalling has no further effect.

keywordnew:
A workaround for an issue related to SIP imposed on Docker has been identified in #371. This PR updates README.md to include friendly instructions for MacOS 11+ users blocked by this issue.

PR created at the request of @apyrgio in https://github.com/freedomofpress/dangerzone/issues/371#issuecomment-1523257761.

To include these instructions in the current README.md, I judged it most appropriate to nest it under a new FAQ section. Please feel free to refactor as needed.

Closes #371.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,17 @@ Set up a development environment by following [these instructions](/BUILD.md).
 
 ## FAQ
 
+### "I'm experiencing an issue while using Dangerzone."
+
+Dangerzone gets updates to improve its features _and_ to fix problems. So, updating may be the simplest path to resolving the issue. Here is how to update:
+
+1. Check which version of Dangerzone you are currently usi: run Dangerzone, then look for the series of numbers to the left of the logo. It will look similar to `0.4.1`
+2. Now find the latest available version of Dangerzone: go to the [download page](https://dangerzone.rocks/#downloads). Look for the number displayed using the same format as in Step 1.
+3. Is the version on the Dangerzone download page higher than the version of your installed app? Go ahead and update.
+
 ### "I get `invalid json returned from container` on MacOS Big Sur or newer (MacOS 11.x.x or higher)"
+
+Are you using the latest Dangerzone?
 
 You _may_ be attempting to convert a file in a directory to which Docker Desktop does not have access. Dangerzone for Mac requires Docker Desktop for conversion. Docker Desktop, in turn, requires permission from MacOS to access the directory in which your target file is located.
 
--- a/README.md
+++ b/README.md
@@ -48,15 +48,15 @@ Set up a development environment by following [these instructions](/BUILD.md).
 
 ### "I'm experiencing an issue while using Dangerzone."
 
-Dangerzone gets updates to improve its features _and_ to fix problems. So, updating may be the simplest path to resolving the issue. Here is how to update:
+Dangerzone gets updates to improve its features _and_ to fix problems. So, updating may be the simplest path to resolving the issue which brought you here. Here is how to update:
 
-1. Check which version of Dangerzone you are currently usi: run Dangerzone, then look for the series of numbers to the left of the logo. It will look similar to `0.4.1`
-2. Now find the latest available version of Dangerzone: go to the [download page](https://dangerzone.rocks/#downloads). Look for the number displayed using the same format as in Step 1.
+1. Check which version of Dangerzone you are currently using: run Dangerzone, then look for a series of numbers to the right of the logo within the app. The format of the numbers will look similar to `0.4.1`
+2. Now find the latest available version of Dangerzone: go to the [download page](https://dangerzone.rocks/#downloads). Look for the version number displayed. The number will be using the same format as in Step 1.
 3. Is the version on the Dangerzone download page higher than the version of your installed app? Go ahead and update.
 
 ### "I get `invalid json returned from container` on MacOS Big Sur or newer (MacOS 11.x.x or higher)"
 
-Are you using the latest Dangerzone?
+Are you using the latest version of Dangerzone? See the FAQ for: "I'm experiencing an issue while using Dangerzone."
 
 You _may_ be attempting to convert a file in a directory to which Docker Desktop does not have access. Dangerzone for Mac requires Docker Desktop for conversion. Docker Desktop, in turn, requires permission from MacOS to access the directory in which your target file is located.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,3 +43,19 @@ Dangerzone can convert these types of document into safe PDFs:
...
+
+### "I get `invalid json returned from container` on MacOS Big Sur or newer (MacOS 11.x.x or higher)"
+
+You _may_ be attempting to convert a file in a directory to which Docker Desktop does not have access. Dangerzone for Mac requires Docker Desktop for conversion. Docker Desktop, in turn, requires permission from MacOS to access the directory in which your target file is located.
+
+To grant this permission:
+
+1. On MacOS 13, choose Apple menu > System Settings. On lower versions, choose System Preferences.
+2. Tap into Privacy & Security in the sidebar. (You may need to scroll down.)
+3. In the Privacy section, tap into Files & Folders. (Again, you may need to scroll down.)
```
apyrgio, line 57: In a Mac system with enabled SIP and OS version 13.2.1, I can't find "Docker" under "Privacy & Security" -> "Files and Folders". Granted, it may be because Docker is not somehow triggering SIP in that system. For what is worth, Docker is installed from the official .dmg.

apyrgio (commented):
Hi @keywordnew. Thanks a lot for the PR, and sorry for the review delay. I needed to sort some things out post 0.4.1 release.

While looking at our instructions, I realized something:

* In Dangerzone 0.4.1, before Docker becomes involved in the conversion process, Dangerzone copies the untrusted file (e.g., from Downloads) to a temporary directory. Then, Docker bind mounts it from the temporary directory into the container.
* In Dangerzone 0.4.0, Docker mounts it directly from the original location (e.g., from Downloads).

In my system, this chain of events does not trigger SIP or produce an error (I fail to see why), but things may be different in your system. Also, this change on the location of the document affects the instructions of this PR. I would assume that Dangerzone, not Docker, would now be the program that would need to access the file under Downloads.

So, would you mind checking once more these instructions for Dangerzone 0.4.1, to make sure they are up-to-date?

apyrgio:
Hi @keywordnew, gentle ping for the review comments above. Unfortunately I can't reproduce these instructions (and the original issue) in my MacOS device, that's why I'd like your feedback on this.

keywordnew:
Thanks for letting me know to reprioritize this, @apyrgio.

"checking once more these instructions for Dangerzone 0.4.1, to make sure they are up-to-date"

Below are findings from two conversion flows: CLI and GUI for Dangerzone 0.4.1

<details>

<summary>CLI</summary>

pre: disable "Privacy & Security" -> "Files and Folders" -> parent directory of test.pdf

1. initiate conversion of test.pdf
2. conversion successful

</details>

<details>

<summary>GUI</summary>

pre: disable "Privacy & Security" -> "Files and Folders" -> parent directory of test.pdf

1. initiate conversion of test.pdf
2. modal triggered: "“Dangerzone.app” would like to access files in your Downloads folder."
4. deny request
5. file picker appears, test.pdf can be selected
6. conversion apparently successful

</details>

keywordnew:
In view of https://github.com/freedomofpress/dangerzone/pull/401#pullrequestreview-1420532783 and https://github.com/freedomofpress/dangerzone/pull/401#issuecomment-1552919304, this PR with instructions to handle permissions-related UX issues needs amending or withdrawal.

1. The instructions in the PR would now be useful **only to users using Dangerzone 0.4.0 and lower**.
  - Does the project have metrics on the percentage of users on older versions of Dangerzone? Keeping the instructions would be a way to avoid difficult user experiences, if a non-trivial cohort of users do not regularly update their installed version. Otherwise, I can withdraw this PR.
2. As **permissions do not actually seem necessary** when using Dangerzone 0.4.1, instructions to deny the permissions modal triggered by the GUI-based conversion of Dangerzone 0.4.1 could be added to the PR.
  - Or made to replace the existing content of the PR, depending on the decision on item 1 above.

apyrgio:
I see, so you're saying that the modal for SIP does not actually prevent Dangerzone from accessing the file. It's like the user's choice is inconsequential. That's... weird.

In any case, this is something that we can monitor in this issue tracker, and gather user feedback.

>  The instructions in the PR would now be useful only to users using Dangerzone 0.4.0 and lower.

This seems to be the case. Still, I would prefer if affected users (ideally, every user) upgraded their Dangerzone installation to the latest one. It contains security fixes and improvements that are important.

If Dangerzone 0.4.0 users reach our GitHub page, I believe the most important thing is to see that there's a new release, rather than circumvent an issue locally.

> Does the project have metrics on the percentage of users on older versions of Dangerzone? 

Nope, we don't have any way to track installed versions, nor a way to inform users about updates, aside from the GitHub releases and our Mastodon channel. This is something that we should solve at some point in a privacy-respecting way.

> As permissions do not actually seem necessary when using Dangerzone 0.4.1, instructions to deny the permissions modal triggered by the GUI-based conversion of Dangerzone 0.4.1 could be added to the PR.

The thing is, I can't wrap my head around this choice, and I think that users will be confused as well. We don't have a good explanation of what's the consequence of their action, in order to help them decide. If we did, I'd be more than fine with adding such an instruction :-/.

keywordnew:
> If Dangerzone 0.4.0 users reach our GitHub page, I believe the most important thing is to see that there's a new release, rather than circumvent an issue locally.

May or may not be the best path, but seemed reasonable to add directions to update the app as the first entry in the FAQ. 

Changed PR to draft since a course of action is still unclear.

apyrgio:
Hey @keywordnew. I like the fact that you put first and foremost that users should upgrade to the latest version, and then offer a potential solution. I think this is a good way forward.

I'll be mostly out this week, so I'll manage to see your PR from next week. But I believe we have something that we can merge :-).

keywordnew:
Glad the PR can be made useful, @apyrgio. Just to note, the commits on the PR branch are atomic only for my benefit. So, if a decision is made to merge, a squash would clean up the history :-)

apyrgio (approved):

apyrgio:
Thanks a lot @keywordnew for improving the PR and offering these MacOS instructions. Even if the issue you encountered in 0.4.0 does no longer bite our users in 0.4.1+, it's nice to have SIP instructions somewhere in our project. Also, the fact that we stress that users must always upgrade when they encounter an error is a very welcome addition.

> Glad the PR can be made useful, @apyrgio. Just to note, the commits on the PR branch are atomic only for my benefit. So, if a decision is made to merge, a squash would clean up the history :-)

Sigh, I was thinking of squashing the commits, but I didn't want to be too intrusive. Unfortunately I also forgot about this message. Anw, that's ok, the extra commit is not that bad.

Issue #403, opened 2023-04-28
Support for multiple OCR languages for parallel conversion of documents in different languages


sudwhiwdh:
![03](https://user-images.githubusercontent.com/58292468/235225505-48700a28-25fc-4bdf-9665-3af1394ed8d8.png)

How can I select the appropriate OCR languages in the application interface if I have documents in different languages at the same time?

And related to the screenshot: Why is there a comma between document and language?

apyrgio:
I'm afraid that neither the GUI nor the CLI offer such an option. The workaround for now would be to batch the documents for each language into separate CLI invocations (i.e., different `--ocr-lang <lang>` parameters).

As for the typo you found, that's interesting. It was there since the first release of Dangerzone. We'll fix it though, thanks.

sudwhiwdh:
> As for the typo you found, that's interesting. It was there since the first release of Dangerzone. We'll fix it though, thanks.

https://github.com/freedomofpress/dangerzone/pull/404

deeplow:
The whole OCR thing will need some re-thinking. One thing we realized as well is that users may not know in advance which language the document is in before opening it.

deeplow:
And not everybody is familiar with the term OCR, but they much more familiar with the idea of having a document searchable. So maybe the text should instead be: make document searchable.

As for the multi-lingual situation, there's a technical limitation. The way we do OCR, the tool (tesseract) needs to know the language. Not sure it can autodetect.

deeplow:
> As for the multi-lingual situation, there's a technical limitation. The way we do OCR, the tool (tesseract) needs to know the language. Not sure it can autodetect.

After a bit of research, while it looks like Tesseract itself isn't equipped to detect document languages, there are some models (like [langid](https://github.com/saffsd/langid.py) -- which seems no longer supported) that do it. We could theoretically run this auto-detection on every page.

But I wonder if having mixed languages will be technically complicated because even in the same page there may be multiple languages. So I'm wondering if we'll need [document layout analysis](https://huggingface.co/blog/document-ai#2-what-is-doc-layout) and then apply OCR in that section. Food for thought. If that's the case, then OCR in documents with multiple languages may be more a long term goal.

deeplow:
Today I learned that tesseract OCR has multilingual support

![multiling](https://github.com/freedomofpress/dangerzone/assets/47065258/d5866dd7-8967-44d8-8bfa-8782ad40c03a)

[...]

PR #404, opened 2023-05-02
fix gui typo

sudwhiwdh:
https://github.com/freedomofpress/dangerzone/issues/403#issuecomment-1531821246


deeplow (approved):
Thanks for finding and fixing this. I had actually never realized this comma was there. I guess this is just a consequence of looking at the tool for too long and a fresh pair of eyes can see something from a new perspective.

sudwhiwdh:
I am happy to be able to contribute along the way!

PR #405, opened 2023-05-03
ci: Add security scanning

### LINKED ISSUE #222: Defense in Depth - Security Scanning
Parent issue: #221

Attackers that want to escape the Dangerzone container, first need to find a vulnerability in the conversion software. Ensuring that no such vulnerabilities exist is the first layer of defense against attackers.

In practice, we can't guarantee that the installed software does not have any security vulnerabilities. However, what we can do is:

1. Continuously scan the `dangerzone.rocks/dangerzone` container image.
    * Use an open source scanner like [Grype](https://github.com/anchore/grype).
    * Use CircleCI to scan the image daily and report any findings.
2. Update the affected software if a Critical/High vulnerability is found.

apyrgio:
Add two GitHub Actions workflows, that perform the following checks:

* Security scan the Python dependencies of the Dangerzone application
  (`poetry.lock`), for the current/main branch.
* Build and security scan the Dangerzone container image for the
  current/main branch.
* Security scan the Python dependencies of the Dangerzone application
  (`poetry.lock`), for the latest release of Dangerzone (currently
  v0.4.1).
* Download and security scan the Dangerzone container image for the
  latest release of Dangerzone (currently v0.4.1).

The first two checks will run on branch pushes, PRs, and nightly. The
last two checks will run only nightly, since the code in the current
branch cannot affect already released artifacts.

Note that in order to be able to run our checks on the released container image, 
we need to make it part of our release artifacts.

Also, besides the security scans, these workflows will also update the
Security alerts in the GitHub page for the Dangerzone project, and print
the SARIF report to the stdout, for debugging purposes.

Finally, ignore two CVEs from our security scans, which were triggered when
scanning the Dangerzone container image for v0.4.1. These CVEs do not
affect out users, and we offer an explanation why.

Closes #222

### DIFF
--- a/.github/workflows/scan_released.yml
+++ b/.github/workflows/scan_released.yml
@@ -1,0 +1,77 @@
+name: Scan released app and container
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run every day at 00:00 UTC.
+
+jobs:
+  security-scan-container:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Download container image for the latest release
+        run: |
+          VERSION=$(curl https://api.github.com/repos/freedomofpress/dangerzone/releases/latest | jq -r '.tag_name')
+          wget https://github.com/freedomofpress/dangerzone/releases/download/${VERSION}/container.tar.gz
+      - name: Load container image
+        run: docker load -i container.tar.gz
+      # NOTE: Scan first without failing, else we won't be able to read the scan
+      # report.
+      - name: Scan container image (no fail)
+        uses: anchore/scan-action@v3
+        id: scan_container
+        with:
+          image: "dangerzone.rocks/dangerzone:latest"
+          fail-build: false
+          only-fixed: true
+          severity-cutoff: critical
+      - name: Upload container scan report
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: ${{ steps.scan_container.outputs.sarif }}
+          category: container
+      - name: Inspect container scan report
+        run: cat ${{ steps.scan_container.outputs.sarif }}
+      - name: Scan container image
+        uses: anchore/scan-action@v3
+        with:
+          image: "dangerzone.rocks/dangerzone:latest"
+          fail-build: true
+          only-fixed: true
+          severity-cutoff: critical
+
+  security-scan-app:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Checkout the latest released tag
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/scan_released.yml
```diff
@@ -0,0 +1,75 @@
+name: Scan released app and container
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run every day at 00:00 UTC.
+env:
+  RELEASED_VERSION: 0.4.1
```
deeplow, line 6: I would suggest that we take this directly from github's API so we don't forget to change. The following code should do the trick:

```
curl https://api.github.com/repos/freedomofpress/dangerzone/releases | jq '.[] | select(.draft==false and .prerelease==false) | .tag_name' | head -n 1
"v0.4.1"
```

Maybe also using [env variables like](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) `GITHUB_REPOSITORY_OWNER` and `GITHUB_REPOSITORY` in case someone wants to fork the project and run this automation there as well.

deeplow (changes requested):

.github/workflows/scan_released.yml
deeplow: > > I would suggest that we take this directly from github's API so we don't forget to change.
> 
> Nice! I wanted to leave this as a todo item, but you found an elegant solution. I can incorporate it, thanks.

There is an even more elegant one, apparently:


```bash
curl https://api.github.com/repos/freedomofpress/dangerzone/releases/latest | jq '.tag_name'
```

>>In case someone wants to fork the project and run and see this automation output we could use generic [env variables like](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) GITHUB_REPOSITORY_OWNER and GITHUB_REPOSITORY instead of hardcoding ours?
>
>I think this may end up breaking this action for other users, since by forking our repo, they don't also copy our releases. So, they won't be able to get the latest releast tag, nor the tarball. If they want to somehow mirror those in their fork, I guess they would change this file as well. On the other hand, if they simply want to run this action, they don't need to touch anything. Does this make sense?

You're right! Then let's leave that hardcoded then.

.github/workflows/scan_released.yml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Checkout the latest released tag
+        run: |
+          VERSION=$(curl https://api.github.com/repos/freedomofpress/dangerzone/releases/latest | jq '.tag_name')
+          git checkout $VERSION
```
deeplow, line 54: > error: pathspec '"v0.4.1"' did not match any file(s) known to git

Looks like we need to trim the `"` inner quotes

deeplow (approved):

Issue #358, opened 2023-02-28
Debian packages produced on Ubuntu Focal do not work on Ubuntu Jammy/Kinetic


apyrgio:
The entry point scripts for Dangerzone don't work on Ubuntu Jammy/Kinetic, and instead throw the following error:

```python
Traceback (most recent call last):
  File "/usr/bin/dangerzone", line 6, in <module>
    from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'
```

The issue is that `python-setuptools` can be missing from these distros.

apyrgio:
We have updated our release instructions to build the .deb package on Debian Bookworm, and we've successfully tested installing this package on the rest of our Debian-based flavors. See also this [successful CI run](https://github.com/freedomofpress/dangerzone/actions/runs/4658789885) on a WIP branch, which we have to tidy up and merge at some point.

PR #406, opened 2023-05-03
ci: Test building a .deb and installing it

apyrgio:
Update our GitHub Actions workflow with the following tests:

1. Build a .deb for Dangerzone on Debian Bookworm.
2. Install this .deb on every Debian-based platform that we support.
3. Test that the installed version runs successfully.

This way, we can be sure that .deb that we create on a single Debian
version (here we choose Debian Bookworm) works on all platforms.

Refs #358

### DIFF
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -71,15 +71,19 @@ RUN . /etc/os-release \
 """
 
 # FIXME: Do we really need the python3-venv packages?
-# XXX: We install uidmap separately, because it is not a hard dependency for Podman, and
-# we use --no-install-recommends.
 DOCKERFILE_BUILD_DEV_DEBIAN_DEPS = r"""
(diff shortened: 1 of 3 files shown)


dev_scripts/env.py
```diff
@@ -307,7 +307,23 @@ def run(
         # We need to retain our UID, because we are mounting the Dangerzone source to
         # the container.
```
deeplow, line 308: Does this comment still apply

apyrgio: It does, it's just that we now offer a more robust way of retaining it.

deeplow (approved):

deeplow (approved):
All good. I just hope that GitHub never charges for its CI, otherwise, we'll have a surprise bill :money_mouth_face:.

PR #423, opened 2023-05-22
ci: Fix transient errors in Debian Bullseye

### LINKED ISSUE #388: Transient CI error on Debian Bullseye
Our CI tests on Debian Bullseye sometimes fail with:

```
Failed to detect the owner for the current cgroup: stat /sys/fs/cgroup/systemd/system.slice/cloud-final.service: no such file or director
```

Re-running the CI tests sometimes works, but we need to figure out why this happens.

apyrgio:
Fix transient errors in Debian Bullseye CI tests by using a different machine image (Ubuntu 22.04 vs Ubuntu 20.04), and solving some Podman config issues along the way.

Fixes #388

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,12 +299,29 @@ jobs:
             ./dev_scripts/env.py --distro debian --version bookworm run --dev \
                 bash -c 'cd dangerzone; poetry run make test'
 
+  # NOTE: Making CI tests work in Debian Bullseye requires some tip-toeing
+  # around certain Podman issues, as you'll see below. Read the following for
+  # more details:
+  #
+  # https://github.com/freedomofpress/dangerzone/issues/388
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -345,6 +357,22 @@ jobs:
           command: |
             ./dev_scripts/env.py --distro debian --version bullseye build-dev
 
+      - run:
+          name: Configure Podman for Debian Bullseye
+          command: |
+            # Copy the Podman config into the container image we created for the
```
deeplow, line 363: Can we add here a reference to the issue that this extra code is trying to solve?

apyrgio: I've updated the above block with a link to #388.

deeplow: can you point me to the line / commit? I can't find it

Issue #436, opened 2023-06-07
Qubes: allow parallel conversions


deeplow:
Dangerzone-Qubes hardcodes the output directory for the last part of the conversion to `/tmp/dangerzone`. We need to use a random temporary directory instead, and clean it after the conversion finishes.

apyrgio:
This is good to know. I'll move this issue to the 0.5.1milestone, since it's not a limitation as long as the queue approach remains (which will be the case for 0.5.0).

Issue #430, opened 2023-05-29
Qubes: Error handling


deeplow:
We need to accommodate exceptions for all edge-cases in the client binary-protocol-parsing code. This was out of scope of the alpha stage (https://github.com/freedomofpress/dangerzone/issues/411).

Errors to check on server:
  - [x] **document wasn't immediately received**:
     probably the user exited dangerzone mid-conversion. Safely exit from dangerzone so that there isn't a dangling disposable qube
- [x] The number of received pages was (> MAX_PAGES)
    - **note:** this one is probably going to have to be done server-side. The (Qubes) client retrieves 2 bytes for pages. "Overflows" won't be detected.

Errors to check on the client:
- [x] The disposable qube could not start due to memory issues.
- [x] The disposable qube could not start because `dz-dvm` does not exist. (**update: done in https://github.com/freedomofpress/dangerzone/pull/564)
- [x] The width and the height of the pages are too large.
- [x] Page conversion timeout reached
- [x] Timed out receiving a page
- [x] Consume stderr ([source](https://github.com/freedomofpress/dangerzone/pull/437#discussion_r1236644777))
- [ ] Show helpful message if OCR step fails (e.g., unsupported language)
- [ ] <strike>  Stop disposable VMs when aborting conversions </strike> **update:** will be tackled separately since it also affects containers. Moved to https://github.com/freedomofpress/dangerzone/issues/563

Errors to check on the client:     

<strike>Also, we need to sanitize tracebacks and errors from the disposable qube, in a way that does not affect the user's terminal (e.g., remove control characters).</strike> **update**: this will be done in https://github.com/freedomofpress/dangerzone/pull/386

deeplow:
I moved the "number of pages received" to be a server-side check. The client won't have a way of knowing this.

apyrgio:
We were a bit overeager to close this issue, as there are still some error cases that are missing:

- [ ]  The disposable qube could not start because `dz-dvm` does not exist.
  * Currently, we return a wrong error message here, that the system was out of RAM, but this is not the case.
  * It's probably best to return a generic message here "Could not start disp qube `dz-dvm`. Consult your system logs". Qubes UI shows a notification with more context, after all.
- [ ]  Show helpful message if OCR step fails (e.g., unsupported language)
  - Admittedly, this error is unlikely to happen, because we now install every OCR language available.
  - I propose moving this to the stabilization effort.
- [ ] Stop disposable VMs when aborting conversions 
[...]

Issue #429, opened 2023-05-29
Qubes: Progress reports


deeplow:
This requires getting / detecting progress information from the two stages of the conversion:
  - **Stage 1**  (Document to Pixels) - we need to detect the progress as pages arrive from the disposable qube
  - **Stage 2** (Pixels to PDF + OCR + compression) - Somehow harmonize the fact that in non-Qubes systems, information is obtained from the docker container and in a Qubes system, this runs natively.

Issue #412, opened 2023-05-15
Qubes: Beta integration


apyrgio:
For Qubes OS Beta integration, users should be able to enjoy most of the important features that Dangerzone provides in the other platforms:

- [x] https://github.com/freedomofpress/dangerzone/issues/429
- [x] https://github.com/freedomofpress/dangerzone/issues/430
- [x] https://github.com/freedomofpress/dangerzone/issues/431
- [x] #446
- [x] #438 
- [x] https://github.com/freedomofpress/dangerzone/issues/455

Optional:
- [ ]  https://github.com/freedomofpress/dangerzone/issues/443

deeplow:
> * :heavy_check_mark: OCR

This will present some challenges since alpine likely ships a different set of tesseract-ocr languages than fedora (on which our Qubes implementation be probably be based).

deeplow:
> > * heavy_check_mark OCR
> 
> This will present some challenges since alpine likely ships a different set of tesseract-ocr languages than fedora (on which our Qubes implementation be probably be based).

No longer an issue thanks to https://github.com/freedomofpress/dangerzone/pull/422

apyrgio:
Now that streaming support for containers has been postponed for another release, and the rest of the tasks have completed, we can officially close this issue :tada:

PR #437, opened 2023-06-07
Qubes Integration Proof of Concept

### LINKED ISSUE #411: Qubes: Alpha integration
For Qubes OS Alpha integration, users should be able to run Dangerzone, and spawn a disposable VM for the conversion. Limited functionality is expected, and more specifically should be:

* :x: https://github.com/freedomofpress/dangerzone/issues/429
* :x: https://github.com/freedomofpress/dangerzone/issues/430
* :x: https://github.com/freedomofpress/dangerzone/issues/431
* :x: #446

What we **expect to work** is:

* :heavy_check_mark: Disposable VMs (https://github.com/freedomofpress/dangerzone/issues/414)
* :heavy_check_mark: Container isolation provider is not affected
* :heavy_check_mark: Manual installation instructions
* :heavy_check_mark: GUI
* :heavy_check_mark: OCR + Compression

### LINKED ISSUE #414: Add Qubes OS isolation provider
Add an isolation provider for Qubes OS. Currently, we have two isolation providers:

* Container: This is our main isolation provider, which is used in every platform. Depending on the platform type, it uses Docker or Podman.
* Dummy: This isolation provider is used in some of our tests, e.g., in Windows/MacOS platforms.

The Qubes OS isolation provider should handle the first and second stages of the conversion:

* Specify a Qubes RPC policy that will allow running a Qubes RPC service on a disposable VM.
* Spawn a disposable VM for the first stage of the conversion, and specify the Qubes service that will run.
* Send the file to the disposable VM in the same way that the TrustedPDF sends the file.
* Receive the pixels from the disposable VM in a secure fashion.
* Convert the pixels to PDF, without OCR or compression.

deeplow:
_Initial work to port Dangerzone to Qubes OS, making use of Qubes VMs as an isolation provider._ See `BUILD.md` for instructions on how to setup and run. Fixes #411 and fixes #414.

## (some) implementation details:
  - one disposable qube is started per document
  - first stage of the conversion (`document_to_pixels`) takes place in disposable qube
  - unlike with containers the second conversion stage (OCR+compression) takes place in the host, since the sanitization already happened.

More details can be found [on the wiki page](https://github.com/freedomofpress/dangerzone/wiki/Qubes-OS-Integration)

## Primary code changes
  - **moves container code to `dangerzone/conversion/` and spiting `dangerzone.py` into the two files**
    This was needed given that the second stage of the conversion original container code is ran on the primary qube (the one which runs the dangerzone GUI)
  - **Qubes conversion is a wrapper** (`dangerzone/conversion/doc_to_pixels_qubes_wrapper.py`)
    At the moment, the Qubes conversion is a simple wrapper around the `doc_to_pixels` code. This was because in Qubes pages are streamed one-by-one instead of all files sent at once after the conversion is done. The container-code however is still not ready to stream pages. So our wrapper converts the pages like before and then streams the results back to the client.

## Limitations

- It does not support simultaneous file conversion as the file being converted [is hard-coded](https://github.com/freedomofpress/dangerzone/compare/main...qubes-integration-poc#diff-4f7d2fafddfde6fdcef12c68e39696a5a11cb56294146a26e94bc207c3252c76R110) as `/tmp/input-file`.

- conversion progress was out of scope, but part of it is already working (for the first container stage). However, due to the "wrapper approach" it will jump from 0-50% in fractions of a second since when it starts sending pages, all the conversion has already happened.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -102,11 +102,193 @@ Create a .rpm:
 ./install/linux/build-rpm.py
 ```
 
-## QubesOS
+## Qubes OS
 
-Create a Debian- or Fedora-based standalone VM with at least 8GB of private storage space, and follow the relevant instructions above.
+<details>
+  <summary><i>:memo: Expand this section if you want to use containers instead of disposable qubes.</i></summary>
+  </br>
 
-Over time, you may need to increase disk space or prune outdated Docker images if you run into build issues on this VM.
+  Create a Debian or Fedora-based development standalone qube with at least
+  8GB of private storage space, and follow the relevant instructions above for
+  the respective template.
+
+  Remember to set the environment variable `DZ_USE_CONTAINERS=1`, before executing
+  Dangerzone.
+
+  Over time, you may need to increase disk space or prune outdated container
+  images if you run into build issues on this VM.
+</details>
+
+> :warning: Native Qubes support is in alpha stage, so the instructions below
+> require switching between qubes, and are subject to change.
+
+### Initial Setup
+
+The following steps must be completed once. Make sure you run them in the
+specified qubes.
+
+#### In `dom0`
+
+1. Create a new Fedora **template** (`fedora-37-dz`) for Dangerzone development:
+
+   ```
+   qvm-clone fedora-37 fedora-37-dz
+   ```
+
+   > :bulb: Alternatively, you can use your base Fedora 37 template in the
+   > following instructions. In that case, replace `fedora-37-dz` with
+   > `fedora-37` in the steps below.
+
+2. Create a **disposable**, offline app qube (`dz-dvm`), based on the
+   `fedora-37-dz` template. This will be the qube where the documents will be
+   sanitized:
+
+   ```
+   qvm-create --class AppVM --label red --template fedora-37-dz \
+       --prop netvm="" --prop template_for_dispvms=True \
+       dz-dvm
+   ```
+
+3. Create an **app** qube (`dz`) that will be used for Dangerzone development
+   and initiating the sanitization process:
+
+   ```
+   qvm-create --class AppVM --label red --template fedora-37-dz dz
+   ```
+
+   > :bulb: Alternatively, you can use a different app qube for Dangerzone
+   > development. In that case, replace `dz` with the qube of your choice in the
+   > steps below.
+
+4. Add an RPC policy (`/etc/qubes/policy.d/50-dangerzone.policy`) that will
+   allow launching a disposable qube (`dz-dvm`) when Dangerzone converts a
+   document, with the following contents:
+
+   ```
+   dz.Convert         *       @anyvm       @dispvm:dz-dvm  allow
+   dz.ConvertDev      *       @anyvm       @dispvm:dz-dvm  allow
+   ```
+
+#### In the `fedora-37-dz` template
+
+1. Install dependencies:
+
+   ```
+   sudo dnf install -y rpm-build pipx qt5-qtbase-gui libreoffice python3-magic \
+       tesseract*
+   ```
+
+2. Shutdown the `fedora-37-dz` template:
+
+   ```
+   shutdown -h now
+   ```
+
+#### In the `dz` app qube
+
+1. Clone the Dangerzone project:
+
+   ```
+   git clone https://github.com/freedomofpress/dangerzone
+   ```
+
+2. Install Poetry using `pipx`:
+
+   ```sh
+   pipx install poetry
+   ```
+
+3. Change to the `dangerzone` folder, and install the poetry dependencies:
+
+   ```
+   poetry install
+   ```
+
+   > **Note**: due to an issue with
+   > [poetry](https://github.com/python-poetry/poetry/issues/1917), if it
+   > prompts for your keyring, disable the keyring with `keyring --disable` and
+   > run the command again.
+
+4. Change to the `dangerzone` folder and copy the Qubes RPC calls into the
+   template for the **disposable** qube that will be used for document
+   sanitization (`dz-dvm`):
+
+   ```
+   qvm-copy-to-vm dz-dvm qubes/
+   ```
+
+#### In the `dz-dvm` template
+
+1. Create the directory that will contain the Dangerzone RPC calls, if it does
+   not exist already:
+
+   ```
+   sudo mkdir -p /rw/usrlocal/etc/qubes-rpc/
+   ```
+
+2. Move the files we copied in the previous step to their proper place:
+
+   ```
+   sudo cp ~/QubesIncoming/dz/qubes/* /rw/usrlocal/etc/qubes-rpc/
+   ```
+
+3. Shutdown the `dz-dvm` template:
+
+   ```
+   shutdown -h now
+   ```
+
+### Developing Dangerzone
+
+From here on, developing Dangerzone is similar as in other Linux platforms. You
+can run the following commands in the `dz` app qube:
+
+```sh
+# start a shell in the virtual environment
+poetry shell
+
+# run the CLI
+./dev_scripts/dangerzone-cli --help
+
+# run the GUI
+./dev_scripts/dangerzone
+```
+
+Create a .rpm:
+
+```sh
+./install/linux/build-rpm.py --qubes
+```
+
+For changes in the server side components, you can simply edit them locally,
+and they will be mirrored to the disposable qube through the `dz.ConvertDev`
+RPC call.
+
+The only reason to update the `fedora-37-dz` template from there on is if:
+1. The project requires new server-side components.
+2. The code for `dz.ConvertDev` needs to be updated. Copy the updated file
+   as we've shown in the steps above.
+
+### Installing Dangerzone system-wide
+
+If you want to test the .rpm you just created, you can do the following:
+
+On the `dz` app cube, copy the built `dangerzone.rpm` to `fedora-37-dz`
+template:
+
+```
+qvm-copy-to-vm fedora-37-dz dist/dangerzone*.noarch.rpm
+```
+
+On the `fedora-37-dz` template, install the copied .rpm:
+
+```
+sudo dnf install -y ~/QubesIncoming/dz/dangerzone-*.rpm
+```
+
+Shutdown the `fedora-37-dz` template and the `dz` app qube, and then you can
+refresh the applications on the `dz` qube, find Dangerzone in the list, and use
+it to convert a document.
 
 ## macOS
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,8 +33,11 @@ RUN mkdir tessdata && cd tessdata \
     && find . -name '*.traineddata' -maxdepth 2 -exec cp {} /usr/share/tessdata \; \
     && cd .. && rm -r tessdata
 
-COPY dangerzone.py /usr/local/bin/
-RUN chmod +x /usr/local/bin/dangerzone.py
+ENV PYTHONPATH=/opt/dangerzone
+
+RUN mkdir -p /opt/dangerzone/dangerzone
+RUN touch /opt/dangerzone/dangerzone/__init__.py
+COPY conversion /opt/dangerzone/dangerzone/conversion
 
 # Add the unprivileged user
 RUN adduser -s /bin/sh -D dangerzone
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,6 +8,7 @@ Dangerzone is available for:
 - Fedora 38
 - Fedora 37
 - Fedora 36
+- Qubes OS (alpha support)
 
 ### Ubuntu, Debian
 
@@ -117,8 +118,24 @@ After confirming that it matches, type `y` (for yes) and the installation should
 
 </details>
 
+### Qubes OS
 
+<details>
+  <summary><i>:memo: Expand this section if you want to use containers instead of disposable qubes.</i></summary>
+  </br>
+
+  Create a Debian or Fedora-based development standalone qube with at least
+  8GB of private storage space, and follow the relevant instructions above for
+  the respective template.
+
+  Remember to set the environment variable `DZ_USE_CONTAINERS=1`, before
+  executing Dangerzone.
+</details>
+
+> :warning: Native Qubes support is in alpha stage, so we don't have official
+> installation instructions yet. If you want to try out Dangerzone with native
+> Qubes support, check out our [build instructions](BUILD.md#qubes-os) instead.
 
 ## Build from source
 
-If you'd like to build from source, follow the [build instructions](https://github.com/freedomofpress/dangerzone/blob/master/BUILD.md).
+If you'd like to build from source, follow the [build instructions](BUILD.md).
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -6,9 +6,11 @@ import click
 from colorama import Back, Fore, Style
 
 from . import args, errors
+from .conversion.common import running_on_qubes
 from .document import ARCHIVE_SUBDIR, SAFE_EXTENSION
 from .isolation_provider.container import Container
 from .isolation_provider.dummy import Dummy
+from .isolation_provider.qubes import Qubes
 from .logic import DangerzoneCore
 from .util import get_version
 
@@ -63,6 +65,8 @@ def cli_main(
 
     if getattr(sys, "dangerzone_dev", False) and dummy_conversion:
         dangerzone = DangerzoneCore(Dummy())
+    elif running_on_qubes():
+        dangerzone = DangerzoneCore(Qubes())
     else:
         dangerzone = DangerzoneCore(Container(enable_timeouts=enable_timeouts))
 
--- a/dangerzone/conversion/common.py
+++ b/dangerzone/conversion/common.py
@@ -17,6 +17,14 @@ TIMEOUT_PER_MB: float = 30  # (seconds)
 TIMEOUT_MIN: float = 60  # (seconds)
 
 
+def running_on_qubes() -> bool:
+    # https://www.qubes-os.org/faq/#what-is-the-canonical-way-to-detect-qubes-vm
+    if os.environ.get("DZ_USE_CONTAINERS", "0") == "0":
+        return os.path.exists("/usr/share/qubes/marker-vm")
+    else:
+        return False
+
+
 async def read_stream(
     sr: asyncio.StreamReader, callback: Optional[Callable] = None
 ) -> bytes:
@@ -126,9 +134,10 @@ class DangerzoneConverter:
         pass
 
     def update_progress(self, text: str, *, error: bool = False) -> None:
-        print(
-            json.dumps(
-                {"error": error, "text": text, "percentage": int(self.percentage)}
+        if not running_on_qubes():
+            print(
+                json.dumps(
+                    {"error": error, "text": text, "percentage": int(self.percentage)}
+                )
             )
-        )
-        sys.stdout.flush()
+            sys.stdout.flush()
--- a/dangerzone/conversion/common.py
+++ b/dangerzone/conversion/common.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python3
+
+import asyncio
+import glob
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+import time
+from abc import abstractmethod
+from typing import Callable, Dict, List, Optional, Tuple, Union
+
+TIMEOUT_PER_PAGE: float = 30  # (seconds)
+TIMEOUT_PER_MB: float = 30  # (seconds)
+TIMEOUT_MIN: float = 60  # (seconds)
+
+
+async def read_stream(
+    sr: asyncio.StreamReader, callback: Optional[Callable] = None
+) -> bytes:
+    """Consume a byte stream line-by-line.
+
+    Read all lines in a stream until EOF. If a user has passed a callback, call it for
+    each line.
+
+    Note that the lines are in bytes, since we can't assume that all command output will
+    be UTF-8 encoded. Higher level commands are advised to decode the output to Unicode,
+    if they know its encoding.
+    """
+    buf = b""
+    while True:
+        line = await sr.readline()
+        if sr.at_eof():
+            break
+        if callback is not None:
+            callback(line)
+        # TODO: This would be a good place to log the received line, mostly for debug
+        # logging.
+        buf += line
+    return buf
+
+
+async def run_command(
+    args: List[str],
+    *,
+    error_message: str,
+    timeout_message: str,
+    timeout: Optional[float],
+    stdout_callback: Optional[Callable] = None,
+    stderr_callback: Optional[Callable] = None,
+) -> Tuple[bytes, bytes]:
+    """Run a command and get its output.
+
+    Run a command using asyncio.subprocess, consume its standard streams, and return its
+    output in bytes.
+
+    :raises RuntimeError: if the process returns a non-zero exit status
+    :raises TimeoutError: if the process times out
+    """
+    # Start the provided command, and return a handle. The command will run in the
+    # background.
+    proc = await asyncio.subprocess.create_subprocess_exec(
+        *args,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+
+    assert proc.stdout is not None
+    assert proc.stderr is not None
+
+    # Create asynchronous tasks that will consume the standard streams of the command,
+    # and call callbacks if necessary.
+    stdout_task = asyncio.create_task(read_stream(proc.stdout, stdout_callback))
+    stderr_task = asyncio.create_task(read_stream(proc.stderr, stderr_callback))
+
+    # Wait until the command has finished, for a specific timeout. Then, verify that the
+    # command has completed successfully. In any other case, raise an exception.
+    try:
+        ret = await asyncio.wait_for(proc.wait(), timeout=timeout)
+    except asyncio.exceptions.TimeoutError:
+        raise TimeoutError(timeout_message)
+    if ret != 0:
+        raise RuntimeError(error_message)
+
+    # Wait until the tasks that consume the command's standard streams have exited as
+    # well, and return their output.
+    stdout = await stdout_task
+    stderr = await stderr_task
+    return (stdout, stderr)
+
+
+class DangerzoneConverter:
+    def __init__(self) -> None:
+        self.percentage: float = 0.0
+
+    def calculate_timeout(
+        self, size: float, pages: Optional[float] = None
+    ) -> Optional[float]:
+        """Calculate the timeout for a command.
+
+        The timeout calculation takes two factors in mind:
+
+        1. The size (in MiBs) of the dataset (document, multiple pages).
+        2. The number of pages in the dataset.
+
+        It then calculates proportional timeout values based on the above, and keeps the
+        large one.  This way, we can handle several corner cases:
+
+        * Documents with lots of pages, but small file size.
+        * Single images with large file size.
+        """
+        if not int(os.environ.get("ENABLE_TIMEOUTS", 1)):
+            return None
+
+        # Do not have timeouts lower than 10 seconds, if the file size is small, since
+        # we need to take into account the program's startup time as well.
+        timeout = max(TIMEOUT_PER_MB * size, TIMEOUT_MIN)
+        if pages:
+            timeout = max(timeout, TIMEOUT_PER_PAGE * pages)
+        return timeout
+
+    @abstractmethod
+    async def convert(self) -> None:
+        pass
+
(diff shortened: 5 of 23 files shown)


deeplow:
CI is awfully broken. I'll can look into that shortly.

BUILD.md
```diff
@@ -102,12 +102,70 @@ Create a .rpm:
...
+3. Copy the built `dangerzone.rpm` to `fedora-37-dz`
+    `qvm-copy dist/dangerzone-0.4.1-1.noarch.rpm`
+
+Run on `fedora-37-dz`:
+1. `sudo dnf install -y ~/QubesIncoming/<DEV_VM>/dangerzone-*.rpm` (replace `<DEV_VM>`)
+2. Shut down the template `fedora-37-dz`
+
+Run `dangerzone-cli` on the `dz` qube (make sure you only start it after the template has shut down)
+
+## Initial Setup (apyrgio proposal)
```
deeplow, line 139: I like this proposal. I propose that we merge them and have two suggested approaches:
  - **fast doc-to-pixels iteration** - your setup
  - **installation testing** - the previous setup

Testing the installed RPM will be needed at some point as well. So we might leave instructions on how to set that up. Later when we have QA in place, that won't be as needed.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+# FIXME: Convert the following to asyncio code, so that there's no need to have a
+# separate function.
+async def main() -> int:
+    out_dir = Path("/tmp/dangerzone")
+    if out_dir.exists():
+        shutil.rmtree(out_dir)
+    out_dir.mkdir()
+
+    try:
+        untrusted_data = await read_bytes()
```
deeplow, line 69: Not sure why I called this one `untrusted_data`. It comes from the client, which is trusted.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -0,0 +1,172 @@
...
+import json
+import os
+import shutil
+import sys
+
+# try:
+#     from .common import DangerzoneConverter, run_command
+# except:
+#     # Importing without a module is needed when this runs in a container
+#     from common import DangerzoneConverter, run_command, running_on_qubes
```
deeplow, line 19: Shall we remove this?

dangerzone/gui/main_window.py
```diff
@@ -71,7 +72,9 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
             self.waiting_widget: WaitingWidget = WaitingWidgetContainer(self.dangerzone)
             self.waiting_widget.finished.connect(self.waiting_finished)
 
-        elif isinstance(self.dangerzone.isolation_provider, Dummy):
+        elif isinstance(self.dangerzone.isolation_provider, Dummy) or isinstance(
+            self.dangerzone.isolation_provider, Qubes
+        ):
             # Don't wait with dummy converter
```
deeplow, line 78: Update comment:
```suggestion
            # Don't wait with dummy converter and on Qubes
```

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+        # Reset hard-coded state
+        if os.path.exists(CONVERTED_FILE_PATH):
+            os.remove(CONVERTED_FILE_PATH)
+
+        percentage = 0.0
+
+        with open(document.input_filename, "rb") as f:
+            # TODO handle lack of memory to start qube
+            if getattr(sys, "dangerzone_dev", False):
+                # Use dz.ConvertDev RPC call instead, if we are in development mode.
```
deeplow, line 59: > @deeplow: This whole section feels a bit like ti could have its own function (update_code_in_dvm() or something)
> @apyrgio: Under a qubes module, most probably. You're right, I'd like something similar.


(from [here](https://github.com/freedomofpress/dangerzone/commit/ade3e2280a4303569626542302c3a770bf6726dd#r116691345))

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+                # module.
+                with zipfile.PyZipFile(temp_file, "w") as z:
+                    for py in pyfiles:
+                        z.writepy(py)
+
+                p = subprocess.Popen(
+                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.ConvertDev"],
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    startupinfo=get_subprocess_startupinfo(),
```
deeplow, line 80: Let's drop this line, since we're not on Windows

```suggestion
```

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+                p.stdin.write(bufsize_bytes)
+                p.stdin.write(temp_file.getvalue())
+                p.stdin.write(f.read())
+                p.stdin.close()
+            else:
+                p = subprocess.Popen(
+                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.Convert"],
+                    stdin=f,
+                    stdout=subprocess.PIPE,
+                    startupinfo=get_subprocess_startupinfo(),
```
deeplow, line 98:
```suggestion
```

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
```diff
@@ -14,36 +15,24 @@
...
+    write_bytes(text.encode(), file=file)
 
-def send(data):
-    """Qrexec wrapper for sending text data to the client
 
-    borrowed from https://github.com/QubesOS/qubes-app-linux-pdf-converter/blob/main/qubespdfconverter/server.py#L77
-    """
-    print(data, flush=True)
+def write_int(num: int, file=sys.stdout):
+    write_bytes(num.to_bytes(2, signed=False), file=file)
```
deeplow: Explicit endianness :

```suggestion
    write_bytes(num.to_bytes(2, byteorder="big", signed=False), file=file)
```

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+        shutil.move(CONVERTED_FILE_PATH, document.output_filename)
+        success = True
+
+        return success
+
+    def get_max_parallel_conversions(self) -> int:
+        return 1
+
+
+def running_on_qubes() -> bool:
```
deeplow, line 171: Replace with `dangerzone/conversion/common.py:running_on_qubes()` now that the ex-container code is now under the same python module.

dangerzone/logic.py
```diff
@@ -69,8 +69,10 @@ def convert_doc(document: Document) -> None:
             )
 
         max_jobs = self.isolation_provider.get_max_parallel_conversions()
-        with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
-            executor.map(convert_doc, self.documents)
+        # with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
+        #    executor.map(convert_doc, self.documents)
+        for document in self.documents:
+            convert_doc(document)
```
deeplow, line 75: There was nothing wrong with the original approach:
```suggestion
        with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
           executor.map(convert_doc, self.documents)
```

pyproject.toml
```diff
@@ -12,6 +12,7 @@ appdirs = "*"
 PySide6 = "^6.4.1"
 colorama = "*"
 pyxdg = {version = "*", platform = "linux"}
+python-magic = "^0.4.27"
```
deeplow, line 15: Unless we plan to run the dummy conversion for testing, I don't think we need this here.

qubes/dz.Convert
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+python -m dangerzone.conversion.doc_to_pixels_qubes_wrapper
```
deeplow, line 2: Add here a new line.

BUILD.md
apyrgio: In my mind, similar to the container isolation provider, we can distinguish between two types of environments:

1. The dev environment, which does not require an RPM, and can use the `dz.ConvertDev` RPC call. Instructions for this environment can be in `BUILD.md`.
2. The installation environment, which does require an RPM, and uses the `dz.Convert` RPC call. Instructions for this environment can be in `INSTALL.md`.

Devs can of course follow any of those two methods, but I think each of those have their time and place, and we need to test both.

Dockerfile
```diff
@@ -33,8 +33,12 @@ RUN mkdir tessdata && cd tessdata \
     && find . -name '*.traineddata' -maxdepth 2 -exec cp {} /usr/share/tessdata \; \
     && cd .. && rm -r tessdata
 
-COPY dangerzone.py /usr/local/bin/
-RUN chmod +x /usr/local/bin/dangerzone.py
+COPY dangerzone/conversion/common.py /usr/local/bin/
+COPY dangerzone/conversion/pixels_to_pdf.py /usr/local/bin/
+COPY dangerzone/conversion/doc_to_pixels.py /usr/local/bin/
```
apyrgio, line 38: At the end of the day, the command that runs is:

```
/usr/bin/python3 /usr/local/bin/pixels_to_pdf.py
```

Not that important, but I'd prefer not polluting the `/usr/local/bin` directory, and use a different one (e.g., `/opt/dangerzone`).

Dockerfile
```diff
@@ -33,8 +33,12 @@ RUN mkdir tessdata && cd tessdata \
...
+RUN chmod +x /usr/local/bin/common.py
+RUN chmod +x /usr/local/bin/pixels_to_pdf.py
+RUN chmod +x /usr/local/bin/doc_to_pixels.py
```
apyrgio, line 41: I think that these lines are not necessary, as we don't execute any of these files directly.

Dockerfile
apyrgio, line 1: General issue with our Dockerfile: the build context may be too big, if an image already exists in the `./share` dir. I suggest using as build context the code directory only.

dangerzone/cli.py
```diff
@@ -63,6 +65,8 @@ def cli_main(
 
     if getattr(sys, "dangerzone_dev", False) and dummy_conversion:
         dangerzone = DangerzoneCore(Dummy())
+    elif running_on_qubes() and os.environ.get("DZ_USE_CONTAINERS", "0") == "0":
```
apyrgio, line 68: Two things here:

1. Let's document this option in the Qubes OS section of `INSTALL.md`, so that users are aware.
2. I propose moving the logic that checks for the envvar in the `running_on_qubes()` function, so that it's used by the GUI as well.

apyrgio (changes requested):

Dockerfile
deeplow: How do we do that? Moving Dockerfile to our code directory? and then `docker build dangerzone/`?

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
apyrgio: D'oh, you're right! I'll change it.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
```diff
@@ -0,0 +1,104 @@
...
+    with open("/tmp/input_file", "wb") as f:
+        f.write(untrusted_data)
+
+    converter = DocumentToPixels()
+
+    try:
+        await converter.convert()
+    except Exception as e:
+        await write_text(f"Conversion failed with: {e}", file=sys.stderr)
+        return 1
```
apyrgio, line 84: I'm unsure if we even want to catch this exception here. The end result will be similar to the `except` clause, after all.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
apyrgio, line 1: Personally, I want to move all the Qubes wrapper code within `doc_to_pixels.py`, since we plan to use the stdio methods in containers as well. Not something that important for now, but something to keep in mind.

dangerzone/conversion/pixels_to_pdf.py
deeplow: Actually, we can't https://app.circleci.com/pipelines/github/freedomofpress/dangerzone/1392/workflows/cceaf5a5-c015-4b28-89fb-fdf24a4fb5f9/jobs/12814

Dockerfile
deeplow: Addressed [ab3ae26](https://github.com/freedomofpress/dangerzone/pull/437/commits/ab3ae26744476faffb759e79b1814b72e8dd29ae).

dangerzone/cli.py
deeplow: 1. :+1: 
2. I don't follow completely. Can you give an example?

dangerzone/conversion/pixels_to_pdf.py
apyrgio: Well, this try-import clause is bad, because it introduces ambiguity. But you're right that removing it breaks the container isolation provider. I think fixing the entry point, so that we can do `python -m` is much more cleaner. I'll take a look at it.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
apyrgio: I didn't specify this because big endian is pretty common nowadays, and the default in Python.

Dockerfile
apyrgio: The location of the Dockerfile does not dictate the build context actually. We can just run `docker build dangerzone/` from the top dir, as you suggested. I'll send a commit for this, and you can revert it if you want to.

dangerzone/cli.py
apyrgio: Sure. What I'm saying is that with the code above, only the Dangerzone CLI is affected. If you are in Qubes and you want to run the Dangerzone GUI with containers, you need the same code in the `gui/__init__.py` module.

Better yet, to avoid duplication, we can move this check (`os.environ.get("DZ_USE_CONTAINERS", "0")`) within `running_on_qubes`:

```python
  def running_on_qubes() -> bool:
      # https://www.qubes-os.org/faq/#what-is-the-canonical-way-to-detect-qubes-vm
      if os.environ.get("DZ_USE_CONTAINERS", "0") == "0":
          return os.path.exists("/usr/share/qubes/marker-vm")
      else:
          return False
```

dangerzone/conversion/common.py
```diff
@@ -0,0 +1,140 @@
...
+    stdout = await stdout_task
+    stderr = await stderr_task
+    return (stdout, stderr)
+
+
+class DangerzoneConverter:
+    def __init__(self) -> None:
+        self.percentage: float = 0.0
+
+    def calculate_timeout(
```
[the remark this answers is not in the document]
deeplow: I think the idea is to try and do this client-side. But we can only do so when the containers code streams pages. But while that's not done, I think we can leave it as-is, no?

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+CONVERTED_FILE_PATH = (
+    "/tmp/safe-output-compressed.pdf"  # FIXME won't work for parallel conversions
+)
+
+
+class Qubes(IsolationProvider):
+    """Uses a disposable qube for performing the conversion"""
+
+    def install(self) -> bool:
+        pass
```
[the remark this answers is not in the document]
deeplow: I agreen in principle in having pre-flight checks here, but I'm not sure how to approach this. To check if there's anything in a dispVM already installed we need to start a disposable qube. Maybe only place we can do this is on disp qube startup. Error out if it doesn't check one of the conditions (e.g. is online).

dangerzone/isolation_provider/qubes.py
deeplow: Let's leave this as-is and tackle it later via https://github.com/freedomofpress/dangerzone/issues/442

pyproject.toml
deeplow: You're right. I used this as a temporary solution and forgot to remove it.

deeplow: Fixed via commit `543f0c9`

qubes/dz.Convert
deeplow: Done in commit https://github.com/freedomofpress/dangerzone/pull/437/commits/3c595ea20da95dadac10f2d308131fb70397669d

dangerzone/conversion/pixels_to_pdf.py
deeplow: I reverted it to the try...import clause so it would run (https://github.com/freedomofpress/dangerzone/commit/e0e5c923a3ba686214a169cb9c2778b7546dc4fa). Leaving this open in case you want to implement another solution.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
deeplow: Solved in https://github.com/freedomofpress/dangerzone/pull/437/commits/5bb695f91c57805375b6473bf6097a9435b86436

Dockerfile
deeplow: Things were already in `/usr/local/bin`. We can change it here, sure, but it may be out of scope of the PR. But it's pretty minor if it slips in. However you prefer.

Dockerfile
deeplow: Thanks for the clarification. Then let's consider this done in your commit https://github.com/freedomofpress/dangerzone/pull/437/commits/ac4a81b26865af1b3b30467ec203f5a72c09bdc6.

dangerzone/cli.py
deeplow: Done in 13aaeda. I also documented this on `BUILD.md`. We don't yet have a section for Qubes on `INSTALL.md`.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
deeplow: Let's address this later. I created an issue for it https://github.com/freedomofpress/dangerzone/issues/443 and made it as an optional step to the beta integration https://github.com/freedomofpress/dangerzone/issues/412

dangerzone/conversion/common.py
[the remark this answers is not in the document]
deeplow: Maybe this can be addressed in https://github.com/freedomofpress/dangerzone/issues/443?

dangerzone/conversion/common.py
```diff
@@ -0,0 +1,140 @@
...
+        if pages:
+            timeout = max(timeout, TIMEOUT_PER_PAGE * pages)
+        return timeout
+
+    @abstractmethod
+    async def convert(self):
+        pass
+
+    def update_progress(self, text: str, *, error: bool = False) -> None:
+        return
```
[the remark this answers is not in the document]
deeplow: I think it's now working well for containers and for Qubes:

https://github.com/freedomofpress/dangerzone/pull/437/files#diff-9efd70f53282d2854b01b8df2280b506873ade281dd07b7c6d7d132bd020a443R136-R143

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+    def _convert(
+        self,
+        document: Document,
+        ocr_lang: Optional[str],
+        stdout_callback: Optional[Callable] = None,
+    ):
+        success = False
+
+        # FIXME won't work on windows, nor with multi-conversion
+        out_dir = Path("/tmp/dangerzone")
```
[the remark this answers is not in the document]
deeplow: Closing this as this will be addressed in 0.5.1 https://github.com/freedomofpress/dangerzone/issues/436

Dockerfile
deeplow: Anyways, this is pretty minor. Slipped it in https://github.com/freedomofpress/dangerzone/pull/437/commits/402cc87e8866a27fd13f6b5a68bbf39ae5e23468.

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,173 @@
...
+            else:
+                p = subprocess.Popen(
+                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.Convert"],
+                    stdin=f,
+                    stdout=subprocess.PIPE,
+                    startupinfo=get_subprocess_startupinfo(),
+                )
+
+            untrusted_n_pages = p.stdout.read(2)
+            n_pages = int.from_bytes(untrusted_n_pages, byteorder="big", signed=False)
```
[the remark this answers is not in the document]
deeplow: Addressed in https://github.com/freedomofpress/dangerzone/pull/437/commits/6ebd9592042e434b75a9a9660ba43b8bff7c92e4. `readline` wasn't needed.

Dockerfile
apyrgio: You're right, we always had code in `/usr/local/bin`, but now we want to have something like a Python package there, which introduces problems like this: https://github.com/freedomofpress/dangerzone/pull/437#discussion_r1226470694

So, these are new issues that I'd like to nip them in the bud. I'll try to tackle this before we merge this PR, so let's keep this discussion open.

dangerzone/conversion/common.py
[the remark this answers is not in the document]
apyrgio: Nice, that works as well, and we will properly fix this once we tackle #429.

dangerzone/conversion/common.py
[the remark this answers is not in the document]
apyrgio: My thinking was that if we leave it as is, folks won't be able to disable timeouts on Qubes. Given that Qubes use disposable VMs for conversion, I'm afraid that timeouts may be more common in that platform.

We can address it in #443, if we make it part of 0.5.0. Have we committed to this though? I'm not sure.

That's why I've created a different issue instead (#446), with a narrower scope, so that we'll certainly tackle it in on 0.5.0. In the meantime, we can resolve this discussion and continue there.

dangerzone/isolation_provider/qubes.py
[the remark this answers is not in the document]
apyrgio: Hm, you're right, this is not the time and place for this. I've opened a separate issue instead that we can work on once we have the PoC merged: https://github.com/freedomofpress/dangerzone/issues/447

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
deeplow: You're right. https://github.com/freedomofpress/dangerzone/pull/437/commits/a7403eefa0b5569c2cea93d0f86aa9b20dd1039f

Please it if you feel it was addressed.

dangerzone/isolation_provider/qubes.py
deeplow: > What would work is giving this function the write end of the pipe, so that it can write to it.

OK. Feel free to adapt it or write it from scratch.

BUILD.md
```diff
@@ -102,11 +102,193 @@ Create a .rpm:
...
+   dz.ConvertDev      *       @anyvm       @dispvm:dz-dvm  allow
+   ```
+
+#### In the `fedora-37-dz` template
+
+1. Install dependencies:
+
+   ```
+   sudo dnf install -y rpm-build pipx qt5-qtbase-gui libreoffice python3-magic \
+       tesseract
```
deeplow, line 178: Just in case the developer want to test OCR on non-english languages
```suggestion
       tesseract*
```

deeplow (commented):
I quite like the simplification of the "code teleport" part. And I spotted a potential issue with the Dockerfile change.

BUILD.md
deeplow: Let's close this since this has been addressed and closed in https://github.com/freedomofpress/dangerzone/pull/437/commits/71b8f674e01b3ce0fd4f36d56e28756f1c0d8c9d.

dangerzone/conversion/pixels_to_pdf.py
apyrgio: We now have a fix for that in 8841d534. We install Dangerzone in the container and disposable VM in such a way that we can import it as a package, meaning that relative imports work now everywhere.

dangerzone/isolation_provider/qubes.py
apyrgio: We now have code for this in 5456f8b2. It should tackle the buffer size limitation.

dangerzone/conversion/doc_to_pixels_qubes_wrapper.py
apyrgio: I forgot to reply to this. Yes, this works great. Resolving this discussion.

INSTALL.md
```diff
@@ -117,8 +118,24 @@ After confirming that it matches, type `y` (for yes) and the installation should
 
 </details>
 
+### Qubes OS
```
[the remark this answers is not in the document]
apyrgio: I have created a separate issue for that (https://github.com/freedomofpress/dangerzone/issues/451), to see what we want to do with the behavior of Dangerzone on Qubes, while it's still on alpha. I believe I echo your sentiment, so let's continue the discussion there.

BUILD.md
apyrgio: Very cool! I didn't know that `dnf` supported globbing. Added it in a4390ea24f7804e740504f6c07eb69e5e47e2eee.

Zsh users may have a [slight problem](https://discussion.fedoraproject.org/t/using-to-search-install-a-package-with-dnf/73095/2) with it, but hopefully they'll adapt.

Dockerfile
```diff
@@ -33,8 +33,10 @@ RUN mkdir tessdata && cd tessdata \
     && find . -name '*.traineddata' -maxdepth 2 -exec cp {} /usr/share/tessdata \; \
     && cd .. && rm -r tessdata
 
-COPY dangerzone.py /usr/local/bin/
-RUN chmod +x /usr/local/bin/dangerzone.py
+ENV PYTHONPATH=/opt/dangerzone
+
+RUN mkdir -p /opt/dangerzone
+COPY . /opt/dangerzone/dangerzone
```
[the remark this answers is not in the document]
deeplow: Sounds good.

deeplow (commented):
Good to merge

Dockerfile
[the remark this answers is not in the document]
apyrgio: We discussed this offline and we're good to go. Resolving this discussion.

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,180 @@
...
+        with open(document.input_filename, "rb") as f:
+            # TODO handle lack of memory to start qube
+            if getattr(sys, "dangerzone_dev", False):
+                # Use dz.ConvertDev RPC call instead, if we are in development mode.
+                # Basically, the change is that we also transfer the necessary Python
+                # code as a zipfile, before sending the doc that the user requested.
+                p = subprocess.Popen(
+                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.ConvertDev"],
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
```
deeplow, line 76: Important thing before we merge this that I just noticed: we need to send the `stderr`
```python
    stderr=subprocess.DEVNULL,
```

dangerzone/isolation_provider/qubes.py
```diff
@@ -0,0 +1,180 @@
...
+                self.teleport_dz_module(p.stdin)
+
+                # Finally, send the document, as in the normal case.
+                p.stdin.write(f.read())
+                p.stdin.close()
+            else:
+                p = subprocess.Popen(
+                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.Convert"],
+                    stdin=f,
+                    stdout=subprocess.PIPE,
```
deeplow, line 90: same here
```suggestion
                    stdout=subprocess.PIPE,
                    stderr=subprocess.DEVNULL,
```

deeplow (commented):
Found one more issue here.

dangerzone/isolation_provider/qubes.py
deeplow: Yes, we can change it later. The thing is, if we don't send it to DEVNULL and we end up sending it through stdout and forgetting about this, the attacker can send arbitrary data to client.

I think the Qubes folks removed control characters already, but for a time this was a possible attack vector.

apyrgio: Yeap, I agree with this attack vector. Let's consume stderr properly in #430 then.

deeplow: OK. I've added it to checklist there so we don't forget.

apyrgio (approved):

PR #439, opened 2023-06-08
Add "change document selection" button

### LINKED ISSUE #428: Add "Change selection" button near "x document(s) selected"
## User story

If I'm unsure if I've selected the right file for conversion, or if I've definitely selected the wrong one, I want to be able to change my selection, so that I'm not converting the wrong file.

## Background

Ideally we'll have a progressive disclosure display of all selected files, but as an immediate step, we've discussed previously that this would be a significant improvement already that does not require UX work.

# Notes / Evidence (added by @deeplow)

During user testing, a participant wanted to change the document selection because they had clicked on the wrong document but they couldn't. Restarting the application turned out the be the only way to do so.

deeplow:
Fixes #428, by adding a button to change the selection of documents:

copy

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -562,7 +562,9 @@ class DocSelectionWidget(QtWidgets.QWidget):
             self.file_dialog.setDirectory(first_doc_dir)
 
         if self.file_dialog.exec():
-            documents = [Document(filename) for filename in self.file_dialog.selectedFiles()]
+            documents = [
+                Document(filename) for filename in self.file_dialog.selectedFiles()
+            ]
             self.documents_selected.emit(documents)
         else:
             # No files selected
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -7,7 +7,7 @@ import subprocess
 import tempfile
 import typing
 from multiprocessing.pool import ThreadPool
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from colorama import Fore, Style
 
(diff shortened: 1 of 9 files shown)


deeplow:
Just tested on windows and it runs fine. We'll have to understand why the windows and mac CI are getting into a recursion issue.

dangerzone/gui/main_window.py
```diff
@@ -276,6 +279,15 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
...
 
     def documents_selected(self, new_docs: List[Document]) -> None:
+        # Get previously selected documents
+        previous_docs = self.dangerzone.get_unconverted_documents()
+        docs_removed = []
+        for prev_doc in previous_docs:
+            if prev_doc not in new_docs:
+                docs_removed.append(prev_doc)
+        for doc in docs_removed:
+            self.dangerzone.remove_document(doc)
```
apyrgio, line 289: General question, is the removal logic actually needed? I'm asking because I'd assume that clicking the "Change Selection" button would simply remove all the previously selected documents, so there would be no need for removing duplicates, `Document` instance comparison, etc.

tests/__init__.py
```diff
@@ -9,6 +9,7 @@
 
 SAMPLE_DIRECTORY = "test_docs"
 BASIC_SAMPLE = "sample-pdf.pdf"
+BASIC_SAMPLE2 = "sample-doc.doc"
```
apyrgio, line 12: Ideally, I'd like to not have variables with a number appended to them, as it's not easy to reason about their purpose. Since it's used in a single test, I'd propose moving this logic there.

apyrgio (commented):
Looks good in general, but I have a question for the removal logic that affects a big portion of this PR. Once we resolve it, we can merge this PR.

dangerzone/gui/main_window.py
deeplow: The issue is that the UI doesn't "react" when things change. So we need to register the removal so that the documents list widget gets the info that certain objects need to removed. Does this make sense?

tests/__init__.py
deeplow: I agree. I renamed `sample_doc` to `sample_pdf` and `sample_doc2` to `sample_doc`. I sneaked in another commit for passing `sample_pdf` as a fixture instead of as a TestBase.sample_doc as well. I hope you don't mind.

dangerzone/gui/main_window.py
apyrgio: Sure I get it, but my question is why can't we just nuke the previously selected documents. For example:

```suggestion
        self.dangerzone.documents = []
        self.documents_list.docs_list = []
        self.documents_list.docs_list_widget_map = {}
        self.documents_list.clear()
```

deeplow:
I had to rebase since I was having some issues rebuilding the image with my faster image build script and they were not worth debugging because it already works on main.

dangerzone/gui/main_window.py
deeplow: I have now implemented an iteration of this in https://github.com/freedomofpress/dangerzone/pull/439/commits/0916344c2ed51bb5f7f7dc7fbd7bf2d38f57e607. The diff looks a bit more complex because I changed the order in `ContentWidget.documents_selected`, but it is essentially the same.

I implemented the following in `documents_list` itself, overriding the clear() method
```
        self.documents_list.docs_list = []
        self.documents_list.docs_list_widget_map = {}
```

apyrgio: Perfect, thanks a lot for this. I think we can also remove the code for checking document equality as well, but I'll leave it up to you.

apyrgio:
One last problem that I see is with our CI. It seems that the Pytest wrapper change breaks our tests.

apyrgio (approved):
Looks good. Feel free to squash and merge.

PR #450, opened 2023-06-19
Make progress reports work on Qubes

### LINKED ISSUE #429: Qubes: Progress reports

deeplow:
Fixes #429. Depends on #437.

### DIFF
--- a/dangerzone/conversion/common.py
+++ b/dangerzone/conversion/common.py
@@ -100,8 +100,9 @@ async def run_command(
 
 
 class DangerzoneConverter:
-    def __init__(self) -> None:
+    def __init__(self, progress_callback: Optional[Callable] = None) -> None:
         self.percentage: float = 0.0
+        self.progress_callback = progress_callback
 
     def calculate_timeout(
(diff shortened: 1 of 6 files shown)


apyrgio:
General comment:

I just realized that some of our progress messages are wrong for Qubes:

https://github.com/freedomofpress/dangerzone/blob/e989069712e41f4721aa2f1fba4e49d70cded7ef/dangerzone/isolation_provider/qubes.py#L122-L131

Now that we will show them in the GUI, it's best to fix them as well.

deeplow:
> there's no change in this PR, right? I mean, we already reported progress reports as soon as we read each page, right?

That's right. that part already happened.

Issue #243, opened 2022-11-03
Feature Request: Convert HWP files (Hancom Office)


CR0CKER:
It would be hugely helpful if Dangezone was capable of converting HWP files produced by Hancom Office, a terrible office suite that's extremely widely used in South Korea and frequently used for targeted malware attacks against journalists, researchers, defectors, and politicians in South Korea.

apyrgio:
Thanks a lot for bringing this to our attention. It definitely looks like something Dangerzone should support.

Looking into HWP, I see that there are several open source projects that are trying to convert HWP files to ODT:

* https://github.com/mete0r/pyhwp
* https://github.com/k2webtech/hwp2odt

Have you tried any of them, or know if something else is more reliable? If we can rely on such a tool, we can use it as a pre-conversion step for HWP files, to grab an ODT document, and then let Dangerzone continue the conversion process as usual.

CR0CKER:
Thanks for flagging these -- I'll have a look. So far I've only tried the HWP import in LibreOffice but compatibility was very poor, mostly because the HWP format apparently changed significantly in recent years and Hancom -- surprise -- has no interest in properly documenting the format.

apyrgio:
Duly noted. Most probably this won't be part of the upcoming 0.4.0 release, but it would be nice providing support for 0.5.0 onwards. I'll keep this issue open to track this request.

One more thing I'd like to ask is what else would help Dangerzone users in South Korea? Are translations, alternative installation sources or stuff like that important to users in South Korea? Are other things more important?

CR0CKER:
While not important to me, a Korean interface would certainly be useful. I'll follow up with further suggestions once I have more experience using this.

CR0CKER:
Unfortunately I have not been able to test Dangerzone because converting a PDF fails with the following error message: 
<img width="593" alt="Screenshot 2022-11-17 at 2 51 42 PM" src="https://user-images.githubusercontent.com/6056387/202367267-ade25765-4741-492b-a16e-f13417eb969a.png">

Any idea what may cause this? Docker installed fine.

apyrgio:
Interesting. It seems that the Docker daemon runs (else Dangerzone would have complained at startup), but the Dangerzone container does not start.

Can you please open a terminal, run the following commands, and provide the output here?

```
sudo docker info
sudo docker run hello-world
```

CR0CKER:
Here's what I got:

```
Client:
 Context:    default
 Debug Mode: false
[...]

PR #459, opened 2023-06-26
Shrink container image with --no-cache option on apk

OctopusET:
This option shrinks container image more than half.
result: 1.95 GB -> 882 MB


apyrgio:
Hey @OctopusET, thanks a lot for the contribution. We really want to reduce the size of the container image, so I tested it locally as well. I found that the original image from `main` was 838MiB, and with your changes, it got reduced to 836MiB.

I was confused a bit so I checked out the size of the APK index under `/var/cache/apk`, and in my case, it was just 2MiB, which explains the size difference:

```console
$ ls -lh /var/cache/apk/
total 2M     
-rw-r--r--    1 root     root      422.1K Jun 26 11:02 APKINDEX.49104001.tar.gz
-rw-r--r--    1 root     root        1.4M Jun 26 11:02 APKINDEX.884188d8.tar.gz
```

So, I have two questions here:

1. Could it be that the size difference that you're seeing is attributed to disabling image compression the first time? 1.95 GiB is the size of the uncompressed image, that's why I'm asking.
2. Using `--no-cache` is indeed a reasonable setting for container images, and it uncovered an issue in our Dockerfile, if we also add it to `apk add`. If you're ok with the following code, can you add it to your branch, so that I can merge it?

```diff
diff --git a/Dockerfile b/Dockerfile
index c247982..5e0f1c6 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG TESSDATA_CHECKSUM=990fffb9b7a9b52dc9a2d053a9ef6852ca2b72bd8dfb22988b0b990a70
 
 # Install dependencies
 RUN apk --no-cache -U upgrade && \
-    apk add \
+    apk --no-cache add \
     ghostscript \
     graphicsmagick \
     libreoffice \
@@ -26,7 +26,6 @@ RUN apk --no-cache -U upgrade && \
 RUN mkdir tessdata && cd tessdata \
     && TESSDATA_VERSION=$(wget -O- -nv https://api.github.com/repos/tesseract-ocr/tessdata/releases/latest \
         | sed -n 's/^.*"tag_name": "\([0-9.]\+\)".*$/\1/p') \
-    && apk --purge del jq \
     && wget https://github.com/tesseract-ocr/tessdata/archive/$TESSDATA_VERSION/tessdata-$TESSDATA_VERSION.tar.gz \
     && echo "$TESSDATA_CHECKSUM  tessdata-$TESSDATA_VERSION.tar.gz" | sha256sum -c \
     && tar -xzvf tessdata-$TESSDATA_VERSION.tar.gz -C . \
```

OctopusET:
Hello, thank you for the comment.

1. I'm not sure, but I checked the size with `podman images`. I just tried re-build whole image without/with my PR. It still shows 1.9GB. I think `--no-cache` option doesn't help shrinking size that much. I think I got around 900MB because podman somehow used the compressed image or it's just part of the image. But still I think `--no-cache` option would be good for the container image maintenance.

2. Thank you for the changes, I forgot to add it. I just added it.

It's not related to this PR.

This PR was made during I'm working on the adding a hwp supports.
But I have a problem, I managed to convert hwp, hwpx to pdf with extension (I mentioned on this [comment](https://github.com/freedomofpress/dangerzone/issues/243#issuecomment-1571458447).) in the container. But dangerzone fails when it's running pdfinfo.
I'd like to discuss about this, should I just open Draft PR first?

apyrgio:
> This PR was made during I'm working on the adding a hwp supports.

That's exciting!

> But I have a problem, I managed to convert hwp, hwpx to pdf with extension (I mentioned on this https://github.com/freedomofpress/dangerzone/issues/243#issuecomment-1571458447.) in the container. But dangerzone fails when it's running pdfinfo.
> I'd like to discuss about this, should I just open Draft PR first?

Sure, go ahead.

apyrgio (approved):

OctopusET:
I just posted my draft PR #460!

Just curious, what was that line for?
```
&& apk --purge del jq
```

apyrgio:
> I just posted my draft PR #460!

Great, thanks! I'll comment there soon.

> Just curious, what was that line for?
> 
> ```
> && apk --purge del jq
> ```

Oh, good question. In order to get the Tesseract data version, I initially installed `jq`, parsed the version with it, and then deleted it. Once I figured out we can use `sed` instead of `jq`, I ditched the `jq` approach, but not the line that removes it.

This didn't cause any problem, up to the point where we removed the APK index. By doing so, this line would fail. So, I removed it.

PR #461, opened 2023-06-27
Add ubuntu 23.04 (lunar) support

### LINKED ISSUE #452: Add support for Debian Trixie
Debian 13 ("Trixie") is the new testing release of Debian. Dangerzone supported Debian Bookworm when it was in testing phase, so we need to support this new release as well.

deeplow:
The Ubuntu 23.04 docker image includes a user by default (ubuntu) which overtakes the 1000 uid and so our user becomes 1001 which makes the user directory unwritable. The solution as suggested in [1] was to remove that user.

[1]: https://bugs.launchpad.net/cloud-images/+bug/2005129

Fixes #452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 - Platform support: Alpha integration with Qubes OS ([issue #411](https://github.com/freedomofpress/dangerzone/issues/411))
 - Platform support: Debian Trixie (13)
+- Platform support: Ubuntu 23.04 (Lunar Lobster)
 
 ### Removed
 
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -70,6 +70,13 @@ RUN . /etc/os-release \
         | apt-key add -
 """
 
+# XXX: overcome the fact that ubuntu images (starting on 23.04) ship with the 'ubuntu'
+# user by default https://bugs.launchpad.net/cloud-images/+bug/2005129
+# Related issue https://github.com/freedomofpress/dangerzone/pull/461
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 ### Added
 
 - Platform support: Alpha integration with Qubes OS ([issue #411](https://github.com/freedomofpress/dangerzone/issues/411))
+- Ubuntu 23.04 (Lunar Lobster)
```
apyrgio, line 13: Minor, "Platform support:" missing.

dev_scripts/env.py
```diff
@@ -70,6 +70,10 @@
         | apt-key add -
 """
 
+DOCKERFILE_UBUNTU_2304_REM_USER = r"""
+RUN touch /var/mail/ubuntu && chown ubuntu /var/mail/ubuntu && userdel -r ubuntu
```
apyrgio, line 74: One small request here, if you can please link to either a GitHub issue of ours, or an external link, to explain why this exception is necessary.

apyrgio (changes requested):
Nice, the changes look good. Same ask as in the PR for Debian Trixie though, can you check that Ubuntu Lunar can install the produced .deb?

Also, I think we should slightly update `RELEASE.md`, now that the latest Ubuntu version is Lunar.

dev_scripts/env.py
deeplow: Sure. I had added it in the commit message but I agree that it's best left in the code. Fixed in https://github.com/freedomofpress/dangerzone/pull/461/commits/8dbb2fe9e04f76c793960ec47989dd0748840fea.

apyrgio (approved):

Issue #458, opened 2023-06-22
QA for 0.4.2


apyrgio:
To ensure that new releases do not introduce regressions, and support existing and newer platforms, we have to run some QA tests. Copied from [RELEASE.md](https://github.com/freedomofpress/dangerzone/blob/main/RELEASE.md):

- [ ] In `.circleci/config.yml`, add new platforms and remove obsolete platforms
- [ ] Make sure that the tip of the `main` branch passes the CI tests.
- [ ] Create a test build in Windows and make sure it works:
  - [ ] Create a new development environment with Poetry.
  - [ ] Build the container image and ensure the development environment uses
    the new image.
  - [ ] Run the Dangerzone tests.
  - [ ] Build and run the Dangerzone .exe
  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
- [ ] Create a test build in macOS (Intel CPU) and make sure it works:
  - [ ] Create a new development environment with Poetry.
  - [ ] Build the container image and ensure the development environment uses
    the new image.
  - [ ] Run the Dangerzone tests.
  - [ ] Create and run an app bundle.
  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
- [ ] Create a test build in macOS (M1/2 CPU) and make sure it works:
  - [ ] Create a new development environment with Poetry.
  - [ ] Build the container image and ensure the development environment uses
    the new image.
  - [ ] Run the Dangerzone tests.
  - [ ] Create and run an app bundle.
  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
- [ ] Create a test build in the most recent Ubuntu LTS platform (Ubuntu 22.10
  as of writing this) and make sure it works:
  - [ ] Create a new development environment with Poetry.
  - [ ] Build the container image and ensure the development environment uses
    the new image.
  - [ ] Run the Dangerzone tests.
  - [ ] Create a .deb package and install it system-wide.
  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
- [ ] Create a test build in the most recent Fedora platform (Fedora 37 as of
  writing this) and make sure it works:
  - [ ] Create a new development environment with Poetry.
  - [ ] Build the container image and ensure the development environment uses
    the new image.
  - [ ] Run the Dangerzone tests.
  - [ ] Create an .rpm package and install it system-wide.
  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).

[...]

PR #464, opened 2023-06-29
Use containers in Qubes by default & add Qubes QA instructions

### LINKED ISSUE #451: Use container isolation provider in Qubes, until beta support
The alpha support for Qubes (#411) is around the corner, and while it's a feature that any user can test, we consider it not stable enough for daily use. Instead, we believe that it will be helpful to most users once beta support (#412) is implemented as well.

If we cut a new Dangerzone release before fixing #412, we need to think of the existing Qubes users that rely on the container isolation provider, and may want to stick with that, until Qubes support gets more stable.

In that case, we can change the default value of `DZ_USE_CONTAINERS` to `"1"` instead of `"0"`, and then switch back once #412 is merged.  Note that while doing so, we may need to differentiate between checking if the code runs in a disposable VM (useful in `dangerzone.conversion.doc_to_pixels`), and deciding if the conversion *should* run in a disposable VM (useful in `dangerzone.isolation_provider.qubes`). Currently we have a single function (`running_on_qubes()`) that does both, but this is not correct.

deeplow:
Fixes #451 

**DONE** :warning: requires updating https://github.com/freedomofpress/dangerzone/issues/458 with the new QA instructions

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -104,23 +104,13 @@ Create a .rpm:
 
 ## Qubes OS
 
-<details>
-  <summary><i>:memo: Expand this section if you want to use containers instead of disposable qubes.</i></summary>
-  </br>
-
-  Create a Debian or Fedora-based development standalone qube with at least
-  8GB of private storage space, and follow the relevant instructions above for
-  the respective template.
-
-  Remember to set the environment variable `DZ_USE_CONTAINERS=1`, before executing
-  Dangerzone.
-
-  Over time, you may need to increase disk space or prune outdated container
-  images if you run into build issues on this VM.
-</details>
 
 > :warning: Native Qubes support is in alpha stage, so the instructions below
 > require switching between qubes, and are subject to change.
+>
+> If you want to build Dangerzone on Qubes and use containers instead of disposable
+> qubes, please follow the intructions of Fedora / Debian instead.
+
 
 ### Initial Setup
 
@@ -248,10 +238,10 @@ can run the following commands in the `dz` app qube:
 poetry shell
 
 # run the CLI
-./dev_scripts/dangerzone-cli --help
+QUBES_CONVERSION=1 ./dev_scripts/dangerzone-cli --help
 
 # run the GUI
-./dev_scripts/dangerzone
+QUBES_CONVERSION=1 ./dev_scripts/dangerzone
 ```
 
 Create a .rpm:
--- a/dangerzone/conversion/common.py
+++ b/dangerzone/conversion/common.py
@@ -19,10 +19,7 @@ TIMEOUT_MIN: float = 60  # (seconds)
 
 def running_on_qubes() -> bool:
     # https://www.qubes-os.org/faq/#what-is-the-canonical-way-to-detect-qubes-vm
-    if os.environ.get("DZ_USE_CONTAINERS", "0") == "0":
-        return os.path.exists("/usr/share/qubes/marker-vm")
-    else:
-        return False
+    return os.path.exists("/usr/share/qubes/marker-vm")
 
 
 async def read_stream(
--- a/dev_scripts/qa.py
+++ b/dev_scripts/qa.py
@@ -55,7 +55,20 @@ and newer platforms, we have to do the following:
   - [ ] Run the Dangerzone tests.
   - [ ] Create an .rpm package and install it system-wide.
   - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
-
+- [ ] Create a test build in the most recent Fedora template in Qubes OS (Fedora 38 as of
+  writing this) and make sure it works:
+  - [ ] Create a new development environment with Poetry.
+  - [ ] Build the container image and ensure the development environment uses
+        the new image.
+  - [ ] Run the Dangerzone tests and confirm that the tests are ran using containers
+        instead of disposable qubes
+  - [ ] Run the Dangerzone tests with the env var `QUBES_CONVERSION=1` and confirm that
+        the tests are ran using diposable qubes instead of containers
+  - [ ] Create an .rpm with `./install/linux/build-rpm.py` package and
+        install it in a fedora qube and make sure conversions run with containers
+  - [ ] Create an .rpm with `./install/linux/build-rpm.py --qubes` package and
+        install it following the instructions in `BUILD.md`
+  - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
 """
 
 CONTENT_QA_SCENARIOS = r"""### Scenarios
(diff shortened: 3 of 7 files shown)


dangerzone/conversion/common.py
```diff
@@ -19,8 +19,27 @@
 
 def running_on_qubes() -> bool:
     # https://www.qubes-os.org/faq/#what-is-the-canonical-way-to-detect-qubes-vm
-    if os.environ.get("DZ_USE_CONTAINERS", "0") == "0":
-        return os.path.exists("/usr/share/qubes/marker-vm")
+    return os.path.exists("/usr/share/qubes/marker-vm")
+
+
+def qubes_native_conversion() -> bool:
```
apyrgio, line 25: Τhe purpose of this function is to check if the conversion should run in a disposable qube. With that in mind, it probably won't run within the conversion sandbox. So I'd propose moving it to the Qubes isolation provider module.

dangerzone/conversion/common.py
```diff
@@ -134,7 +153,7 @@ async def convert(self) -> None:
         pass
 
     def update_progress(self, text: str, *, error: bool = False) -> None:
-        if not running_on_qubes():
+        if not qubes_native_conversion():
```
apyrgio, line 156: I think we should keep `running_on_qubes()` here, since we are in the sandbox environment. Granted, the result is the same, but that may change.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -146,7 +146,7 @@ async def convert(self) -> None:
         self.update_progress("Safe PDF created")
 
         # Move converted files into /safezone
-        if not running_on_qubes():
+        if not qubes_native_conversion():
```
apyrgio, line 149: Same here, we should keep `running_on_qubes()`.

dev_scripts/qa.py
```diff
@@ -55,7 +55,20 @@
...
   - [ ] Create an .rpm package and install it system-wide.
   - [ ] Test some QA scenarios (see [Scenarios](#Scenarios) below).
-
+- [ ] Create a test build in the most recent Fedora template in Qubes OS (Fedora 38 as of
+  writing this) and make sure it works:
+  - [ ] Create a new development environment with Poetry.
+  - [ ] Build the container image and ensure the development environment uses
+        the new image.
+  - [ ] Run the Dangerzone tests and confirm that the tests are ran using diposable qubes
+        instead of containers
```
apyrgio, line 64: You probably mean containers instead of disposable qubes, right?

dangerzone/conversion/common.py
```diff
@@ -19,8 +19,27 @@
...
+    return os.path.exists("/usr/share/qubes/marker-vm")
+
+
+def qubes_native_conversion() -> bool:
+    """Returns True if the conversion should be run using Qubes OS's diposable
+    VMs and False if not."""
+    if running_on_qubes():
+        if os.environ.get("VM_CONVERSION", "0") == "1":
+            return True
+        elif not getattr(sys, "dangerzone_dev", False):
```
apyrgio, line 31: I'm not sure I get why we need to check if the code runs in dev mode. Can you explain it please?

BUILD.md
```diff
@@ -248,10 +238,10 @@ can run the following commands in the `dz` app qube:
 poetry shell
 
 # run the CLI
-./dev_scripts/dangerzone-cli --help
+VM_CONVERSION=1 ./dev_scripts/dangerzone-cli --help
```
apyrgio, line 241: Given that we follow Qubes terminology (where VM -> Qube), shouldn't we do something similar here? I.e., `QUBES_CONVERSION` instead of `VM_CONVERSION`?

apyrgio (changes requested):

dangerzone/conversion/common.py
deeplow: This is the case where the user installed the package. We want to know if they installed the package with a qubes built target (i.e. if it was using `build-rpm.py --qubes`) or if it is just a fedora version, which just happens to be running on Qubes.

But now looking again at the logic it feels a bit convoluted.

BUILD.md
deeplow: Good point. Addressed in [dd96de3](https://github.com/freedomofpress/dangerzone/pull/464/commits/dd96de32eb181b6a954c35f08fa2b31335be2008)

dangerzone/conversion/pixels_to_pdf.py
deeplow: done in [6ec1b0b](https://github.com/freedomofpress/dangerzone/pull/464/commits/6ec1b0b15e6824c34a039540d4e9331773c5253c)

dangerzone/conversion/common.py
deeplow: Done in [6ec1b0b](https://github.com/freedomofpress/dangerzone/pull/464/commits/6ec1b0b15e6824c34a039540d4e9331773c5253c)

dangerzone/conversion/common.py
deeplow: Done in [6ec1b0b](https://github.com/freedomofpress/dangerzone/pull/464/commits/6ec1b0b15e6824c34a039540d4e9331773c5253c)

apyrgio: Great, thanks a lot for this change. One more thing, since we now moved this function under `dangerzone/isolation_provider/qubes.py`, I believe we can remove this logic:

```
# In Qubes we can safely access parent modules
from ..util import get_resource_path
```

dev_scripts/qa.py
apyrgio: Thanks, but I see a typo ("containers qubes").

dangerzone/conversion/common.py
apyrgio: Right, but that's not related with `dangerzone_dev`, if I understand correctly. I mean, replacing `elif not getattr(sys, "dangerzone_dev", False):` with `else:` would be effectively the same. I believe the only way this `elif` affects us is:

* if a developer runs Dangerzone in Qubes,
* **and** `QUBES_CONVERSION` is not defined or false,
* **then** it will always use the container isolation provider, without checking if `containers.tar.gz` exists.

Could it be that the above logic is necessary in your environment? If not, I suggest removing it, because it's indeed convoluted.

dev_scripts/qa.py
deeplow: should be resolved in https://github.com/freedomofpress/dangerzone/pull/464/commits/c3e439d4c1f9ff91dbb449cc1d456bb8b5087574

dangerzone/conversion/common.py
deeplow: (facepalm) you're right. I've simplified this logic now. Please check if it's correct now.

dangerzone/conversion/common.py
deeplow: Done in [6ec1b0b](https://github.com/freedomofpress/dangerzone/pull/464/commits/6ec1b0b15e6824c34a039540d4e9331773c5253c) and [e6020da](https://github.com/freedomofpress/dangerzone/pull/464/commits/e6020da6adc507ec9d25c1cb15d98c0b10c95f23).

PR #467, opened 2023-07-05
Bump python version on Windows to 3.11

deeplow:
Python 3.10.12 fixes some CVEs for which Dangerzone does not appear to be affected, however its binaries are not made available by the python foundation. Moving to 3.11 should be trivial since this was already deployed in Fedora 37+.


apyrgio (approved):
Looks good, feel free to merge.

Issue #373, opened 2023-03-22
Replace `pipes` dependency (deprecation due in python 3.13)


deeplow:
```python
dangerzone/isolation_provider/container.py:6
  /home/user/dangerzone/dangerzone/isolation_provider/container.py:6: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
    import pipes
```

We use it only for logging escaping purposes.

deeplow:
We can just replace `pipes.quote()` with `shlex.quote()` since python [literally uses that under the hood](https://github.com/python/cpython/blob/3.11/Lib/pipes.py#L66).

apyrgio:
Sounds good :+1:

PR #481, opened 2023-07-23
Remove pipes module and use shlex instead

OctopusET:
pipes module will be removed in python 3.13

I didn't add compatible code for 3.3 or lower, since all supported system has python 3.3+ version.

There may be problems with non-Unix shells such as Windows. (https://docs.python.org/3/library/shlex.html#shlex.quote)

Thanks: https://github.com/tox-dev/tox/pull/2418/files

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -3,8 +3,8 @@ import json
 import logging
 import os
 import pathlib
-import pipes
 import platform
+import shlex
 import shutil
 import subprocess
 import tempfile
@@ -152,7 +152,7 @@ class Container(IsolationProvider):
         document: Document,
         args: List[str],
     ) -> int:
-        args_str = " ".join(pipes.quote(s) for s in args)
+        args_str = " ".join(shlex.quote(s) for s in args)
         log.info("> " + args_str)
 
         with subprocess.Popen(
--- a/dangerzone/gui/logic.py
+++ b/dangerzone/gui/logic.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import pipes
 import platform
 import shlex
 import subprocess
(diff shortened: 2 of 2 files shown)


dangerzone/isolation_provider/container.py
```diff
@@ -3,8 +3,8 @@
 import logging
 import os
 import pathlib
-import pipes
 import platform
+import shlex
```
OctopusET, line 7: `from shlex import quote`
alternative way

deeplow:
Quite nice! Thanks! This fixes https://github.com/freedomofpress/dangerzone/issues/373

OctopusET:
Oh I should've searched the issue before. Thanks!

dangerzone/isolation_provider/container.py
apyrgio: I like keeping the namespace, so `shlex.quote` is fine :-)

apyrgio:
I'll add the following extra patch, since `pipes.quote()` appears in `dangerzone/gui/logic.py` as well:

```patch
diff --git a/dangerzone/gui/logic.py b/dangerzone/gui/logic.py
index 654a2ac..8370430 100644
--- a/dangerzone/gui/logic.py
+++ b/dangerzone/gui/logic.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import pipes
 import platform
 import shlex
 import subprocess
@@ -68,7 +67,7 @@ class DangerzoneGui(DangerzoneCore):
             args = ["open", "-a", "Preview.app", filename]
 
             # Run
-            args_str = " ".join(pipes.quote(s) for s in args)
+            args_str = " ".join(shlex.quote(s) for s in args)
             log.info(Fore.YELLOW + "> " + Fore.CYAN + args_str)
             subprocess.run(args)
 
@@ -89,7 +88,7 @@ class DangerzoneGui(DangerzoneCore):
                     args[i] = filename
 
             # Open as a background process
-            args_str = " ".join(pipes.quote(s) for s in args)
+            args_str = " ".join(shlex.quote(s) for s in args)
             log.info(Fore.YELLOW + "> " + Fore.CYAN + args_str)
             subprocess.Popen(args)
```

apyrgio (approved):

Issue #482, opened 2023-07-24
Bump Python dependencies


apyrgio:
Prior to releasing 0.4.2, we need to bump our Python dependencies with `poetry lock`.

apyrgio:
Our solution to this issue is to migrate our CI runners and development environments to Qt6. This means that we will replace:

* `libqt5gui5` with `libqt6gui6` in Debian-based platforms, and
* `qt5-qtbase-gui` with `qt6-qtbase-gui` in Fedora platforms.

This brings us even more closer to full Qt6 / PySide6 support in all platforms (see https://github.com/freedomofpress/dangerzone/issues/211).

Note that there are some caveats:

- On Ubuntu Focal, there is no Qt6. We still need to use Qt5.
- On Debian Bullseye, the Qt6 library misses a dependency to `libgl1`.

apyrgio:
Note that in Debian-based systems, `libxcb-cursor0` is not installed via any Qt library:

```console
$ apt-cache rdepends libxcb-cursor0
libxcb-cursor0
Reverse Depends:
  libxcb-cursor-dev
  yambar
  rofi
  polybar
  awesome
  kwin-common
  kde-spectacle
  i3-wm
```

This means that we need to install it explicitly in our development environments.

apyrgio:
On Ubuntu Jammy, `PySide6` requires `libxkbcommon-x11.so.0`:

```
qt.core.library: "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.10/lib/python3.10/site-packages/PySide6/Qt/plugins/platforms/libqxcb.so" cannot load: Cannot load library /home/user/.cache/pypoetry/virtualenvs/dangerzone-h
QU0mwlP-py3.10/lib/python3.10/site-packages/PySide6/Qt/plugins/platforms/libqxcb.so: (libxkbcommon-x11.so.0: cannot open shared object file: No such file or directory)
qt.core.plugin.loader: QLibraryPrivate::loadPlugin failed on "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.10/lib/python3.10/site-packages/PySide6/Qt/plugins/platforms/libqxcb.so" : "Cannot load library /home/user/.cache
[...]

PR #487, opened 2023-07-26
Force text color to be black

apyrgio:
Add a rule in our `dangerzone.css` file to make the text in all of our widgets black. This is supposedly the default option, based on our experience so far. After the update to the latest PySide6 libraries (see #482), we've seen that in Fedora 37 the text color is set to grey. This gives the impression that the window is disabled, so we want to set the text color to black by default.

Before:

dz_grey

After:

dz_black

### DIFF
--- a/share/dangerzone.css
+++ b/share/dangerzone.css
@@ -5,6 +5,10 @@ QLineEdit {
     padding: 3px;
 }
 
+QWidget {
+    color: black;
+}
+
 /*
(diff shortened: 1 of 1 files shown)


deeplow:
What you show some pictures of the difference, please?

apyrgio:
Sure, just did.

deeplow (approved):

PR #490, opened 2023-07-28
Improve the UX of the update check flow

apyrgio:
Slightly update some messages and dialogs to improve the overall UX.


deeplow:
I have added the missing :red_circle: to the hamburger menu.

error

deeplow (changes requested):
Looks good. But just two final things in the updater failure case

went-wrong

1. do you think we can make this dialog use the collapsible bar. The error in specific is a bit too technical and I feel like that part below (we recommend checking manually) is a bit more important. So I'd move that part up and have the error message as a collapsible.
2. There are no paragraphs in the error message. Could something be eating the `\n`?

apyrgio:
> I have added the missing red_circle to the hamburger menu.

Yay, thanks!

> do you think we can make this dialog use the collapsible bar. 

Right, it's not easy in the eyes. I can put it under the collapsible bar, sure.

> There are no paragraphs in the error message. Could something be eating the \n?

There's no `\n` actually, it's just a single line. At some point, I wanted to also provide a traceback, and convert everything to HTML, but I ran out of time. Are you ok with taking a stab at this at a later time?

deeplow:
> There's no \n actually, it's just a single line. At some point, I wanted to also provide a traceback, and convert everything to HTML, but I ran out of time. Are you ok with taking a stab at this at a later time?

Sure. Let's not let that be a blocker and for now we can just swap order.

apyrgio:
I have updated the PR with the change for putting the error details in a collapsible box. Feel free to take a look.

deeplow (approved):
Good do go.

PR #491, opened 2023-07-31
Sanitize sandbox output

apyrgio:
Sanitize the output that the container/disposable VM sends back to the user, before displaying it on screen/terminal. Also tackle some issues that are closely related to the sanitization logic, and add some tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,11 @@ lint-apply: lint-black-apply lint-isort-apply ## apply all the linter's suggesti
 
 .PHONY: test
 test:
-	pytest -v --cov --ignore dev_scripts
+	# Make each GUI test run as a separate process, to avoid segfaults due to
+	# shared state.
+	# See more in https://github.com/freedomofpress/dangerzone/issues/493
+	pytest --co -q tests/gui | grep -v ' collected' | xargs -n 1 pytest -v
+	pytest -v --cov --ignore dev_scripts --ignore tests/gui
 
 
 # Makefile self-help borrowed from the securedrop-client project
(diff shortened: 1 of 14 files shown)


dangerzone/isolation_provider/base.py
```diff
@@ -49,21 +50,34 @@ def _convert(
...
 
         if self.progress_callback:
             self.progress_callback(error, text, percentage)
 
+    def print_progress_trusted(
+        self, document: Document, error: bool, text: str, percentage: float
+    ) -> None:
+        return self._print_progress(document, error, text, percentage)
+
+    def print_progress(
```
deeplow, line 73: If we mark the other as trusted, maybe we should name this untrusted. I feel like untrusted should be the highlighted part.

deeplow (approved):
One tiny comment, but otherwise a quite a good hardening. And thanks for catching also the GUI plaintext part.

dangerzone/isolation_provider/base.py
apyrgio: Right, I get that for symmetry, naming it this way makes sense. But I wanted to insert an asymmetry here. Treating progress as untrusted should be the default, hence why I want us to use `print_progress` instead of `print_progress_untrusted` (which is a bit negative as well).

This way, a dev doesn't have to think twice about what function to use. They will use the default one always, and only switch to `print_progress_trusted` if they have a valid reason to do so.

Makefile
```diff
@@ -37,7 +37,8 @@ lint-apply: lint-black-apply lint-isort-apply ## apply all the linter's suggesti
 
 .PHONY: test
 test:
-	pytest -v --cov --ignore dev_scripts
+	pytest --co -q tests/gui | grep -v ' collected' | xargs -n 1 pytest -v
```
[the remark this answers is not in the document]
apyrgio: Aye, adding it now.

PR #508, opened 2023-08-08
Updater dialog: make "yes" the default button

### LINKED ISSUE #507: Default update install button is wrong
We made the <kbd>Yes</kbd> button be on the right, to make it more easily clickable, however, what we didn't expect was that on macOS it would show <kbd>No</kbd> as the default (in blue). We should make the other one the default.

button_default

Kudos to @apyrgio for spotting this.

deeplow:
Fixes #507
updtr


apyrgio:
Fire away :rocket:

apyrgio (approved):

PR #509, opened 2023-08-09
Adds minimal install advice on README.md

### LINKED ISSUE #475: Improve organization of installation docs
Currently we have:

- minimal instructions for Windows/Mac in the README.md
- detailed instructions for Linux package repository configuration in INSTALL.md

There are a few issues with this:
- README.md does not make it sufficiently clear that you need to install Docker for Desktop to use Dangerzone on Mac or Windows. The app itself will warn the user about this, but we should state the prerequisites more clearly upfront. (Docker is mentioned a couple of times in the README, but not in the section with the download/install info.)
- INSTALL.md only covers Linux, which means that users who land on that guide may assume that Dangerzone is a Linux-only app.
- INSTALL.md also does not make the prerequisites for Linux clear (i.e. it does not mention Docker or Podman).

I would suggest that we take the following approach:
- Treat README.md as a "quickstart guide" that summarizes the steps needed on every supported platform
- Treat INSTALL.md as detailed installation instructions on every supported platform (and that may include reference to secondary installation options, such as Homebrew, which we might omit from the README for reasons of brevity)

deeplow:
Makes it clear that one needs to install Docker for Desktop to use Dangerzone on Mac or Windows and Podman on linux. The app itself will warn the user about this, but we should state the prerequisites more clearly upfront.

Mentions mac and windows in INSTALL.md so that anyone reading this page does not wrongly assume that Dangerzone is a Linux-only app.

Fixes #475

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,14 @@
-Dangerzone is available for:
+## MacOS
+See instructions on [README.md](README.md).
 
+## Windows
+See instructions on [README.md](README.md).
+
+## Linux
+On Linux, Dangerzone uses [Podman](https://podman.io/) instead of Docker Desktop for creating
+an isolated environment. It will be installed automatically when installing Dangerzone.
+
+Dangerzone is available for:
 - Ubuntu 23.04 (lunar)
 - Ubuntu 22.04 (jammy)
 - Ubuntu 20.04 (focal)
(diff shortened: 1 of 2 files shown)


eloquence (approved):
LGTM, I pushed minor tweaks in a follow-up commit, if those look good to you, feel free to merge :)

PR #510, opened 2023-08-09
Update windows certificate in build-app.bat

deeplow:
@apyrgio I believe you may have some change suggestions


apyrgio:
My only comment here (I think) was replacing the Digicert timestamp server:

```
/tr http://timestamp.digicert.com
```

with the Certum one, since we are issuing Certum certificates:

```
/t http://time.certum.pl/
```

Probably doesn't make a difference in the long run, but it's a bit clearer this way.

apyrgio (approved):

PR #515, opened 2023-08-14
Propagate "update check" prompt to UI checkbox

### LINKED ISSUE #513: Update check does not place "Check for updates" checkmark in UI
# Steps to reproduce

1. On a Mac, install and run Dangerzone so you get the update check prompt.
2. Confirm that you want automatic update checks.
3. Click the menu button to reveal the current state of the "update check" settings.

# Expected behavior

Update check should have a checkmark next to it, since I've enabled update checks

# Actual behavior

Checkmark is not set until I restart Dangerzone

deeplow:
The "check for updates" button wasn't showing up immediately as checked as soon as the user is prompted for checking updates. This fixes that.

Fixes #513

### DIFF
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -123,7 +123,6 @@ def gui_main(
     # Check for updates
     log.debug("Setting up Dangezone updater")
     updater = UpdaterThread(dangerzone)
-    updater.update_check_toggled.connect(window.refresh_updates_checkbox)
     window.register_update_handler(updater.finished)
 
     log.debug("Consulting updater settings before checking for updates")
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -123,6 +123,7 @@ def gui_main(
     # Check for updates
     log.debug("Setting up Dangezone updater")
     updater = UpdaterThread(dangerzone)
+    updater.update_check_toggled.connect(window.refresh_updates_checkbox)
     window.register_update_handler(updater.finished)
 
     log.debug("Consulting updater settings before checking for updates")
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -132,6 +132,10 @@ def gui_main(
     else:
         log.debug("Will not check for updates, based on updater settings")
 
+    # Ensure the status of the toggle updates checkbox is updated, after the user is
+    # prompted to enable updates.
+    window.toggle_updates_action.setChecked(bool(updater.check))
+
     if filenames:
         open_files(filenames)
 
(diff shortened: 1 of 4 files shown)


tests/gui/test_updater.py
```diff
@@ -165,8 +168,10 @@ def test_user_prompts(
     # Check disabling update checks.
     prompt_mock().launch.return_value = False  # type: ignore [attr-defined]
     expected_settings["updater_check"] = False
-    assert updater.should_check_for_updates() == False
-    assert updater.dangerzone.settings.get_updater_settings() == expected_settings
+
+    with qtbot.waitSignal(updater.update_check_toggled) as blocker:
+        assert updater.should_check_for_updates() == False
+        assert updater.dangerzone.settings.get_updater_settings() == expected_settings
```
apyrgio, line 174: I totally agree with having a test, but I think that this diff does not end up checking the status of the `toggle_updates_action` checkbox.

dangerzone/gui/__init__.py
```diff
@@ -123,6 +123,7 @@ def open_files(filenames: List[str] = []) -> None:
     # Check for updates
     log.debug("Setting up Dangezone updater")
     updater = UpdaterThread(dangerzone)
+    updater.update_check_toggled.connect(window.refresh_updates_checkbox)
```
apyrgio, line 126: Personally, I'd replace everything with these three lines:

```diff
diff --git a/dangerzone/gui/__init__.py b/dangerzone/gui/__init__.py
index f0249ac..c35d232 100644
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -132,6 +132,10 @@ def gui_main(
     else:
         log.debug("Will not check for updates, based on updater settings")
 
+    # Ensure the status of the toggle updates checkbox is updated, after the user is
+    # prompted to enable updates.
+    window.toggle_updates_action.setChecked(bool(updater.check))
+
     if filenames:
         open_files(filenames)
 
```

I'd suggest against including the signal mechanism, since the main window has not started yet, meaning that we are not constrained by the limitations of the Qt event loop.

apyrgio (changes requested):

deeplow:
@apyrgio that's way more elegant. Let's go with that. I haven't made a test, though since I didn't find a decent place to go a test for this. Since the behavior only really happens in `gui/__init__.py`. Do you have suggestions regarding testing this? Or should we leave it untested?

apyrgio:
Hm, you're right, there's no easy way to test it. Given that it's a one line thing, maybe we can just test it manually.

deeplow:
Anything else missing @apyrgio?

apyrgio:
Nope, just checked it manually as well, and it seems to work. Feel free to merge it with just the latest commit.

apyrgio (approved):

PR #523, opened 2023-08-17
Post-release fixes for MacOS issues

### LINKED ISSUE #519: Allow running the docker command without sudo
Our `dev_scripts/env.py` script assumes that Podman commands run rootless, and Docker commands run with `sudo`. This is wrong in MacOS though, where Docker symlinks its Unix Domain socket under the home directory of the user who started it, and thus does not require `sudo`.

There are two ways we can handle this: either remove the `sudo` invocation from the Docker command directly, since us devs use Podman mostly, or auto-detect if the `DOCKER_HOST` envvar is defined and accessible by the user without permissions.

### LINKED ISSUE #522: Automatically get the altool password from the MacOS keychain
During DMG notarization, we need an application password for `altool`. This password can be retrieved if one opens the Keychain GUI and copy pastes it to the prompt. To avoid this interaction though, we can make `altool` grab it directly from the keychain with `-p "keychain:altool`, provided we have saved it under that name.

This approach has the benefit that the user will be prompted to enter their login password once, or not at all, depending on when they logged in.

apyrgio:
This PR includes some small fixes for issues that bit us while releasing Dangerzone v0.4.2 on MacOS platforms.

Closes #519 
Closes #522

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,6 +10,9 @@ and newer platforms, we have to do the following:
 - [ ] In `.circleci/config.yml`, add new platforms and remove obsolete platforms
 - [ ] Bump the Python dependencies using `poetry lock`
 - [ ] Make sure that the tip of the `main` branch passes the CI tests.
+- [ ] Make sure that the Apple account has a valid application password and has
+      agreed to the latest Apple terms (see [macOS release](#macos-release)
+      section).
 - [ ] Create a test build in Windows and make sure it works:
   - [ ] Check if the suggested Python version is still supported.
   - [ ] Create a new development environment with Poetry.
@@ -199,6 +202,11 @@ To make a macOS release, go to macOS build machine:
 
 - Build machine must have:
   - Apple-trusted `Developer ID Application: Freedom of the Press Foundation (94ZZGGGJ3W)` code-signing certificates installed
+- Apple account must have:
+  - A valid application password for `altool` in the Keychain. You can verify
+    this by running: `xcrun altool --notarization-history 0 -u "<email>" -p "@keychain:altool"`
+  - Agreed to any new terms and conditions. You can find those if you visit
+    https://developer.apple.com and login with the proper Apple ID.
 - Verify and checkout the git tag for this release
 - Run `poetry install`
 - Run `poetry run ./install/macos/build-app.py`; this will make `dist/Dangerzone.app`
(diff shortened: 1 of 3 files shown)


RELEASE.md
```diff
@@ -199,21 +199,26 @@ To make a macOS release, go to macOS build machine:
 
 - Build machine must have:
   - Apple-trusted `Developer ID Application: Freedom of the Press Foundation (94ZZGGGJ3W)` code-signing certificates installed
+- Apple account must have:
+  - A valid application password for `altool` in the Keychain. You can verify
+    this by running: `xcrun altool --notarization-history 0 -u "<email>" -p "@keychain:altool"`
+  - Agreed to any new terms and conditions. You can find those if you visit
+    https://developer.apple.com and login with the proper Apple ID.
```
deeplow, line 206: Should we move this up a bit for the "Pre-release section". It feels like stuff we can verify before we start any actual release procedures.

dev_scripts/env.py
```diff
@@ -262,10 +262,14 @@ def __init__(self, distro, version, runtime):
...
         # Try to autodetect the runtime, if the user has not provided it.
         #
-        # FIXME: Typically Podman runs without sudo, whereas Docker with sudo, but this
-        # is not always the case. We may need to autodetect that as well.
+        # XXX: Typically Podman runs without sudo, whereas Docker runs with sudo only in
+        # Linux environments. Given that we don't tend to use Docker in Linux
+        # environments, we can just ditch the sudo requirement, which breaks
+        # non-interactivity.
+        #
+        # See also: https://github.com/freedomofpress/dangerzone/issues/519
```
deeplow, line 270: This feels a bit verbose. I'd be happy as well with just dropping all the comments in this case. But if you feel it's necessary, let's leave it as is.

deeplow (commented):
Nice trick: @toolchain :) (didn't mean to poke the "toolchain" folks)

RELEASE.md
apyrgio: I agree, with an asterisk. Since the checklist is a bit verbose, I propose adding a brief item in the QA section, that points to the macOS release section. Check out 6d4e5cab0172f75f284a9db472b323464c694844.

dev_scripts/env.py
apyrgio: Removed it in 6d4e5cab0172f75f284a9db472b323464c694844.

deeplow (approved):

PR #544, opened 2023-09-07
qubes: Pass OCR parameters properly

### LINKED ISSUE #455: Qubes: Pass OCR parameters properly
The Qubes isolation provider currently passes OCR parameters to the `PixelsToPDF()` code as environment variables, and then clears them after the conversion finishes.

This is a bad practice, since the envvars affect the whole program, and may bite us if we have OCR enabled for some documents only.

apyrgio:
Pass OCR parameters to conversion functions as arguments, instead of setting environment variables.

Fixes #455

### DIFF
--- a/dangerzone/conversion/pixels_to_pdf.py
+++ b/dangerzone/conversion/pixels_to_pdf.py
@@ -11,12 +11,13 @@ import json
 import os
 import shutil
 import sys
+from typing import Optional
 
 from .common import DangerzoneConverter, running_on_qubes
 
 
 class PixelsToPDF(DangerzoneConverter):
-    async def convert(self) -> None:
+    async def convert(self, ocr_lang: Optional[str] = None) -> None:
         self.percentage = 50.0
 
         num_pages = len(glob.glob("/tmp/dangerzone/page-*.rgb"))
@@ -43,7 +44,7 @@ class PixelsToPDF(DangerzoneConverter):
             total_size += page_size
             timeout = self.calculate_timeout(page_size, 1)
 
-            if os.environ.get("OCR") == "1":  # OCR the document
+            if ocr_lang:  # OCR the document
                 self.update_progress(
                     f"Converting page {page}/{num_pages} from pixels to searchable PDF"
                 )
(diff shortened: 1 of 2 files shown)


dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -11,12 +11,16 @@
...
 import sys
+from typing import Optional
 
 from .common import DangerzoneConverter, running_on_qubes
 
 
 class PixelsToPDF(DangerzoneConverter):
-    async def convert(self) -> None:
+    async def convert(self, ocr_lang: Optional[str] = None) -> None:
+        ocr_enable = ocr_lang is not None or os.environ.get("OCR") == "1"
```
deeplow, line 21: I would only add a comment here because for future people taking a look at this code it may be weird why to have a parameter but also obtain it from an env var
```suggestion
        # XXX Qubes converter uses the parameter, while the containers version passes via env vars
        ocr_enable = ocr_lang is not None or os.environ.get("OCR") == "1"
```

deeplow: Actually, I think we could have a better approach, if you agree. We could move this env var logic onto `main()` [down in the file](https://github.com/freedomofpress/dangerzone/pull/544/files#diff-bdfbfeb3ce16c19ed296150fa806b7ac3165fa32aa249a3225ec54d70594888eR163) (that code is only executed in containers).

Then this would always receive some `ocr_lang` (or None). Furthermore, we can drop the `ocr_enable` variable and just keep the `ocr_lang`. If not `None` then do OCR with that lang, if not we drop it.

deeplow (changes requested):
Added some comments about an alternative way of adding this (in my mind a bit clearer). Let me know what you think.

dangerzone/conversion/pixels_to_pdf.py
apyrgio: Damn, you're totally right! That's indeed a great suggestion. I've fixed it in a6f65de.

deeplow (approved):

PR #546, opened 2023-09-12
Handle errors in Qubes

### LINKED ISSUE #430: Qubes: Error handling

deeplow:
Deals with error-handling in Qubes.

Fixes #430. But depends on #537


tests/isolation_provider/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+        ["pdfunite", *([pdf_file_1000_pages.name] * 70), pdf_file_70000_pages.name]
+    )
+
+    yield pdf_file_70000_pages.name
+
+    # Delete files
+    svg_file.close()
+    pdf_file_one_page.close()
+    pdf_file_1000_pages.close()
+    pdf_file_70000_pages.close()
```
apyrgio, line 49: This block has two issues:

1. It requires having some specific commands in the user's system. This may not be the case in Windows/MacOS, and even on Linux systems.
2. It takes a bit of a time, and produces a big document.

Instead, let's precompute this document. We can create a document with 70K pages with:

```
gs -sDEVICE=pdfwrite -o sample-70k-pages.pdf -dDEVICEWIDTHPOINTS=612 -dDEVICEHEIGHTPOINTS=792 -c  70000 {showpage} repeat
```

This takes roughly 21MiB of space, and we can compress it down to 1.8MiB.

We can also use the same technique (minus the compression) to create a document with large width/height, since we want to catch errors for that as well.

apyrgio (commented):
What I see looks good for now. Since it's still WIP, it will be worth having in its description which errors from #430 are not handled yet.

tests/isolation_provider/__init__.py
deeplow: That's exactly what I was looking for. Thank you.

deeplow: closing via 5cc9f17

deeplow:
It's now ready for review @apyrgio.

I've tackled most of of the error handling situations. The ones remaining are:
  -  **document wasn't immediately received** - maybe we can borrow `nonblocking_read()` in the qubes wrapper
  - **The disposable qube could not start because dz-dvm does not exist.** - I don't know how to disambiguate this from the lack of RAM situation because in both cases the exit code of `qrexec-client-vm` is `126`. I'm afraid we may need the [Admin API](https://www.qubes-os.org/doc/admin-api/) to be able to tell them apart.
  - **pre-flight check:** netVM is None - we discussed that this would be implemented later.

And one things need UX polishing:
  - **lack of RAM exception:** in this case, if the first document fails because of a lack of RAM, all other conversions should be aborted and the user should be advised to free up RAM and try again. For the moment I just left it to try (and fail) every single document because it's easier logic wise. But we can iterate on this in a future PR.

dangerzone/isolation_provider/qubes.py
```diff
@@ -84,50 +84,37 @@ def _convert(
...
-                p = subprocess.Popen(
-                    ["/usr/bin/qrexec-client-vm", "@dispvm:dz-dvm", "dz.Convert"],
-                    stdin=f,
-                    stdout=subprocess.PIPE,
-                    stderr=subprocess.DEVNULL,
-                )
+            except BrokenPipeError as e:
+                # Likely that qube did not have enough RAM to start
+                # in which case, it exits with code 126
+                if running_on_qubes() and p.wait() == 126:
```
apyrgio, line 95: Nitpick, but if we are here, should we even check `running_on_qubes()`? Also, if the process has closed its stdin, but has not finished, then execution will hang in this line. I suggest setting a timeout of 3 seconds here.

tests/isolation_provider/test_qubes.py
```diff
@@ -0,0 +1,117 @@
...
+            p = subprocess.Popen(
+                # XXX error 126 simulates a qrexec-policy failure. Source:
+                # https://github.com/QubesOS/qubes-core-qrexec/blob/fdcbfd7/daemon/qrexec-daemon.c#L1022
+                ["exit 126"],
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                shell=True,
+            )
+            time.sleep(0.2)  # Avoid race condition
```
apyrgio, line 110: What is the race condition that we are avoiding with this sleep argument?

tests/isolation_provider/test_qubes.py
```diff
@@ -0,0 +1,117 @@
...
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            assert p.stdin is not None
+            p.stdin.write(b"\x00\x01")  # 1 page
+            p.stdin.write(b"\x00\x04")  # 4px width
+            p.stdin.write(b"\x00\x05")  # 5px height
+            p.stdin.write(b"\x00")  # only one byte received but 3*4*5 expected
+            return p
```
apyrgio, line 84: This test now fails in my machine, due to the PRs that have just been merged. The reason is that `tee` does not exit, so the client-side will not receive EOF. It will fail with a timeout error instead, not `errors.InterruptedConversion`.

tests/isolation_provider/test_qubes.py
```diff
@@ -0,0 +1,117 @@
...
+        provider.progress_callback = mocker.MagicMock()
+        mocker.patch(
+            "dangerzone.conversion.errors.MAX_PAGES", 1
+        )  # sample_doc has 4 pages > 1
+        doc = Document(sample_doc)
+        with pytest.raises(errors.MaxPagesException):
+            success = provider._convert(doc, ocr_lang=None)
+            assert not success
+
+    def test_max_dimentions(
```
apyrgio, line 48:
```suggestion
    def test_max_dimensions(
```

tests/isolation_provider/test_qubes.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        provider.progress_callback = mocker.MagicMock()
+        with pytest.raises(errors.MaxPageWidthException):
+            success = provider._convert(Document(sample_bad_width), ocr_lang=None)
+            assert not success
+        with pytest.raises(errors.MaxPageHeightException):
+            success = provider._convert(Document(sample_bad_height), ocr_lang=None)
+            assert not success
+
+    def test_inconsistent_dimentions(
```
apyrgio, line 63:
```suggestion
    def test_inconsistent_dimensions(
```

apyrgio (changes requested):
I made a first pass of the PR. General comment, now that we have merged some other PRs, it seems that some Qubes tests fail:

```
FAILED tests/isolation_provider/test_qubes.py::TestIsolationProvider::test_max_pages_received - dangerzone.conversion.errors.InterruptedConversion: Something interrupted the conversion and it could not be completed.
FAILED tests/isolation_provider/test_qubes.py::TestQubes::test_max_pages_received - dangerzone.conversion.errors.InterruptedConversion: Something interrupted the conversion and it could not be completed.
FAILED tests/isolation_provider/test_qubes.py::TestQubes::test_max_dimentions - dangerzone.errors.InputFileNotFoundException: Input file not found: make sure you typed it correctly.
FAILED tests/isolation_provider/test_qubes.py::TestQubes::test_inconsistent_dimentions - Failed: DID NOT RAISE <class 'dangerzone.conversion.errors.InterruptedConversion'>
```

Can you give them a look?

dangerzone/isolation_provider/qubes.py
deeplow: Duh. Yeah, obviously we don't need to. Somehow I muse have imagined this was shared code.

> I suggest setting a timeout of 3 seconds here.

You're right. I also wanted that. I recall me wanting to reuse you timeout code but not doing it because something was still missing or we need to adapt something to make use of that here. I'll recheck.

dangerzone/isolation_provider/qubes.py
```diff
@@ -84,50 +84,37 @@ def _convert(
...
             except ValueError:
                 error_code = p.wait()
                 # XXX Reconstruct exception from error code
-                raise exception_from_error_code(error_code)  # type: ignore [misc]
-
-            if n_pages == 0:
-                # FIXME: Fail loudly in that case
-                return False
+                raise errors.exception_from_error_code(error_code)  # type: ignore [misc]
+            if 1 < n_pages > errors.MAX_PAGES:
```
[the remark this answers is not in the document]
deeplow: Sorry about that. I was disliking this double inequality. Fixed in 8dea8d55.

tests/__init__.py
```diff
@@ -73,6 +88,20 @@ def unreadable_pdf(tmp_path: Path) -> str:
     return str(file_path)
 
 
+@pytest.fixture
+def pdf_11k_pages(tmp_path: Path) -> str:
+    """11K page document with pages of 1x1 px. Generated with the command:
+
+    gs -sDEVICE=pdfwrite -o sample-11k-pages.pdf -dDEVICEWIDTHPOINTS=1 -dDEVICEHEIGHTPOINTS=1 -c 1000 {showpage} repeat
```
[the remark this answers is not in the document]
deeplow: Fixed in 4365e591.

tests/isolation_provider/test_qubes.py
deeplow: Oh my English! 3dbacdb3

tests/isolation_provider/test_qubes.py
deeplow: Fixed in 32f7cda.

tests/isolation_provider/test_qubes.py
deeplow: It was an old comment. Thanks for catching it! Fixed in be0fbf21.

deeplow:
`tests/isolation_provider/test_qubes.py::TestIsolationProvider::test_max_pages_received` is failing for me. I think it will be fixed by #551. In short it's failing server-side because it hit the 10k page limit and then it's receiving no bytes for what was supposed to be the `n_pages`. But the reality is that it should have detected that an error occurred.

tests/isolation_provider/test_qubes.py
apyrgio: Hm, I think that `tee & exit` now makes the test racy. What if `tee` exits before it sends back the input? I think a better way is to change `tee & exit` with `echo -e`:

```diff
diff --git a/tests/isolation_provider/test_qubes.py b/tests/isolation_provider/test_qubes.py
index a1e5efb..f3ab918 100644
--- a/tests/isolation_provider/test_qubes.py
+++ b/tests/isolation_provider/test_qubes.py
@@ -71,17 +71,14 @@ class TestQubes(TestIsolationProvider):
 
         def qrexec_subprocess() -> subprocess.Popen:
             p = subprocess.Popen(
-                ["tee & exit"],  # pipe out whatever we feed into stdin
+                ["echo",
+                 "-n",
+                 "-e",
+                 r"\x00\x01\x00\x04\x00\x05\x99"],
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
-                shell=True,
             )
-            assert p.stdin is not None
-            p.stdin.write(b"\x00\x01")  # 1 page
-            p.stdin.write(b"\x00\x04")  # 4px width
-            p.stdin.write(b"\x00\x05")  # 5px height
-            p.stdin.write(b"\x00")  # only one byte received but 3*4*5 expected
```

apyrgio:
> `tests/isolation_provider/test_qubes.py::TestIsolationProvider::test_max_pages_received` is failing for me. I think it will be fixed by #551. In short it's failing server-side because it hit the 10k page limit and then it's receiving no bytes for what was supposed to be the `n_pages`. But the reality is that it should have detected that an error occurred.

I have rebased this test on top of my WIP PR, and it does seem to work. I don't understand why though, since it should fail immediately, once it reads the number of pages.

> `FAILED tests/isolation_provider/test_qubes.py::TestQubes::test_max_dimentions - dangerzone.errors.InputFileNotFoundException: Input file not found: make sure you typed it correctly.`

Note that this fails because most probably, a file has not been added to Git.

tests/isolation_provider/test_qubes.py
deeplow: That also works. Thanks for the suggestions. Implemented in 0f189c1d.

deeplow:
>> st_max_pages_received is failing for me. I think it will be fixed by https://github.com/freedomofpress/dangerzone/pull/551. In short it's failing server-side because it hit the 10k page limit and then it's receiving no bytes for what was supposed to be the n_pages. But the reality is that it should have detected that an error occurred.
>
> I have rebased this test on top of my WIP PR, and it does seem to work. I don't understand why though, since it should fail immediately, once it reads the number of pages.

I don't follow. You're saying the test should fail in that case, yet it's passing? The test should pass _because it failed_ when it received the pages.

deeplow:
>>    FAILED tests/isolation_provider/test_qubes.py::TestQubes::test_max_dimentions - dangerzone.errors.InputFileNotFoundException: Input file not found: make sure you typed it correctly.
> 
> Note that this fails because most probably, a file has not been added to Git.

Whoops. I just added them in 757f0881.

deeplow:
> EDIT: One more issue with the tests that we must deal with is that some of them run twice, e.g., test_max_pages_received.

This should be fixed with 7c9e552. Turns out pytest also recognizes test classes that start with `Test` in their name (even though the file they're in, doesn't start with `test_`. So I just moved the `Test` in the name to the end.

apyrgio:
> I don't follow. You're saying the test should fail in that case, yet it's passing? The test should pass because it failed when it received the pages (server-side).

I was saying that the test should have passed as is, without my PR, because the server-side check for the max number of pages would get immediately triggered, raise an exception, and our `try ... except:` clause in the client would catch it. In any case, I'll do one more read of the code once all PRs are merged, because it's a bit confuring right now. Let's ignore this comment for now.

dangerzone/isolation_provider/qubes.py
```diff
@@ -84,50 +84,37 @@ def _convert(
...
+            except BrokenPipeError as e:
+                # Likely that qube did not have enough RAM to start
+                # in which case, it exits with code 126
+                if p.wait() == 126:
+                    # TODO: disambiguate from qrexec policy misconfigured, if possible
+                    raise errors.QubesNotEnoughRAMError()
+                else:
+                    # XXX preserve the exception since it can also have other causes
+                    # e.g.: Ctrl+C with the right timing can trigger BrokenPipeError
+                    raise e
```
apyrgio, line 101: Now that I have reworked the #551  PR, I think I have a much better solution to this:

 ```suggestion
            except BrokenPipeError as e:
                    raise errors.InterruptedConversion()
```

and change the error code of `QubesNotEnoughRAMError` to 126. This way, this error will be detected at a central point (the `try ... except` clause that I introduce in this PR), and we will avoid having custom error handling logic scattered in the code.

_(minor note, when re-raising exception, it's best to use `raise` instead of `raise e`, since the former preserves the original traceback)_

deeplow: Yes, that works. I had also considered that 126 "hack" but ultimately didn't submit it because I it felt like we were mixing two distinct things. But now seeing this again and that you had a similar intuition, it might make sense.

apyrgio: Yeah, I feel this asymmetry as well. We _could_ say that this error is actually returned by `qrexec`, and not by the disposable qube, just to explain why there is this different there.

deeplow:
I have rebased it on top of main now that #551 was merged.

A few notes:
  - I implemented the OutOfRAM detection as suggested (worked perfectly)
  - **I kept the race condition avoidance timeout** - even though the test was very stable (passed 5 times in a row), because we move the detection code to `_convert`. So it has lots of time to exit before the error is caught. But I decided to keep a `0.1` sleep just in case.
  - there was a `raise errors.InterruptedConversion` (without the `()`) which I fixed (that passed review [here](https://github.com/freedomofpress/dangerzone/pull/551#discussion_r1337384759))


Also, how do you feel about the inconsistent dimentions test? I have removed it temporarily, but to make it work, it may be a bit tricky. We need to let it send data partially but not exit. (maybe a sleep?). And then in any case it will fail due to timeout and not due to missing data.

apyrgio:
> I implemented the OutOfRAM detection as suggested (worked perfectly)

:tada:

> I kept the race condition avoidance timeout - even though the test was very stable (passed 5 times in a row), because we move the detection code to _convert. So it has lots of time to exit before the error is caught. But I decided to keep a 0.1 sleep just in case.

I believe I understand what was going on with that test. You needed to sleep there so that `qrexec` definitely exits, and thus writing the document to the `qrexec`'s stdin fails. This was necessary because it was the only point where we looked for the 126 error code. If the `qrexec` streams closed afterwards, e.g., while waiting for the number of pages, we wouldn't detect that we are out of RAM.

By having a sleep there, we actually masked the fact that our "out of RAM" error handling did not work in all cases! Now that we always check for the 126 error code, I propose removing this sleep call.

> there was a raise errors.InterruptedConversion (without the ()) which I fixed (that passed review https://github.com/freedomofpress/dangerzone/pull/551#discussion_r1337384759)

Thanks. Note that's actually valid syntax: https://stackoverflow.com/questions/19768515/is-there-a-difference-between-raising-exception-class-and-exception-instance. I understand though that it may break symmetry, so I don't mind your change.

apyrgio (approved):

Issue #443, opened 2023-06-13
Containers: have progress streamed instead of via mounted volumes (and deprecate doc_to_pixels_qubes_wrapper.py)


deeplow:
The current approach to sending data back from the "doc-to-pixels" container is via [mounted volumes](https://github.com/freedomofpress/dangerzone/blob/main/dangerzone/isolation_provider/container.py#L272). This was practical because we had essentially two streams to send. However, the Qubes implementation is simpler and has less attacker-controlled data passing, since error messages are not sent. Only the really needed data.

> Personally, I want to move all the Qubes wrapper code within doc_to_pixels.py, since we plan to use the stdio methods in containers as well. Not something that important for now, but something to keep in mind. ([source](https://github.com/freedomofpress/dangerzone/pull/437#discussion_r1221661460))

## Tasks
  - [ ] change container logic to stream pages as they are converted
  - [ ] implement client-side timeouts (see also #446)
  - [ ] deprecate json progress information add way to receive debug information when conversion fails (probably after the conversion fails)
  - [ ] deprecate the [`doc_to_pixels_qubes_wrapper.py`](https://github.com/freedomofpress/dangerzone/pull/437/files/36de9e74cf0cc8db2eb9f1ee97655b3d984c2326#diff-132272f6adcaf5e1148cdd5a76ce186ebd62a9224532038f49d458a5a9e9d402) since now both containers and Qubes will be streaming pages.

deeplow:
Looks like our files-based approach lead to one more issue: https://github.com/freedomofpress/dangerzone/issues/620

PR #551, opened 2023-09-25
Detect if we received EOF due to a command that failed

apyrgio:
Handle incomplete reads due to EOF by checking if the underlying command has exited. If so, raise the corresponding exception.

Refs #430

### DIFF
--- a/dangerzone/conversion/errors.py
+++ b/dangerzone/conversion/errors.py
@@ -68,6 +68,15 @@ class PDFtoPPMInvalidDepth(PDFtoPPMException):
     error_message = "Error converting PDF to Pixels (Invalid PPM depth)"
 
 
+class InterruptedConversion(ConversionException):
+    """Protocol received num of bytes different than expected"""
+
+    error_code = ERROR_SHIFT + 60
+    error_message = (
+        "Something interrupted the conversion and it could not be completed."
+    )
+
+
 class UnexpectedConversionError(PDFtoPPMException):
     error_code = ERROR_SHIFT + 100
     error_message = "Some unexpected error occurred while converting the document"
--- a/dangerzone/isolation_provider/qubes.py
+++ b/dangerzone/isolation_provider/qubes.py
@@ -68,7 +68,7 @@ class Qubes(IsolationProvider):
     def install(self) -> bool:
         return True
 
-    def _convert(
+    def __convert(
         self,
         document: Document,
         ocr_lang: Optional[str] = None,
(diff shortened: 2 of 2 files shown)


deeplow (commented):
Only added a note on the fact that we have now two args and one (`p.stdout`) can be accessed from the other (`p`).

Other than that it looks good. The only extra thing is that this essentially removes the need for the `InterruptedConversion` exception from #546 as this will catch the exit code before it detects that it received less data than it expected. But I think that's OK.

deeplow:
This also has issues with the `test_out_of_ram` but I already have a fix for that which we can add to #430 once it gets rebased on top of this:

```diff
diff --cc tests/isolation_provider/test_qubes.py
index f3ab9185,70d57463..00000000
--- a/tests/isolation_provider/test_qubes.py
+++ b/tests/isolation_provider/test_qubes.py
diff --git a/dangerzone/isolation_provider/qubes.py b/dangerzone/isolation_provider/qubes.py
index 80c3e067..dfef8a96 100644
--- a/dangerzone/isolation_provider/qubes.py
+++ b/dangerzone/isolation_provider/qubes.py
@@ -51,6 +51,8 @@ def read_bytes(
             # We may have received an EOF while the process was exiting. Find the real
             # reason why.
             error_code = p.wait(3)
+            if error_code == 126:
+                raise errors.QubesNotEnoughRAMError()
             raise errors.exception_from_error_code(error_code)  # type: ignore [misc]
         except subprocess.TimeoutExpired:
             # The process is still running, but it suddenly closed its fds. Raise an
@@ -99,20 +101,9 @@ class Qubes(IsolationProvider):
 
         with open(document.input_filename, "rb") as f:
             p = self.qrexec_subprocess()
-            try:
-                assert p.stdin is not None
-                p.stdin.write(f.read())
-                p.stdin.close()
-            except BrokenPipeError as e:
-                # Likely that qube did not have enough RAM to start
-                # in which case, it exits with code 126
-                if p.wait() == 126:
-                    # TODO: disambiguate from qrexec policy misconfigured, if possible
-                    raise errors.QubesNotEnoughRAMError()
-                else:
-                    # XXX preserve the exception since it can also have other causes
-                    # e.g.: Ctrl+C with the right timing can trigger BrokenPipeError
-                    raise e
+            assert p.stdin is not None
+            p.stdin.write(f.read())
+            p.stdin.close()
 
             assert p.stdout is not None
             os.set_blocking(p.stdout.fileno(), False)
```

apyrgio:
I have reworked the PR to catch errors caused by EOF in a separate place. I think this should address your concerns about our helpers and the `InterruptedConversion` exception. Tell me if you're ok with it.

deeplow (approved):
Woah, we're now two levels deep into the `__convert`. I guess we're just going to have to live with it until we handle #443. That issue being delayed in particular is creating a lot of technical.

So I'll just approve this but when we tackle #443 we'll have lots of stuff to address.

PR #554, opened 2023-09-26
Round conversion progress from float to int

### LINKED ISSUE #553: Conversion progress in qubes shoud be rounded up.
round

deeplow:
Fixes #553

| Before | After |
|---|----|
| round | afterr |

This affects both the containers version and the Qubes one.


dangerzone/isolation_provider/base.py
```diff
@@ -78,7 +78,7 @@ def _print_progress(
     def print_progress_trusted(
         self, document: Document, error: bool, text: str, percentage: float
     ) -> None:
-        return self._print_progress(document, error, text, percentage)
+        return self._print_progress(document, error, text, int(percentage))
```
apyrgio, line 81: Given that:

```python
>>> int(99.9)
99
>>> round(99.9)
100
```

I'd suggest going with `round()`. Other than that, feel free to merge.

apyrgio (approved):

dangerzone/isolation_provider/base.py
deeplow: Hum. I feel like if it's 99.9, it's best to just truncate it to 99% rather than false hope of showing the user 100% and it not really being there. Thoughts?

PR #558, opened 2023-09-26
Update notarization process (altool deprecated)

### LINKED ISSUE #506: Migrate from altool to notarytool for app notarization on MacOS
Apple has decided to stop support for `altool` in 2023-11-01. Instead, they urge users to switch to `notarytool`. Here's how other users have dealt with this: https://successfulsoftware.net/2023/04/28/moving-from-altool-to-notarytool-for-mac-notarization/

deeplow:
Following the deprecation notice of the Apple notarization tool 'altool', we're updating the instructions to reflect the change to the new tool 'notarytool'.

The migration process essentially required updating the commands and migrating credentials. It is documented in [1].

Fixes #506

[1]: https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -197,8 +197,10 @@ To make a macOS release, go to macOS build machine:
 - Build machine must have:
   - Apple-trusted `Developer ID Application: Freedom of the Press Foundation (94ZZGGGJ3W)` code-signing certificates installed
 - Apple account must have:
-  - A valid application password for `altool` in the Keychain. You can verify
-    this by running: `xcrun altool --notarization-history 0 -u "<email>" -p "@keychain:altool"`
+  - A valid application password for `notarytool` in the Keychain. You can verify
+    this by running: `xcrun notarytool history --apple-id "<email>" --keychain-profile "dz-notarytool-release-key"`. If you don't find it, you can add it to the Keychain by running
+    `xcrun notarytool store-credentials dz-notarytool-release-key --apple-id <email> --team-id <team ID>`
+    with the respective `email` and `team ID` (the latter can be obtained [here](https://developer.apple.com/help/account/manage-your-team/locate-your-team-id))
   - Agreed to any new terms and conditions. You can find those if you visit
     https://developer.apple.com and login with the proper Apple ID.
 - Verify and checkout the git tag for this release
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -207,15 +207,16 @@ To make a macOS release, go to macOS build machine:
...
+  * In the end you'll get a `REQUEST_UUID`, which identifies the submission. Keep it to check on its status.
   * You need to change the `<email>` in the above command with the email
     associated with the Apple Developer ID.
   * This command assumes that you have created, and stored in the Keychain, an
     application password associated with your Apple Developer ID, which will be
-    used specifically for `altool`.
-- Wait for it to get approved, check status with: `xcrun altool --notarization-history 0 -u "<email>" -p "@keychain:altool"`
+    used specifically for `notarytool`.
+- Wait for it to get approved, check status with: `xcrun notarytool info <REQUEST_UUID> --apple-id "<email>" --keychain-profile "dz-notarytool-release-key"`
+  * If it gets rejected, you should be able to see why with the same command
```
apyrgio, line 218: I think the user can see full info with the `log` command, instead of the `info` one. What's your experience there?

RELEASE.md
```diff
@@ -197,8 +197,8 @@ To make a macOS release, go to macOS build machine:
 - Build machine must have:
   - Apple-trusted `Developer ID Application: Freedom of the Press Foundation (94ZZGGGJ3W)` code-signing certificates installed
 - Apple account must have:
-  - A valid application password for `altool` in the Keychain. You can verify
-    this by running: `xcrun altool --notarization-history 0 -u "<email>" -p "@keychain:altool"`
+  - A valid application password for `notarytool` in the Keychain. You can verify
```
apyrgio, line 200: Would it make sense to say that, if the application password is not stored there, that the user can store it with `xcrun notarytool store-credentials dz-notarytool-release-key --apple-id <email> --team-id <team ID>`?

apyrgio (commented):
I have some very small comments, otherwise LGTM.

RELEASE.md
deeplow: `log` is terribly verbose. So I think for a summary the best one is `info`. But I added a note in 9c19776 in case someone wants to see the full log.

RELEASE.md
deeplow: Sure. I just added something like that in 20a05d39.

apyrgio: Nice, thanks.

apyrgio (approved):

PR #561, opened 2023-09-26
Qubes: Stream page data in real time

### LINKED ISSUE #557: Qubes: Report number of pages as early as possible
Report the number of pages as early as possible, instead of doing so at the end of the conversion, else the client-side timeouts logic (#446) may exit early at the first `read_int()` call. This touches a bit upon #443.

### LINKED ISSUE #560: Dangerzone may exclude the last page of a document
# Background

Dangerzone uses a tool called `pdftoppm` in order to convert a PDF document into pixels:

https://github.com/freedomofpress/dangerzone/blob/18b73d94b0b0eb5a1dc7d990ca64885f4963cdad/dangerzone/conversion/doc_to_pixels.py#L313-L319

The way we use `pdftoppm` is the following:

1. Provide the PDF document as an argument.
2. Provide a temporary directory where `pdftoppm` will write PPM files (one per page).
3. Ask `pdftoppm` to report progress in stderr. Progress reports are lines with the following format: `<page> <num_pages> <file>`
4. Register a callback handler for the stderr of `pdftoppm`, which will be called with each line of the progress report as an argument.
5. (callback handler) For each progress line, read the file with the PPM data and get the following info: page width, page height, RGB data
6. (callback handler) Write this data as separate files under `/tmp/dangerzone`, which is the directory that the second phase will use to convert pixels to PDF.

# Bug :lady_beetle: 

Let's dig deeper into how the callback handler is called. We have a generic function that reads lines from a command's output:

https://github.com/freedomofpress/dangerzone/blob/18b73d94b0b0eb5a1dc7d990ca64885f4963cdad/dangerzone/conversion/common.py#L66-L74

For each line, it appends it to a buffer, and then calls a callback function with the line as an argument.

What's the bug here? We **first** read the line, and **then** check if we reached EOF :facepalm:. So, it's possible that we read the last line of the stream, and then immediately discard it, because we have indeed reached EOF. This means that the callback handler will not be called, and we will not create the necessary files under `/tmp/dangerzone` for the last page.

# Impact

This bug was introduced in version 0.4.1 (aeeed411a07699795aefae9f3feab84200c136bd). Users of affected versions **may** have documents with the last page missing. This bug should not have any impact on the security of the sanitization.

During our QA we never stubled into this bug, and we don't have any report from our users hinting to such an issue. I only managed to trigger it today, while working on something that made the callback handler run twice as slow.

If you have been affected though, please let us know. In any case, we will fix this issue ASAP.

# Remediation

Change the order of the checks: **first** check if we are at EOF, and **then** read the line. Note that we can't check the output of `readline()` for EOF (i.e, `if line == ""`), because it will detect empty lines as EOF.

The fix for this bug will be included in the upcoming 0.5.0 release.

apyrgio:
Stream page data back to the caller, immediately after we read them from pdftoppm. This way, we have more accurate progress reports and timeouts.

**IMPORTANT:** This PR contains a fix for a grave bug (#560) that we uncovered will working on this issue, and is required for making this issue work.

Fixes #557
Fixes #560

### DIFF
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -349,15 +349,21 @@ class DocumentToPixels(DangerzoneConverter):
             timeout=timeout,
         )
 
-        self.update_progress("Converted document to pixels")
-
-        # Move converted files into /tmp/dangerzone
-        for filename in (
+        final_files = (
             glob.glob("/tmp/page-*.rgb")
             + glob.glob("/tmp/page-*.width")
             + glob.glob("/tmp/page-*.height")
-        ):
+        )
+
+        # XXX: Sanity check to avoid situations like #560.
+        if not running_on_qubes() and len(final_files) != 3 * num_pages:
+            raise errors.PageCountMismatch()
+
+        # Move converted files into /tmp/dangerzone
+        for filename in final_files:
             shutil.move(filename, "/tmp/dangerzone")
+
+        self.update_progress("Converted document to pixels")
 
     async def install_libreoffice_ext(self, libreoffice_ext: str) -> None:
         self.update_progress(f"Installing LibreOffice extension '{libreoffice_ext}'")
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -22,6 +22,27 @@ from .common import DangerzoneConverter, running_on_qubes
 
 
 class DocumentToPixels(DangerzoneConverter):
+    # XXX: These functions write page data and metadata to a separate file. For now,
+    # they act as an anchor point for Qubes to stream back page data/metadata in
+    # real time. In the future, they will be completely replaced by their streaming
+    # counterparts. See:
+    #
+    # https://github.com/freedomofpress/dangerzone/issues/443
+    async def write_page_count(self, count: int) -> None:
+        pass
+
+    async def write_page_width(self, width: int, filename: str) -> None:
+        with open(filename, "w") as f:
+            f.write(str(width))
+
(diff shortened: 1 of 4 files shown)


dangerzone/conversion/doc_to_pixels.py
```diff
@@ -326,16 +346,26 @@ def pdftoppm_progress_callback(line: bytes) -> None:
...
             + glob.glob("/tmp/page-*.width")
             + glob.glob("/tmp/page-*.height")
-        ):
+        )
+
+        # XXX: Sanity check to avoid situations like #560.
+        if len(final_files) != 3 * num_pages:
+            raise errors.UnexpectedConversionError(
+                "The final number of pages does not correspond to the page count of the"
+                " original document"
```
deeplow, line 359: I don't think this will reach the client. We only pass the error code. So we may want a dedicated error message for this edge-case.

deeplow, line 359: Now this part only makes sense in the containers version. Not sure why this didn't trigger the exception in Qubes now that it doesn't generate any files.

Should we add here a check to see if it's not running on Qubes?

deeplow (commented):
I really like this refactor. It'll help a lot when doing #443. I have some little comments

dangerzone/conversion/doc_to_pixels.py
apyrgio: ... damn. Yes, I guess I didn't test this part in Qubes. I'll add a fixup.

apyrgio: Fixed in ed6722a73fbaac070aecd90c4b8fd2f3dccaf78a

deeplow (approved):
LGTM

PR #564, opened 2023-10-02
Generalize "out of RAM" error to reflect other issues

deeplow:
When qrexec-client-vm fails, it could be a symptom of various issues:
  - the system being out of RAM
  - dz-dvm not existing

The exit code is the same in all cases (126), which makes it particularly tricky to solve in the client application. For this reason the approach is now to tell the user to see the Qubes error notification on the top right of their screen.

### DIFF
--- a/dangerzone/conversion/errors.py
+++ b/dangerzone/conversion/errors.py
@@ -24,7 +24,7 @@ class ConversionException(Exception):
         return subclasses
 
 
-class QubesNotEnoughRAMError(ConversionException):
+class QubesConversionStartFailed(ConversionException):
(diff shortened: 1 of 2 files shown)


dangerzone/conversion/errors.py
```diff
@@ -27,8 +27,8 @@ def get_subclasses(cls) -> List[Type["ConversionException"]]:
 class QubesNotEnoughRAMError(ConversionException):
```
apyrgio, line 27: Might be worth changing the name of the error class as well?

deeplow: You're right. How does `QubesConversionStartFailed` sound?

apyrgio (approved):

PR #566, opened 2023-10-02
Prevent PDF from overwriting num_pages in Qubes

### LINKED ISSUE #565: Qubes: maliciously creafted PDF could limit the number of pages converted
7 months ago in commit [aeeed4](https://github.com/freedomofpress/dangerzone/commit/aeeed411a07699795aefae9f3feab84200c136bd)
  - `search = re.search(r"^Pages: (\d+)", line)` got replaced with
  - `search = re.search(r"Pages:\s*(\d+)\s*\n", stdout.decode())`
This code parsed the output of `pdfinfo` and this meant that if the PDF had metadata called `Pages: 1`, the end "safe" pdf would be 1 page long, regardless of the original length. 

[This]([sample-pdf-1-page.pdf](https://github.com/freedomofpress/dangerzone/files/12781937/sample-pdf-1-page.pdf)) document is a proof of concept. It has 4 pages but when converted (in Qubes) it results in 1 page. **This is not security-critical** since the Qubes port is in alpha and with no security guarantees and in the containers version (Windows, macOS and Linux) it only affects the timeout.

A fix is pretty simple:

```diff
diff --git a/dangerzone/conversion/doc_to_pixels.py b/dangerzone/conversion/doc_to_pixels.py
index 38f16150..48f301dc 100644
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -254,7 +254,7 @@ class DocumentToPixels(DangerzoneConverter):
             timeout=timeout,
         )
 
-        search = re.search(r"Pages:\s*(\d+)\s*\n", stdout.decode())
+        search = re.search(r"^Pages:\s*(\d+)\s*\n", stdout.decode(), re.MULTILINE)
         if search is not None:
             num_pages: int = int(search.group(1))
         else:
```

However, a more long-term fix is finding python bindings for `pdfinfo` or a similar tool with such bindings or python module.

deeplow:
This should only affect the alpha version of Qubes OS (in containers it only allows the attacker to control the timeout). In short, an attacker could have PDF metadata that would show before "Pages:" in the `pdfinfo` command output and this would essentially override the number of pages measured in the server. This could enable the attacker to shorten the number of pages of a document for example.

Fixes #565

### DIFF
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -254,7 +254,7 @@ class DocumentToPixels(DangerzoneConverter):
             timeout=timeout,
         )
 
-        search = re.search(r"Pages:\s*(\d+)\s*\n", stdout.decode())
+        search = re.search(r"^Pages:\s*(\d+)\s*\n", stdout.decode(), re.MULTILINE)
         if search is not None:
             num_pages: int = int(search.group(1))
         else:


apyrgio (approved):
Looks good, thanks!

Issue #567, opened 2023-10-02
Qubes: address remaining error handling


deeplow:
This is a follow up from #430, which tackled most of the error handling in Qubes but not all.

- [ ] If we cannot start disposable qubes due to RAM issues, then the user will be bombarded with error messages.
  * We should check if it makes sense to stop after the first failed conversion, assuming that there are no transient issues with `qrexec`.
- [ ] Get notified about the command that failed while doing document conversion:
  * In practice, we have at most three commands that may run:
     1. [not required for PDFs] `libreoffice` , `gm`
     2. `pdfinfo`
     3. `pdftoppm`
    
     An error during `pdftoppm` is easy to detect, as we are in the middle of getting pages back. Errors between `pdfinfo` and `libreoffice` | `gm` are less easy to detect, so this still warrants a fix, although it's not very pressing. I'd move this to the stabilization effort as well.

PR #568, opened 2023-10-02
qubes: Display all errors in second stage

### LINKED ISSUE #430: Qubes: Error handling

apyrgio:
If a command encounters an error or times out during the second stage of
the conversion in Qubes, handle it the same way as we would have handled
it in the first stage:

1. Get its error message.
2. Throw an UnexpectedConversionError exception, with the original
   message.

Note that, because the second stage takes place locally, users will see
the original content of the error.

Refs #567
Closes #430

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -352,7 +352,7 @@ class Container(IsolationProvider):
                 with open(log_path, "r", encoding="ascii", errors="replace") as f:
                     text = (
                         f"Container output: (pixels to PDF)\n"
-                        f"{PIXELS_TO_PDF_LOG_START}{f.read()}{PIXELS_TO_PDF_LOG_END}"
+                        f"{PIXELS_TO_PDF_LOG_START}\n{f.read()}{PIXELS_TO_PDF_LOG_END}"
                     )
                     log.info(text)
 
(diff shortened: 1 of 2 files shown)


deeplow (approved):
LGTM

Issue #528, opened 2023-08-22
Better "dark mode" support (on macOS)


deeplow:
Dark mode support in Dangerzone is less than stellar. Some points in particular:
  - We have some fonts set to black, which makes it really hard to read
  - the "[document]-safe" box is white and doesn't change to dark when dark mode is enabled:

|![black3](https://github.com/freedomofpress/dangerzone/assets/47065258/8486388b-dfdc-4cb1-bf9b-1d89bce1ad30) | ![black2](https://github.com/freedomofpress/dangerzone/assets/47065258/81fe4aea-4521-47f4-8280-14388abdf6bf) | ![black1](https://github.com/freedomofpress/dangerzone/assets/47065258/cc81f7a7-fa37-4de5-82b9-0c9ee1dda354) |
|---|---|---|

garrettr:
This is because these colors are set in `dangerzone.css`, which always sets the same colors regardless of the system color scheme.

We need to figure out how to:
1. Determine the current system color scheme.
2. Detect when the system color scheme is changed so we can keep dangerzone's appearance in sync with the system's if the system color scheme is changed while dangerzone is running.
3. Communicate the system color scheme to the stylesheet so it knows which colors to use.

IMHO 1 and 3 are essential, and 2 is a nice-to-have that could be relegated to a follow-up issue.

I'm inexperienced with Qt, so I did some research. This is what I've found so far:

## Determine the current system color scheme

### [Qt Group Blog | Dark Mode on Windows 11 with Qt 6.5](https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5)

> QStyleHints has a new property, colorScheme, that holds a Qt::ColorScheme enum value, either Qt::ColorScheme::Light or Qt::ColorScheme::Dark (or Qt::ColorScheme::Unknown on systems where there is no such system color scheme). This allows applications to determine whether a hand-crafted application palette should be dark or light, respecting the user's system preference. And like all good properties, it comes with a change notification signal QStyleHints::colorSchemeChanged that allows applications to respond to changes in the system color scheme.

This new property and signal would be useful, but it looks like we can't rely on features introduced in Qt 6.5 because dangerzone sometimes falls back to PySide2/Qt 5 due to #320.

For earlier versions of Qt, the post suggests the following method of inferring the current system color scheme:
```
static bool shouldApplyDarkFrame()
{
    // ...
    const QPalette defaultPalette;
[...]

PR #569, opened 2023-10-02
Add dark mode logic in our dialogs

apyrgio:
Make our Qt dialogs set the `OSColorMode` CSS property, same as the `MainWindow` class does, so that we can properly style them.

Refs #528

## Before this change

dark_mode_vs_updates

## After this change

dark_mode_correct

---

Note: I have not tested how this change affects dark mode, because I can't easily set it in my device. Given that this change mirrors what we do for `MainWindow`, I am confident that it should not cause a problem.

### DIFF
--- a/dangerzone/gui/logic.py
+++ b/dangerzone/gui/logic.py
@@ -135,6 +135,7 @@ class Dialog(QtWidgets.QDialog):
     ) -> None:
         super().__init__()
         self.dangerzone = dangerzone
+        self.setProperty("OSColorMode", self.dangerzone.app.os_color_mode.value)
 
         self.setWindowTitle(title)
         self.setWindowIcon(self.dangerzone.get_window_icon())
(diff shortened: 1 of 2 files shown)


deeplow:
It looks fine on linux:
looks_fine

deeplow:
Actually, in macOS it puts white text. But I think having this inconsistency is fine, for now.

dark

deeplow (approved):

apyrgio:
Nice, thanks for taking a look at this, and for the the Linux and MacOS screenshots. Regarding the white text, I see that this is the case for MacOS in the original PR as well (#550). I believe this inconsistency happens because we apply a font color globally, only in case of light mode, not dark mode. For the dark mode, we accept the system preferences.

This seems fine to me for now, but if it becomes an issue, we should revisit it.

PR #579, opened 2023-10-04
qubes: Clean up temporary files

### LINKED ISSUE #436: Qubes: allow parallel conversions

### LINKED ISSUE #575: Qubes: Clean up temporary files
Currently, our Qubes isolation provider and the `PixelsToPdf` converter use `/tmp/dangerzone` and `/tmp/page*` paths as their temporary storage space. The problem is that these paths are not cleared after a conversion, leading to stale files left behind.

apyrgio:
Create a temporary dir before the conversion begins, and store every file necessary for the conversion there. We are mostly concerned about the second stage of the conversion, which runs in the host. The first stage runs in a disposable qube and cleanup is implicit.

Fixes #575
Fixes #436

### DIFF
--- a/dangerzone/conversion/pixels_to_pdf.py
+++ b/dangerzone/conversion/pixels_to_pdf.py
@@ -17,22 +17,26 @@ from .common import DangerzoneConverter, running_on_qubes
 
 
 class PixelsToPDF(DangerzoneConverter):
-    async def convert(self, ocr_lang: Optional[str] = None) -> None:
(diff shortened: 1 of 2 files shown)


deeplow (approved):
LGTM!

PR #580, opened 2023-10-04
Catch out of /tmp space errors when converting documents

### LINKED ISSUE #578: Catch out of RAM errors in client and server
Many components in Dangerzone use temporary storage for the conversion but we are not catching all of these issues.

deeplow:
Many components in Dangerzone use temporary storage for the conversion. This catches some of those. The exception is catching them the containers isolation provider (host), since I couldn't find the right place to catch this, especially since `OSError` is such a generic error. So we can't have a catch-all thing for it.

Fixes #578


apyrgio:
My main problems with this approach are:

1. We catch `OSError`, without looking at the `errno`. We should check if it's [`ENOSPC`](https://docs.python.org/3/library/errno.html#errno.ENOSPC) (28) before handling it accordingly.
2. We catch this error very low in the stack, which is not maintainable in the long run.

Moreover, with PR #582 (side-effect), we actually show a nice error when the client is out of space:

current_error

So, as is, I'm afraid we can only gain something from this PR just for the containers version, and only on the server-side, provided we catch `ENOSPC` in a higher up in the stack. Any thoughts?

eloquence:
(Will not be merged into 0.5.0)

apyrgio:
@deeplow: I'd suggest closing this PR, given that we will no longer store lots of files in the sandbox with the PyMuPDF approach. Storing pixel data on host is an open issue, but a solution may come up there as well.

PR #582, opened 2023-10-05
Do not show file previews in case of exceptions

### LINKED ISSUE #581: Qubes: Do not open document previewer on error
During our experiments, we have seen that some failed conversions on Qubes lead to opening the document viewer regardless. Since the safe document does not exist, the user just sees the following error:

Screenshot_2023-10-03_15-56-21

apyrgio:
If a Qubes conversion encounters an exception that is not a subclass of ConversionException, it will still show a preview of a file that does not exist.

Send an error progress report in that case, so that the GUI code can detect that an error occurred and not open a file preview

Fixes #581


deeplow (approved):
LGTM

PR #587, opened 2023-10-08
Qubes: Fix issues with RPM packaging

apyrgio:
This PR fixes 3 small issues with the way we produce our Qubes RPM:

1. The `.exists()` method follows symlinks by default, whereas we want
   to check if a symlink exists. This functionality has been added in
   Python
   3.12.

   Instead of checking if a symlink exists and then removing it, simply
   remove it and don't throw an error if it doesn't exist in the first
   place.

2. The `dz.Convert*` policies were not installed with the executable bit
   set, therefore the qube could not start.

3. The `dz.ConvertDev` policy in particular had an ambiguous shebang,
   thus we change it to explicitly call Python3

Also, it introduces a CI action for checking if Qubes RPM packaging works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,12 @@ aliases:
       PATH=/root/.local/bin:$PATH ./install/linux/build-rpm.py
       ls -lh dist/
 
+  - &build-rpm-qubes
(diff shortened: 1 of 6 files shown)


deeplow (approved):
LGTM

PR #588, opened 2023-10-08
Provide fixes for small issues found during QA

apyrgio:
This PR fixes some general issues that we found during QA. Before merging it, we should update the "Change Selection" tests to make sure that they cover the scenario we fixed.


deeplow (approved):
Just added the "change selection" issue to the changelog and commit message so it can be referenced and noticed by users.

LGTM

deeplow (approved):

deeplow (approved):

PR #589, opened 2023-10-08
build-image.sh: Support building the image on Docker.

EtiennePerot:
This pull request modifies the `install/linux/build-image.sh` script to work with either `podman` or `docker` installed.

(It also enables some more bash features to make shell scripts less error-prone.)

### DIFF
--- a/install/linux/build-image.sh
+++ b/install/linux/build-image.sh
@@ -1,14 +1,25 @@
 #!/bin/sh
 
-set -e
+set -euo pipefail
 
 TAG=dangerzone.rocks/dangerzone:latest
 
-echo "Building container image"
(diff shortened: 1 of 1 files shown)


deeplow:
Thanks for the contribution. The issue is that for linux releases we want to rely on podman. Installing docker is rather painful and we're relying on an external repository. So a while ago we moved away from `docker` for this reason, I think.

And as I understand your gvisor PR is based on the assumption that Docker would be available on Linux, which may become a problem. Anyways, this is something that we can discuss with @apyrgio, but I don't think we'll move away from podman on linux (sadly for this and the other PR).

EtiennePerot:
Alright, closing this PR.

PR #590, opened 2023-10-08
Sandbox all document processing in gVisor

### LINKED ISSUE #126: Is it Possible to add gVisor Support to minimize the risk of an container escape exploit?
@micahflee @mikaelf  @haplo @legoktm  @mig5 @deeplow gVisor ( https://gvisor.dev/ ) provides a virtualized environment in order to sandbox containers. The system interfaces normally implemented by the host kernel are moved into a distinct, per-sandbox application kernel in order to minimize the risk of an container escape exploit. 

Sorry, If it is a dumb question, But as far as i see being OCI and runsc ,podman should be compatible with gvisor , so adding support for gVisor might help increase security. Is it possible with dangerzone or will there be any possible issues with this approach on linux?

 Incase if gvisor is compatible i guess it might be useful for high risk users.

### LINKED ISSUE #224: Defense in Depth - Prevent attacker from becoming root within the container
Parent issue: https://github.com/freedomofpress/dangerzone/issues/221

Once an attacker has found a flaw that allows them to take control of a process (see #222), they will probably attempt to become root within the container. Currently, this is not easy for two reasons:

1. The user within the container is a regular user (`dangerzone`, UID 1000).
2. We use `--security-opt=no-new-privileges`, so the kernel does not allow a process to elevate their privileges.

However, it may be worth ensureingthat it's not easy to become root in the container, even if this safeguard does not manage to work. For an extra layer of security, we could follow the latest CIS Docker advisories ([official link](https://www.cisecurity.org/benchmark/docker), [mirror](https://edu.anarcho-copy.org/GNU%20Linux%20-%20Unix-Like/Docker/CIS_Docker_Benchmark_v1.3.1_PDF.pdf)). 

Because the advisories are numerous, there are tools that check if a container image/runtime complies with the advisories:

* https://github.com/containers/podman-security-bench
* https://github.com/docker/docker-bench-security

I have already ran these two tools on Windows and Linux, and the noteworthy sections that were reported are the following:

* 4.3 - Ensure that unnecessary packages are not installed in the container
* 4.8 - Ensure setuid and setgid permissions are removed
  - The flagged binary was `sudo`, which is not used and can be removed.
* 5.12 - Ensure that the container's root filesystem is mounted as read only

We can easily placate section 4.8 by removing `sudo`, and section 5.12 by running the `podman` / `docker` command with the `--read-only` flag. The 4.3 section may require a bit more digging.

### LINKED ISSUE #225: Defense in Depth - Generate a restricted seccomp profile
Parent issue: https://github.com/freedomofpress/dangerzone/issues/221

Once an attacker has taken control of a process, and probably is root in the container (#224) they may attempt to escape the container by exploiting a vulnerability in the Linux kernel. The common interfaces that the Linux kernel provides are either special filesystems (#223) or system calls. In this issue, we will focus in the latter.

Containers run with a default seccomp filter that allows roughly 300 out of 435 system calls. Any vulnerability in the code path of these 300 system calls may be a way for an attacker to escape the container. So, it is very important to generate a seccomp profile for Dangerzone, that allowlists the system calls that are strictly necessary.

Further reading:
* https://podman.io/blogs/2019/10/15/generate-seccomp-profiles.html
* https://github.com/blacktop/seccomp-gen
* https://github.com/freedomofpress/dangerzone/issues/52 - seccomp mentioned but not implemented

### LINKED ISSUE #228: Defense in Depth - User Namespaces
Parent issue: https://github.com/freedomofpress/dangerzone/issues/221

User namespaces are very important, since they ensure that:
* Root within the container maps to the parent user outside the container.
* Users within the container map to non-existing users outside the container.

By ensuring that the user within the container (`dangerzone`, UID 1000) maps to a non-existing user outside the container, we complicate the attacker significantly. The current situation is:

* On Linux, we don't use user namespaces fully, since we run containers with `--userns keep-id`, which makes the `dangerzone` user within the container have the same UID as the user outside the container.
* On Windows/MacOS, they don't support user namespaces (see https://github.com/docker/for-win/issues/6897 and https://github.com/docker/for-mac/issues/3280 respectively).

## Linux

* Decide on a UID mapping (1000 inside the container, `x > 1000` outside the container) before starting the container.
* Create temporary directories for container I/O, owned by `x > 1000`  outside the container.
* Copy in the source files to the temporary directory for the first container (will also fix https://github.com/freedomofpress/dangerzone/issues/157)
* Run `podman` and specify the mapping for the container.
* Copy out the converted files.

## Windows/MacOS

Test Podman Desktop and check if it uses user namespaces.

Further reading:
* https://man7.org/linux/man-pages/man7/user_namespaces.7.html
* https://docs.docker.com/engine/security/userns-remap/

EtiennePerot:
When running on Linux, Dangerzone currently uses Podman with its default `crun`/`runc` runtime. These runtimes rely on Linux's built-in containerization primitives (namespaces). These parts of the Linux kernel have historically been the target of many container escape vulnerabilities. This is due to the fact that the Linux host kernel is fully exposed to the application running inside the container.

In particular, PDF and PostScript libraries such as Ghostscript have been notorious for having been targeted to run precisely this type of exploit. For this reason, while running these tools within Linux containers is better than running them directly on the host, it does not fully shield the host kernel from the malicious code that may be running in the container.

This pull request implements optional support for the [gVisor container runtime](https://gvisor.dev), called `runsc` ("run *sandboxed* container"). gVisor is an open-source OCI-compliant container runtime. It is a userspace reimplementation of the Linux kernel in a memory-safe language.

It works by creating a sandboxed environment in which regular Linux applications run, but their system calls are intercepted by gVisor. gVisor reinterprets these system calls using the logic in its own kernel written in Go, and responds to the system call by itself, rather than passing it onto the host Linux kernel. This means the host Linux kernel is isolated from the sandboxed application, thereby providing a significant level of protection against Linux container escape attacks.

gVisor further hardens itself by using the typical container primitives (isolating its own view of the host filesystem, running in the various types of namespaces that Linux support), and also sets a restrictive `seccomp-bpf` policy that only allows basic system calls through. This way, even if its userspace kernel were to get compromised, attackers would have to additionally have a "typical" Linux container escape vector, and that exploit would have to fit within the restricted `seccomp-bpf` rules that gVisor adds on itself.

This provides a level of protection comparable to a hardened hypervisor running workloads in a VM. However, gVisor doesn't actually use virtualization, so it is portable to all Linux environments and doesn't require virtualization support. It runs on x86 and ARM.

The initial commit of this pull request only adds support for using it inside `isolation_provider/container.py`. If there is appetite for this runtime, I'm happy to add CircleCI tests to integrate it better. Let me know what you think!

Fixes #126
Fixes #224
Fixes #225
Fixes #228

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN mkdir /libreoffice_ext && cd libreoffice_ext \
 ###########################################
 # Dangerzone image
 
-FROM alpine:latest
+FROM alpine:latest AS dangerzone-image
 
 # Install dependencies
 RUN apk --no-cache -U upgrade && \
@@ -68,15 +68,49 @@ COPY --from=h2orestart-dl /libreoffice_ext/ /libreoffice_ext
 
 RUN install -dm777 "/usr/lib/libreoffice/share/extensions/"
 
-ENV PYTHONPATH=/opt/dangerzone
-
 RUN mkdir -p /opt/dangerzone/dangerzone
 RUN touch /opt/dangerzone/dangerzone/__init__.py
 COPY conversion /opt/dangerzone/dangerzone/conversion
 
-# Add the unprivileged user
-RUN adduser -s /bin/sh -D dangerzone
+# Add the unprivileged user. Set the UID/GID of the dangerzone user/group to
+# 1000, since we will point to it from the OCI config.
+#
+# NOTE: A tmpfs will be mounted over /home/dangerzone directory,
+# so nothing within it from the image will be persisted.
+RUN addgroup -g 1000 dangerzone && \
+    adduser -u 1000 -s /bin/true -G dangerzone -h /home/dangerzone -D dangerzone
+
+###########################################
+# gVisor wrapper image
+
+FROM alpine:latest
+
+RUN apk --no-cache -U upgrade && \
+    apk --no-cache add python3
+
+RUN GVISOR_URL="https://storage.googleapis.com/gvisor/releases/release/latest/$(uname -m)"; \
+    wget "${GVISOR_URL}/runsc" "${GVISOR_URL}/runsc.sha512" && \
+    sha512sum -c runsc.sha512 && \
+    rm -f runsc.sha512 && \
+    chmod 555 runsc && \
+    mv runsc /usr/bin/
+
+# Add the unprivileged `dangerzone` user.
+RUN addgroup dangerzone && \
+    adduser -s /bin/true -G dangerzone -h /home/dangerzone -D dangerzone
+
+# Switch to the dangerzone user for the rest of the script.
 USER dangerzone
 
-# /safezone is a directory through which Pixels to PDF receives files
-VOLUME /safezone
+# Copy the Dangerzone image, as created by the previous steps, into the home
+# directory of the `dangerzone` user.
+RUN mkdir /home/dangerzone/dangerzone-image
+COPY --from=dangerzone-image / /home/dangerzone/dangerzone-image/rootfs
+
+# Create a directory that will be used by gVisor as the place where it will
+# store the state of its containers.
+RUN mkdir /home/dangerzone/.containers
+
+COPY gvisor_wrapper/entrypoint.py /
+
+ENTRYPOINT ["/entrypoint.py"]
--- a/dangerzone/gvisor_wrapper/entrypoint.py
+++ b/dangerzone/gvisor_wrapper/entrypoint.py
@@ -1,0 +1,161 @@
+#!/usr/bin/python3
+
+import json
+import os
+import shlex
+import subprocess
+import sys
+import typing
+
+# This script wraps the command-line arguments passed to it to run as an
+# unprivileged user in a gVisor sandbox.
+# Its behavior can be modified with the following environment variables:
+#   RUNSC_DEBUG: If set, print debug messages to stderr, and log all gVisor
+#                output to stderr.
+#   RUNSC_FLAGS: If set, pass these flags to the `runsc` invocation.
+# These environment variables are not passed on to the sandboxed process.
+
+
+def log(message: str, *values: typing.Any) -> None:
+    """Helper function to log messages if RUNSC_DEBUG is set."""
+    if os.environ.get("RUNSC_DEBUG"):
+        print(message.format(*values), file=sys.stderr)
+
+
+command = sys.argv[1:]
+if len(command) == 0:
+    log("Invoked without a command; will execute 'sh'.")
+    command = ["sh"]
+else:
+    log("Invoked with command: {}", " ".join(shlex.quote(s) for s in command))
+
+# Build and write container OCI config.
+oci_config: dict[str, typing.Any] = {
+    "ociVersion": "1.0.0",
+    "process": {
+        "user": {
+            # Hardcode the UID/GID of the container image to 1000, since we're in
+            # control of the image creation, and we don't expect it to change.
+            "uid": 1000,
+            "gid": 1000,
+        },
+        "args": command,
+        "env": [
+            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+            "PYTHONPATH=/opt/dangerzone",
+            "TERM=xterm",
+        ],
+        "cwd": "/",
+        "capabilities": {
+            "bounding": [],
+            "effective": [],
+            "inheritable": [],
+            "permitted": [],
+        },
+        "rlimits": [
+            {"type": "RLIMIT_NOFILE", "hard": 4096, "soft": 4096},
+        ],
+    },
+    "root": {"path": "rootfs", "readonly": True},
+    "hostname": "dangerzone",
+    "mounts": [
+        {
+            "destination": "/proc",
+            "type": "proc",
+            "source": "proc",
+        },
+        {
+            "destination": "/dev",
+            "type": "tmpfs",
+            "source": "tmpfs",
+            "options": ["nosuid", "noexec", "nodev"],
+        },
+        {
+            "destination": "/sys",
+            "type": "tmpfs",
+            "source": "tmpfs",
+            "options": ["nosuid", "noexec", "nodev", "ro"],
+        },
+        {
+            "destination": "/tmp",
+            "type": "tmpfs",
+            "source": "tmpfs",
+            "options": ["nosuid", "noexec", "nodev"],
+        },
+        # LibreOffice needs a writable home directory, so just mount a tmpfs
+        # over it.
+        {
+            "destination": "/home/dangerzone",
+            "type": "tmpfs",
+            "source": "tmpfs",
+            "options": ["nosuid", "noexec", "nodev"],
+        },
+        # Used for LibreOffice extensions, which are only conditionally
+        # installed depending on which file is being converted.
+        {
+            "destination": "/usr/lib/libreoffice/share/extensions/",
+            "type": "tmpfs",
+            "source": "tmpfs",
+            "options": ["nosuid", "noexec", "nodev"],
+        },
+    ],
+    "linux": {
+        "namespaces": [
+            {"type": "pid"},
+            {"type": "network"},
+            {"type": "ipc"},
+            {"type": "uts"},
+            {"type": "mount"},
+        ],
+    },
+}
+not_forwarded_env = set(
+    (
+        "PATH",
+        "HOME",
+        "SHLVL",
+        "HOSTNAME",
+        "TERM",
+        "PWD",
+        "RUNSC_FLAGS",
+        "RUNSC_DEBUG",
+    )
+)
+for key_val in oci_config["process"]["env"]:
+    not_forwarded_env.add(key_val[: key_val.index("=")])
+for key, val in os.environ.items():
+    if key in not_forwarded_env:
+        continue
+    oci_config["process"]["env"].append("%s=%s" % (key, val))
+if os.environ.get("RUNSC_DEBUG"):
+    log("Command inside gVisor sandbox: {}", command)
+    log("OCI config:")
+    json.dump(oci_config, sys.stderr, indent=2, sort_keys=True)
+    # json.dump doesn't print a trailing newline, so print one here:
+    log("")
+with open("/home/dangerzone/dangerzone-image/config.json", "w") as oci_config_out:
+    json.dump(oci_config, oci_config_out, indent=2, sort_keys=True)
+
+# Run gVisor.
+runsc_argv = [
+    "/usr/bin/runsc",
+    "--rootless=true",
+    "--network=none",
+    "--root=/home/dangerzone/.containers",
+]
+if os.environ.get("RUNSC_DEBUG"):
+    runsc_argv += ["--debug=true", "--alsologtostderr=true"]
+if os.environ.get("RUNSC_FLAGS"):
+    runsc_argv += [x for x in shlex.split(os.environ.get("RUNSC_FLAGS", "")) if x]
+runsc_argv += ["run", "--bundle=/home/dangerzone/dangerzone-image", "dangerzone"]
+log(
+    "Running gVisor with command line: {}", " ".join(shlex.quote(s) for s in runsc_argv)
+)
+runsc_process = subprocess.run(
+    runsc_argv,
+    check=False,
+)
+log("gVisor quit with exit code: {}", runsc_process.returncode)
+
+# We're done.
+sys.exit(runsc_process.returncode)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 - Fix a deprecation warning in PySide6, thanks to [@naglis](https://github.com/naglis) ([issue #595](https://github.com/freedomofpress/dangerzone/issues/595))
 - Make update notifications work in systems with PySide2, thanks to [@naglis](https://github.com/naglis) ([issue #788](https://github.com/freedomofpress/dangerzone/issues/788))
 
+### Security
+
+- Integrate Dangerzone with gVisor, a memory-safe application kernel, thanks to [@EtiennePerot](https://github.com/EtiennePerot) ([#126](https://github.com/freedomofpress/dangerzone/issues/126))
+  As a result of this integration, we have also improved Dangerzone's security
+  in the following ways:
+  * Prevent attacker from becoming root within the container ([#224](https://github.com/freedomofpress/dangerzone/issues/224))
+  * Use a restricted seccomp profile ([#225](https://github.com/freedomofpress/dangerzone/issues/225))
+  * Make use of user namespaces ([#228](https://github.com/freedomofpress/dangerzone/issues/228))
+
 ## Dangerzone 0.6.1
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ See [installing Dangerzone](INSTALL.md#linux) for adding the Linux repositories 
 ## Some features
 
 - Sandboxes don't have network access, so if a malicious document can compromise one, it can't phone home
+- Sandboxes use [gVisor](https://gvisor.dev/), an application kernel written in Go, that implements a substantial portion of the Linux system call interface.
 - Dangerzone can optionally OCR the safe PDFs it creates, so it will have a text layer again
 - Dangerzone compresses the safe PDF to reduce file size
 - After converting, Dangerzone lets you open the safe PDF in the PDF viewer of your choice, which allows you to open PDFs and office docs in Dangerzone by default so you never accidentally open a dangerous document
--- a/dangerzone/conversion/common.py
+++ b/dangerzone/conversion/common.py
@@ -14,7 +14,9 @@ def running_on_qubes() -> bool:
 
 
 def get_tessdata_dir() -> str:
-    if running_on_qubes():
+    if os.environ.get("TESSDATA_PREFIX"):
+        return os.environ["TESSDATA_PREFIX"]
+    elif running_on_qubes():
         return "/usr/share/tesseract/tessdata/"
     else:
         return "/usr/share/tessdata/"
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -86,6 +86,49 @@ class Container(IsolationProvider):
         return runtime
 
     @staticmethod
+    def get_runtime_security_args() -> List[str]:
+        """Security options applicable to the outer Dangerzone container.
+
+        Our security precautions for the outer Dangerzone container are the following:
+        * Do not let the container assume new privileges.
+        * Drop all capabilities, except for CAP_SYS_CHROOT, which is necessary for
+          running gVisor.
+        * Do not allow access to the network stack.
+        * Run the container as the unprivileged `dangerzone` user.
+
+        For Podman specifically, where applicable, we also add the following:
+        * Do not log the container's output.
+        * Use a newer seccomp policy (for Podman 3.x versions only).
+        * Do not map the host user to the container, with `--userns nomap` (available
+          from Podman 4.1 onwards)
+          - This particular argument is specified in `start_doc_to_pixels_proc()`, but
+            should move here once #748 is merged.
+        """
+        if Container.get_runtime_name() == "podman":
+            security_args = ["--log-driver", "none"]
+            security_args += ["--security-opt", "no-new-privileges"]
+
+            # NOTE: Ubuntu Focal/Jammy have Podman version 3, and their seccomp policy
+            # does not include the `ptrace()` syscall. This system call is required for
+            # running gVisor, so we enforce a newer seccomp policy file in that case.
+            # This file has been copied as is [1] from the official Podman repo.
+            #
+            # [1] https://github.com/containers/common/blob/d3283f8401eeeb21f3c59a425b5461f069e199a7/pkg/seccomp/seccomp.json
+            if Container.get_runtime_version() < (4, 0):
+                seccomp_json_path = get_resource_path("seccomp.gvisor.json")
+                security_args += ["--security-opt", f"seccomp={seccomp_json_path}"]
+        else:
+            security_args = ["--security-opt=no-new-privileges:true"]
+
+        security_args += ["--cap-drop", "all"]
+        security_args += ["--cap-add", "SYS_CHROOT"]
+
+        security_args += ["--network=none"]
+        security_args += ["-u", "dangerzone"]
+
+        return security_args
+
+    @staticmethod
     def install() -> bool:
         """
         Make sure the podman container is installed. Linux only.
@@ -218,25 +261,12 @@ class Container(IsolationProvider):
         extra_args: List[str] = [],
     ) -> subprocess.Popen:
         container_runtime = self.get_runtime()
-
-        if self.get_runtime_name() == "podman":
-            security_args = ["--log-driver", "none"]
-            security_args += ["--security-opt", "no-new-privileges"]
-            security_args += ["--userns", "keep-id"]
-        else:
-            security_args = ["--security-opt=no-new-privileges:true"]
-
-        # drop all linux kernel capabilities
-        security_args += ["--cap-drop", "all"]
-        user_args = ["-u", "dangerzone"]
+        security_args = self.get_runtime_security_args()
         enable_stdin = ["-i"]
         set_name = ["--name", name]
-
         prevent_leakage_args = ["--rm"]
-
         args = (
-            ["run", "--network", "none"]
-            + user_args
+            ["run"]
             + security_args
             + prevent_leakage_args
             + enable_stdin
@@ -245,7 +275,6 @@ class Container(IsolationProvider):
             + [self.CONTAINER_NAME]
             + command
         )
-
         args = [container_runtime] + args
         return self.exec(args)
 
@@ -291,6 +320,36 @@ class Container(IsolationProvider):
             "-e",
             f"OCR_LANGUAGE={ocr_lang}",
         ]
+        # XXX: Until #748 gets merged, we have to run our pixels to PDF phase in a
+        # container, which involves mounting two temp dirs. This does not bode well with
+        # gVisor for two reasons:
+        #
+        # 1. Our gVisor integration chroot()s into /home/dangerzone/dangerzone-image/rootfs,
+        #    meaning that the location of the temp dirs must be relevant to that path.
+        # 2. Reading and writing to these temp dirs requires permissions which are not
+        #    available to the user within gVisor's user namespace.
+        #
+        # For these reasons, and because the pixels to PDF phase is more trusted (and
+        # will soon stop being containerized), we circumvent gVisor support by doing the
+        # following:
+        #
+        # 1. Override our entrypoint script with a no-op command (/usr/bin/env).
+        # 2. Set the PYTHONPATH so that we can import the Python code within
+        #    /home/dangerzone/dangerzone-image/rootfs
+        # 3. Run the container as the root user, so that it can always write to the
+        #    mounted directories. This container is trusted, so running as root has no
+        #    impact to the security of Dangerzone.
+        img_root = "/home/dangerzone/dangerzone-image/rootfs"
+        extra_args += [
+            "--entrypoint",
+            "/usr/bin/env",
+            "-e",
+            f"PYTHONPATH={img_root}/opt/dangerzone:{img_root}/usr/lib/python3.12/site-packages",
+            "-e",
+            f"TESSDATA_PREFIX={img_root}/usr/share/tessdata",
+            "-u",
+            "root",
+        ]
 
         name = self.pixels_to_pdf_container_name(document)
         pixels_to_pdf_proc = self.exec_container(command, name, extra_args)
@@ -329,8 +388,15 @@ class Container(IsolationProvider):
             "-m",
             "dangerzone.conversion.doc_to_pixels",
         ]
+        # NOTE: Using `--userns nomap` is available only on Podman >= 4.1.0.
+        # XXX: Move this under `get_runtime_security_args()` once #748 is merged.
+        extra_args = []
+        if Container.get_runtime_name() == "podman":
+            if Container.get_runtime_version() >= (4, 1):
+                extra_args += ["--userns", "nomap"]
+
         name = self.doc_to_pixels_container_name(document)
-        return self.exec_container(command, name=name)
+        return self.exec_container(command, name=name, extra_args=extra_args)
 
     def terminate_doc_to_pixels_proc(
         self, document: Document, p: subprocess.Popen
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -7,7 +7,7 @@ import shlex
 import shutil
 import subprocess
 import sys
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 from ..conversion import errors
 from ..document import Document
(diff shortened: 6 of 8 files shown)


deeplow:
Thanks for the contribution! We are just wrapping up a release and I'll come back to this after we're done with that.

EtiennePerot:
As noted in https://github.com/freedomofpress/dangerzone/pull/589#issuecomment-1759343903, the usage of Docker is problematic. So I have updated this PR to use gVisor with Podman rather than Docker. Let me know if this is more interesting, and I can add tests, QA instructions, and better documentation.

EtiennePerot:
Thanks for the details; great to see there's already been lots of investigatory work on this.

I actually had independently tried to implement it via the "in-container `runsc do`" mechanism you mention; it would indeed mean better security on Windows/Mac, because it would protect the two containers from each other (otherwise, AIUI they run in the same VM that Docker desktop manages).  I actually got it to mostly work with `runsc do`; the caveats was that instead of using rootless `runsc` in an unprivileged container, it runs it as `root` in a privileged container, and wraps the command that runs inside the gVisor sandbox by prefixing it with `sudo dangerzone --`. This sounds scary, but I actually don't think that's such a problem, given that the security boundary would move to gVisor rather than to the container boundary: with such a setup, the role of Docker/Podman would no longer be that of a security boundary, and instead just fulfill the role of a cross-platform software portability solution of sorts.

The only remaining problems with that approach I had were around file permissions and ownership of the files written back to the out-of-sandbox filesystem (fixable but probably requires me to dig more into where such fix-ups would need to occur), but also how `runsc do` doesn't have good control over which directories in its sandbox are writable back to the host and which aren't. Relying on gVisor's enforcement of file permissions would work but seems like it's sub-optimal. #443 sounds like it partially solves this, but also it'd be nice to just support this better in `runsc` itself.

Alternatively, manually creating an OCI container spec and starting it within runsc-in-Podman by using manual invocations of `runsc create` + `runsc start` would allow fine-grained control over mounted directories without needing `runsc` modifications. `runsc do` is basically just a helper command to do just that, it just doesn't have command-line flags to control the "volumes" part of that container spec.

If that sounds like an interesting direction, then let me know and I can go that way rather than the "just make it work in Linux with Podman" approach

deeplow:
Nice to see that your exploration has crossed the path with @apyrgio's.

> The only remaining problems with that approach I had were around file permissions and ownership of the files written back to the out-of-sandbox filesystem (fixable but probably requires me to dig more into where such fix-ups would need to occur)

This won't be a problem one we fix #443 (which we probably need to do for the next release anyways since we want to have the Qubes code shared as much as possible with the containers one).

apyrgio:
Hey, sorry for the delay @EtiennePerot, life got a bit in the way :wink: 

> I actually had independently tried to implement it via the "in-container runsc do" mechanism you mention

Hey, that's great!

> with such a setup, the role of Docker/Podman would no longer be that of a security boundary, and instead just fulfill the role of a cross-platform software portability solution of sorts.

I was thinking about this backwards, that the main sandbox would be the Docker/Podman runtime, and gVisor would supplement it with it's kernel protection capabilities. The way you think about this makes sense though, having a very strong inner sandbox, and using the outer sandbox for cross platform support.

Before going down the full privileged container route, I'd try to figure out which capabilities are necessary to make this work, so that we minimize the blast radius in case of a `runsc` escape. One would argue that the blast radius would be the same as in the case of a `runc` escape of course, but a little defense in depth here wouldn't hurt.

> The only remaining problems with that approach I had were around file permissions and ownership of the files written back to the out-of-sandbox filesystem (fixable but probably requires me to dig more into where such fix-ups would need to occur),

Quick note regarding this: Since #443 is a blocker for oh-so-many-things, we've already started working on this and will soon have a prototype. I'd suggest you consider this as a non-issue, i.e., that we won't mount any directories into the container.

> If that sounds like an interesting direction, then let me know and I can go that way rather than the "just make it work in Linux with Podman" approach.

It very much is :slightly_smiling_face: . If you're up for it, feel free to let us know. Else, we can try to tackle it once we're done with #443, and some Qubes fixes. In any case, thanks a lot for your persistence on this issue and sharing your gVisor knowledge.

apyrgio:
Quick update here: we have some PRs under way (see #622, #627) that fix #443, at least for the first stage of the conversion. We can start experimenting with gVisor on top of them, but this task will be much clearer once the PRs are merged.

EtiennePerot:
Yes, I've been following that :)

Still planning on getting around to this PR.

EtiennePerot:
Just an update here: I have successfully managed to run `runsc` within podman in rootless mode, with the file permissions all lining up correctly for the `/safezone` files. All tests pass under `podman`. It still requires some changes to gVisor that haven't been released yet, so not yet ready to review.

I am wondering how Docker support works currently. I don't have a Windows or OS X machine to test on, but how are file permissions handled for these platforms? My mental model of how Docker works on these operating systems is by creating a Linux VM where the Docker daemon runs as root. How do volumes get mounted in it? Does everything in mounted volume appear as being owned by root as well, or are there user mappings?

EtiennePerot:
Uploaded a new commit that uses the approach of putting gVisor *inside* the Dangerzone container image, and use it as a wrapper regardless of the container runtime. It works with both Podman and Docker, although I have only tested on Linux and I am not sure how file permissions with Docker volumes work on non-Linux platforms.

Also, the tests will fail because the latest gVisor release does not yet include https://github.com/google/gvisor/commit/88f7bb66f0dc4607d67281d6d61040f98375166b. I have tested it against a freshly-compiled gVisor binary and all tests pass on my end. I expect https://github.com/google/gvisor/commit/88f7bb66f0dc4607d67281d6d61040f98375166b will be part of next week's gVisor release.

Please read the long comment in `dangerzone/gvisor_wrapper/entrypoint.py` which explains the approach and why it works the way it does... It is difficult to get all the permissions lined up properly.

Let me know what you think!

apyrgio:
Woah, that's exciting! We're currently in the midst of releasing Dangezone 0.6.0 so I can't take a proper look right now, but I promise to do so as soon as possible.

One quick comment I have while skimming through the comments for the entrypoint is this: I understand that the `/safezone` mount is a factor that makes things more complicated. In practice though, it shouldn't affect the gVisor work. There are two reasons for that:

1. In practice, we want to use gVisor for the sanitization that takes place on the first container. This container is the most sensitive one, since it's the only one [1] which is affected by RCEs. The good thing about this container is that we don't mount anything to it. It just reads the document from stdin and writes the pixels to stdout.
2. The `/safezone` mount point will soon be removed, along with the second container. We are now very close to running the conversion of pixels to PDF on the user's host, which has several important benefits (smaller container image size, faster OCR, ability to update the container image from an image repository etc.)

So, in an nutshell, I believe that we can simplify things more, by targeting just the first container, and not worry about mounts. But I'll have more concrete things to comment on once we're done with 0.6.0.

[1] Well, there's always the scenario where the code that converts pixels to PDFs is subject to an RCE, but that's a very narrower attack surface, and we can't effectively defend against it. In Qubes, for instance, this part runs on the user's host.

EtiennePerot:
Thanks for the details. I believe removing the need to have the `/safezone` mount would indeed simplify things a lot; specifically, it removes the need to have precise UID:GID alignment in the unsandboxed entrypoint, and it totally removes the need for the sandboxed entrypoint, as well as the three capabilities granted inside the gVisor sandbox. (Part of my thought process while building this was actually "can we just not have this mount at all and instead pass the files back and forth via something like a tarball over stdin/stdout?", but I guess simply not running the pixels-to-PDF part in a container solves that problem too.)

apyrgio:
Alright, I looked more carefully into the PR. I have several questions, some of those are just basic gVisor questions, and some apply to Dangerzone specifically. Here goes:

1. Assuming that we use `runsc` only on the first container, what is the practical difference between using `runsc do` and `runsc run`?

   **Context**: The reason I'm asking is because the following invocation :

   ```
   runsc --rootless --network=none do env PYTHONPATH=/opt/dangerzone \
      python3 -m dangerzone.conversion.doc_to_pixels
   ```
  
    works when I test things locally, and in the spirit of keeping things as simple as possible, it's very appealing. However, I do read that this option is for testing only, so I'm skeptical if there's a footgun that I'm missing.
    
    _(I've seen your OCI config btw, and I already spot some differences, but I'm mainly asking if there's anything foundational that I'm missing)_

2. Before including gVisor support, we should clarify what are the security guarantees that the outer container provides (Podman/Docker), and what are the security guarantees that the inner container providers (gVisor). From the code and the documentation on the entrypoint script, I understand that the outer layer:
   * Does no longer drop capabilities. This is the responsibility of the inner layer.
   * Does no longer set a seccomp filter. This is done by the inner layer, and in a more strict fashion as well.
   
   Anything else that's changed?
   
3. We have a pending issue for supporting user namespaces (#228). This is something that Docker does not support, but gVisor does, so that's great! My main issue with user namespaces is that UID 0 typically maps to the owner of the namespace in the host. In Linux (Podman) that's the user who starts the Dangerzone application. In Windows / macOS, it's the `root` user that runs in the WSL/HyperKit VM. What's the case in gVisor?

    _(My plan was to use something like [`userns=nomap`](https://docs.podman.io/en/v4.6.1/markdown/options/userns.container.html) for supported Podman versions. If gVisor can support something similar across OSes, that would be amazing)_

apyrgio:
@EtiennePerot kind ping on the above questions, so that we don't lose context.

EtiennePerot:
Hey, sorry I missed the set of questions.

> Assuming that we use `runsc` only on the first container, what is the practical difference between using `runsc do` and `runsc run`?

`runsc do` is just a convenience helper. It is true that under the hood it's basically just doing `runsc create` + `runsc run`, but there are many knobs that `runsc do` doesn't expose. For example, `runsc do` gives read-only access to the whole host filesystem by default, which is unnecessarily wide for Dangerzone. There are also some subtleties around how `runsc do` operates to act as a convenience wrapper; for example, in root-ful mode with networking enabled, it'll [do a whole dance with network namespace setup](https://github.com/google/gvisor/blob/9b28a3f2eeff35d57549f96842ac4f7c39a2911a/runsc/cmd/do.go#L262). It's meant to be for convenience, not so much for the tightest possible security.

But beyond the OCI differences (which I think are worth digging into), the main reason is simply supportability. `runsc create` and `runsc run` as specified as part of the [OCI runtime interface spec](https://github.com/opencontainers/runtime-spec), whereas `runsc do` is just a gVisor-specific helper. It's therefore less stable of an API and it makes no guarantee to keep working the way it does into the future (e.g. it makes no difference that its spec will stay the same and not make further security compromises). I think it's worth avoiding `runsc do` for that reason alone.

*(Well, technically the OCI spec only specifies `runsc start` rather of `runsc run`, but the interface `runsc run` takes is meant to be the same as `runsc start` so I still think the argument applies. Though I could edit the script to use `runsc start` if you think it's better.)*

In terms of practical OCI spec differences:

- `runsc do` enables terminal emulation by default. This is a notoriously complicated part of the POSIX API, and dangerzone doesn't need it, so better to have it disabled.
- `runsc do` would mount the whole filesystem as `/` by default (as per above), and does so non-readonly. (In practice it's still read-only because gVisor implements an overlay on top of the root filesystem, but specifying the OCI spec also allows explicitly marking the lower level of the overlay as read-only.). This also means that the in-sandbox workload can see files such as the out-of-sandbox entrypoint (`entrypoint.py`) and the `runsc` binary itself, i.e. it can trivially learn information about its own container spec. The manual spec only mounts specific parts of the filesystem that Dangerzone actually needs.
- `runsc do` doesn't allow setting mount options, whereas the manual spec can specify mount options like `nosuid`, `noexec`, `nodev` to further restrict what the workload can do with files exposed from `/safezone`.
- `runsc do` doesn't make guarantees about which host namespaces it isolates itself in, whereas the manual OCI spec causes `runsc` to isolate itself in a separate host PID+Network+IPC+UTS+mount namespace.
- `runsc do` doesn't allow specifying an in-sandbox seccomp filter. We could add one for more defense-in-depth; see below.

> Before including gVisor support, we should clarify what are the security guarantees that the outer container provides (Podman/Docker), and what are the security guarantees that the inner container providers (gVisor). [...] Anything else that's changed?

I think the main thing is, well, that the inner layer now has gVisor in the middle. I don't want to sound too salesman-y, but gVisor is its own full-fledged POSIX kernel implementation. It's not just one security measure being addressed at the inner vs outer layer; for some of them, it means that there are now *two* distinct implementations of the same security measure that need to _simultaneously_ be exploited in order to breach through.

For example, if the sandboxed workload is running as an non-root user inside the gVisor sandbox, and the gVisor sandbox is itself running as an unprivileged user on the host, then for the sandboxed workload to escalate to root on the host, they would need to have a user escalation exploit for both kernels (and, since those two kernels don't share code, the same exploit generally won't simultaneously work against both). Same thing for filesystem-level isolation: that security measure exists at *both* levels. More generally speaking, it takes at least two unique kernel vulnerabilities (one specifically against the gVisor kernel, plus one specifically against the Linux kernel) in order to fully break out onto the host system. On Windows/OSX, it takes one gVisor kernel exploit plus a VM escape exploit to fully break out.

There are some layers that aren't "redundant" in this manner, like the seccomp one you point out, although actually it would be possible to add a seccomp filter enforced _within the gVisor sandbox_ (using gVisor's `seccomp-bpf` implementation) on top of what's already here. I can add that to this PR if you wish. But if we're going to do that, I actually think there's potential to go beyond applying something like Docker's default seccomp filter, and come up with a fine-tuned seccomp filter that specifically allows strictly what the PDF-to-pixel program needs.

In terms of thinking of security _responsibilities_ (i.e. what each level "ought to" provide), at a high-level, I think the framing in an earlier comment is pretty well formulated: the outer container's main responsibility is to act as a "platform compatibility" solution, whereas the inner container's responsibility is solely security. The outer layer's own security measures (e.g. filesystem isolation, non-privileged user) can be seen as just an added security bonus, so long as they don't interfere with the inner layer's ability to work properly.

> We have a pending issue for supporting user namespaces [...] In Linux (Podman) that's the user who starts the Dangerzone application. In Windows / macOS, it's the root user that runs in the WSL/HyperKit VM. What's the case in gVisor?

The user namespace handling is probably the most complicated part of the current implementation, because of the need to preserve UIDs on Linux so that files in the `/safezone` volumes are mapped to the user's UID on the host. As long as that remains necessary, then running as a user that ultimately maps to that UID on the host is unavoidable.

However, if, as per the above discussion, we can get rid of this, then all of the current user namespace stuff can be simplified and further locked down to have no relationship with any existing user on the host system. This means we could create a user that only exists inside the outer container, and run the gVisor process as that user in a user namespace that exposes no other user.

Then, on top of that, since gVisor is its own kernel and thus implements its own notion of users and user namespaces, the workload within the sandbox can itself run in an in-sandbox user namespace that doesn't have a mapping to the host user (i.e. to the user that the gVisor process runs as). (The current implementation of this PR _kind of_ does this already: it has a "UID 1000" which has no mapping to any user on the host, but it still maps in the in-sandbox "UID 0" to the host UID that owns `/safezone` so that file permissions can still line up properly.)

apyrgio:
Thanks Etienne for answering all the my questions in great detail. Not only I'm covered, but I think we have enough material to update the parent issue, and write down a design document. I plan to follow up on the above on Monday, and maybe offer some next steps. My guess is that our lives will be much easier once we've tackled #625, so I'll make sure to prioritize it next week.

EtiennePerot:
Sounds good. One small question: which issue do you mean by "updating the parent issue"?

I agree that addressing #625 first makes sense, otherwise this PR would add temporary complexity that doesn't need to ever exist if #625 is addressed first. If you wish, I can already start simplifying this PR to what it _would_ look like if it only needed to support PDF-to-pixels conversion.

apyrgio:
Quick update here. I actually prioritized implementing the on-host pixels to PDF conversion PR (https://github.com/freedomofpress/dangerzone/pull/748), which is a prerequisite for vastly simplifying this one. Now that it's out, I'll follow up here soon.

EtiennePerot:
I've updated this PR to remove the support for the `/safezone` volume. As promised, it simplifies `entrypoint.py` a whole lot.
Also, thanks for pointing out Podman's `--userns=nomap` option. Turns out that already does what I was otherwise planning to do manually from within `entrypoint.py` to run as a user that's as unprivileged as possible, so that simplifies things further. I've added this flag to `container.py`.

dangerzone/gvisor_wrapper/entrypoint.py
```diff
@@ -0,0 +1,267 @@
...
+
+if args.gvisor_debug:
+    print(
+        "Running gVisor with command line:",
+        " ".join(shlex.quote(s) for s in runsc_argv),
+        file=sys.stderr,
+    )
+runsc_process = subprocess.run(
+    runsc_argv,
+    executable=runsc_binary,
```
apyrgio, line 260: Minor, but what's the case for `executable=` here (and `os.path.basename(runsc_binary)` above)? Is it for stylistic reasons?

dangerzone/gvisor_wrapper/entrypoint.py
```diff
@@ -0,0 +1,267 @@
...
+    print("No command line specified.", file=sys.stderr)
+    parser.print_help()
+    sys.exit(2)
+
+# We can't pass `sys.argv` directly to `parser` here because this may end up
+# processing flags from the wrapped command line that happen to match flags
+# that are defined in `parser`. For example, if the user were to somehow
+# make a file named `--pre_gvisor`, that would be pretty bad.
+# So we scan for the first argument that does not look like it is intended
+# for this entrypoint script, and we only pass this subset to `parser`.
```
apyrgio, line 54: This is probably a relic from a previous iteration of Dangerzone, but we no longer let attack-controlled strings in our (tokenized) command arguments. Other than that though, I think we can simplify the argument parsing here, by doing the following:

1. Make the container isolation provider explicitly run the full `podman ... runsc ...` command. This will make things more verbose, but less magic. Also, it will make development faster, since we will have just one place to tweak gvisor arguments, and not recompile the Dangerzone image. 
2. In the entrypoint script, check if the basename of `sys.argv[0]` is  `runsc`. In that case, create the OCI config.
3. Do not handle any extra arguments in the entrypoint script, since they will provided by the isolation provider, if necessary.

apyrgio (commented):
Nice, thanks a lot for updating the PR! I think we're in a very good shape, from a cursory look of the PR. I have a few minor comments, but other than that, nothing important that comes to mind.

And yeah, I still owe a design document, but we're currently in hotfix mode, so I didn't have time to work on that yet. Once we're done with the hotfix, we can switch focus to this PR once more, and merge it :slightly_smiling_face:

Dockerfile
```diff
@@ -59,23 +59,43 @@ RUN apk --no-cache -U upgrade && \
...
 RUN touch /opt/dangerzone/dangerzone/__init__.py
 COPY conversion /opt/dangerzone/dangerzone/conversion
 
-# Add the unprivileged user
-RUN adduser -s /bin/sh -D dangerzone
-USER dangerzone
+# Add the unprivileged user.
+# NOTE: A tmpfs will be mounted over /home/dangerzone directory,
+# so nothing within it from the image will be persisted.
+ARG DANGERZONE_UID=65042
```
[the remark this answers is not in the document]
apyrgio: Saw it, thanks.

dangerzone/gvisor_wrapper/entrypoint.py
apyrgio: I like the way you simplified this, I think we're good.

apyrgio:
The latest branch looks almost ready for inclusion. I want to do a last pass, document our architectural choices, and run tests on every platform. The latter are currently failing, so I'll try to debug a bit there.

apyrgio:
While experimenting with this PR, I realized that we can re-introduce `--cap-drop all`, if we add the following capabilities: `SETFCAP` and `SYS_CHROOT`. It's not much for the security of the outer layer (Podman/Docker), but it's something.

Also, my understanding is that the sandboxed UID/GID that we search for with `chroot /dangerzone-image/rootfs id -u dangerzone` will always be 1000, so we can remove these commands. Note that the `SYS_CHROOT` capability will still be required by `runsc`, so we can't remove it from the aforementioned list.

apyrgio:
Heads up, I have a design document ready, that should explain how the gVisor integration works in Dangerzone, to people who have not seen the code: https://github.com/freedomofpress/dangerzone/pull/815/commits/8641b66b0db634d1b6b849f9047a93671d7c5a13

@EtiennePerot if you feel I have missed something important, or that I've explained something incorrectly, feel free to chime in. In the meantime, I'll take a look at the CI build failures.

EtiennePerot:
> While experimenting with this PR, I realized that we can re-introduce `--cap-drop all`, if we add the following capabilities: `SETFCAP` and `SYS_CHROOT`. It's not much for the security of the outer layer (Podman/Docker), but it's something.

I'm not at my development machine now but will do when I get the chance.

> Also, my understanding is that the sandboxed UID/GID that we search for with `chroot /dangerzone-image/rootfs id -u dangerzone` will always be 1000, so we can remove these commands. Note that the `SYS_CHROOT` capability will still be required by `runsc`, so we can't remove it from the aforementioned list.

I'm not sure that's actually always the case; it depends on which UID/GID the `adduser`/`addgroup` commands use during the image building process. Now that they no longer have an explicit UID/GID passed to them, I'm not sure we can assume that they are stable across build environments.

It's true that they are stable *per image*, so it would be possible to write the `dangerzone` UID/GID to a file in the outer container filesystem, and read that file from the Python code. But I figure that the "spawn a `chroot`ed process to find out" solution is essentially the same thing: it's reading the UID/GID values from a file in the container filesystem just as well. :)

> Heads up, I have a design document ready, that should explain how the gVisor integration works in Dangerzone, to people who have not seen the code: [8641b66](https://github.com/freedomofpress/dangerzone/commit/8641b66b0db634d1b6b849f9047a93671d7c5a13)
> 
> @EtiennePerot if you feel I have missed something important, or that I've explained something incorrectly, feel free to chime in. In the meantime, I'll take a look at the CI build failures.

Thanks, this is great! Left some comments on the commit.

apyrgio:
> I'm not sure that's actually always the case; it depends on which UID/GID the adduser/addgroup commands use during the image building process. Now that they no longer have an explicit UID/GID passed to them, I'm not sure we can assume that they are stable across build environments.

Hm, I'm ultimately ok with that. Those commands are stable enough, and I expect them to assign UID 1000 to the created user consistently. It may be the case though that Alpine Linux decides to add a built-in user in the container image (looking at you, Ubuntu :roll_eyes: ), and the UID will then become 1001. Even then, our nightly checks will fail loudly and we can fix it.

On the upside, we'll be closer to removing one more capability from the Docker/Podman invocation, and slightly improve readability. So, I'm more inclined to go that way, if you don't have a strong objection.

apyrgio:
I've made some [fixes](https://github.com/freedomofpress/dangerzone/commit/d480ca52f65a84c266d798794fd675774e67f447) to another branch of mine (see [`wip-gvisor-2`](https://github.com/freedomofpress/dangerzone/tree/wip-gvisor-2)), to make the code work for the second phase as well, and make it suitable for our various Linux platforms. I think the Debian Bullseye tests are still failing though....

In any case, tell me what you think. I'm especially curious about the handling of the second phase, since I'm doing some radical stuff there :grimacing:

apyrgio:
I think I've made the PR work in every platform (haven't tested on Windows yet). I've made some [extra fixes](https://github.com/freedomofpress/dangerzone/commit/673e3ce657bc3685616f27a5b7f166b6a5a60986) in my [wip-gvisor-2](https://github.com/freedomofpress/dangerzone/tree/wip-gvisor-2) branch:

* Make PyMuPDF within the container get the correct tesseract data path. It's hacky and I don't like it at all, but thankfully all these will soon be removed.
* Debian Focal/Bullseye were failing, because the Podman version that they ship does not contain the `ptrace()` syscall in the default seccomp policy. I've added a [`seccomp.json`](https://github.com/containers/common/blob/ed2a5b62597ce1c35d939945c787658a3679e2f8/pkg/seccomp/seccomp.json) policy file (taken from the official repo) that includes this syscall. This policy will be used just for these distros.
* I tested this PR on macOS, which uses Docker, and to my surprise, I saw that using an unconfined seccomp policy is not necessary.

EtiennePerot:
Hi, thanks for the review and all the work on the follow-up changes. Will review this weekend

apyrgio:
Oh, one more thing for when you have time. I was under the impression that gVisor was using a more modern approach to `ptrace()` called `systrap`. However, I'm not sure if it's used within the container, or if it silently fallbacks to `ptrace()`. I'm questioning this because some of the issues were resolved once I allowed the `ptrace()` syscall in selected platforms.

apyrgio:
> Oh, one more thing for when you have time. I was under the impression that gVisor was using a more modern approach to `ptrace()` called `systrap`. However, I'm not sure if it's used within the container, or if it silently fallbacks to `ptrace()`. I'm questioning this because some of the issues were resolved once I allowed the `ptrace()` syscall in selected platforms.

I actually checked the above by adding the (very useful) `-e RUNSC_DEBUG=1` argument in our container image. From the logs, I see that it does use `systrap`, so we're good. However, I also see the following:

> I0605 11:43:46.836202      20 filter.go:58] No precompiled program found for config options GOARCH="amd64" Platform="systrap" HostNetwork=false HostNetworkRawSockets=false HostFilesystem=true ProfileEnable=false Instrumentation=false NVProxy=false TPUProxy=false, building seccomp program from scratch. This may slow down container startup.

In practice, I haven't noticed any performance overhead, so I believe we can safely ignore it. Let me know if you have a different opinion.

apyrgio:
In what I hope is one of the last changes I make to this PR, I added one more important commit to my WIP branch (see commit https://github.com/freedomofpress/dangerzone/commit/df815dd608bfc314bd82ec20d72eb1f17922a273, branch [`wip-gvisor-2`](https://github.com/freedomofpress/dangerzone/tree/wip-gvisor-2)).

In this commit, we:
1. Add an unprivileged user (`dangerzone`) in the **outer** container as well. This means that `runsc` runs fully rootless.
2. Remove the need for `su-exec`, since we no longer use `--userns keep-id` for the first stage of the conversion. This also means that we can specify in our OCI config that the UID for the **inner** container will be 1000. Note that the `su-exec` was originally required due to https://github.com/google/gvisor/issues/9918.
3. Remove the need for `--userns keep-id` from the second stage of the conversion as well, so that we can finally close https://github.com/freedomofpress/dangerzone/issues/624 as well.
4. Remove the `CAP_SET_FCAP` Linux capability from the **outer** container, which was necessary when `runsc` was running as root.

---

With all these changes, if we compare our approach with the current situation on the `main` branch, the only difference is that we add a single Linux capabilitiy; `CAP_SYS_CHROOT`. We still drop all others Linux capabilities, we use the default seccomp filter, and run as an unprivileged user within the container. Not only that, be we remove a stale flag that was mapping the host user into the container, so that's another win.

As for the inner container, it has no capabilities, runs as an unprivileged user, has an even stricter seccomp filter, and does not let the command interfere with the Linux kernel directly. And I'm listing just the major wins from the gVisor integration, not all the other ways that gVisor protects us.

I'm personally extremely satisfied with how this turned out, to the point where we can even close these "defense in depth issues" I had opened a long time ago:
* https://github.com/freedomofpress/dangerzone/issues/228
* https://github.com/freedomofpress/dangerzone/issues/225
* https://github.com/freedomofpress/dangerzone/issues/224

Looking forward to your comments @EtiennePerot, and thanks for significantly improving the security of this project :heart:.

EtiennePerot:
Thanks for all the fixes for other distributions! I will post comments on the commits in your own branch.

How would you like this PR to proceed? Should I merge your commits on the branch for this PR? And if so, should I do so now, or once everything is ready?

> Oh, one more thing for when you have time. I was under the impression that gVisor was using a more modern approach to `ptrace()` called `systrap`. However, I'm not sure if it's used within the container, or if it silently fallbacks to `ptrace()`. I'm questioning this because some of the issues were resolved once I allowed the `ptrace()` syscall in selected platforms.
> Also, while reading more on the systrap platform, I realized that it still uses the ptrace() syscall, so this explains why it's still necessary.

As you've noted, this isn't gVisor falling back to the "ptrace" platform. The gVisor Systrap platform uses `ptrace` system calls as part of how it spawns threads and sets them up before they start running untrusted code.

> > I0605 11:43:46.836202 20 filter.go:58] No precompiled program found for config options GOARCH="amd64" Platform="systrap" HostNetwork=false HostNetworkRawSockets=false HostFilesystem=true ProfileEnable=false Instrumentation=false NVProxy=false TPUProxy=false, building seccomp program from scratch. This may slow down container startup.
>
> In practice, I haven't noticed any performance overhead, so I believe we can safely ignore it. Let me know if you have a different opinion.

To save time at container startup, the gVisor binary comes with a set of bundled precompiled seccomp-bpf programs (as cBPF bytecode) for its syscall filters, so that it doesn't have to compile its syscall filters to cBPF at runtime for common configurations. I wrote about this [here if you're interested](https://gvisor.dev/blog/2024/02/01/seccomp/), Ctrl+F "precompilation" is the relevant section.

When started with a configuration for which it doesn't have a precompiled seccomp-bpf program, gVisor will compile its syscall filters to cBPF bytecode at runtime instead, and this warning is there to show that it is doing that. This compilation process should take on the order of ~tens of milliseconds on a modern processor, so at the scale Dangerzone runs, this difference should be imperceptible. That being said, the configuration that Dangerzone uses here should be one of the "common" configurations for which there should be a precompiled seccomp-bpf program, so this is surprising nonetheless. I'll look at why this is happening...

> With all these changes, if we compare our approach with the current situation on the main branch, the only difference is that we add a single Linux capabilitiy; CAP_SYS_CHROOT. We still drop all others Linux capabilities, we use the default seccomp filter, and run as an unprivileged user within the container. Not only that, be we remove a stale flag that was mapping the host user into the container, so that's another win.

This is ingenious, wish I had thought of it. Very nice, and no more dependency on https://github.com/google/gvisor/issues/9918 is needed.

> Thanks for significantly improving the security of this project ❤️.

:saluting_face:

EtiennePerot:
> That being said, the configuration that Dangerzone uses here should be one of the "common" configurations for which there should be a precompiled seccomp-bpf program, so this is surprising nonetheless. I'll look at why this is happening...

Turns out this is because the `runsc` release pushed to `storage.googleapis.com/gvisor/releases` is built without optimizations (missing the `-c opt` option). The build process skips seccomp program precompilation from non-optimized `runsc` builds (to optimize for build speed), hence the warning. I'll change the `runsc` release process to turn on `-c opt` by default to address this.

apyrgio:
> How would you like this PR to proceed? Should I merge your commits on the branch for this PR? And if so, should I do so now, or once everything is ready?

Yeah, I'd prefer if my changes where squashed into your commits, so that our iterations are not shown in our Git history. I propose the following:
1. Add some extra commits in the `wip-gvisor-2` branch, based on your comments. Probably will do some polishing as well in the code, since it was pretty much a WIP effort.
2. Provide a `wip-gvisor-2-squashed` branch, with the exact same code, but with clean Git history.

This way, you can look both at the delta, and the final branch.

> I'll change the runsc release process to turn on -c opt by default to address this.

Nice!

apyrgio:
> Add some extra commits in the [`wip-gvisor-2`](https://github.com/freedomofpress/dangerzone/tree/wip-gvisor-2) branch, based on your comments. Probably will do some polishing as well in the code, since it was pretty much a WIP effort.

Done, I have added a few more commits that address your review comments and polish the code a bit.

> Provide a [`wip-gvisor-2-squashed`](https://github.com/freedomofpress/dangerzone/tree/wip-gvisor-2-squashed) branch, with the exact same code, but with clean Git history.

Done, this branch is equivalent with the `wip-gvisor-2` branch, but I have squashed our iterations into a single commit. In order to make it as small as possible, but still functional, I have factored out some changes in separate commits.

Next steps would be to:
* Take a look at these branches, and incorporate `wip-gvisor-2-squashed` into your own `gvisor` branch, if you agree.
* Make any other changes that you want on top of mine.
* In the meantime, I'll try one more time to run Dangerzone in the all of our supported platforms (x86/ARM macOS, and Windows).
* ???
* ~Profit~ Merge!

EtiennePerot:
Thanks, I have incorporated `wip-gvisor-2-squashed` and re-signed my commit within it. I also incorporated your `CHANGELOG.md` fixes within that commit, since I could not otherwise have signed them (also fixed bug links within that).
I left some comments on your commits but as they are all trivial comment spelling/wording changes, I figured I'd just incorporate them in my commit as well.

apyrgio (approved):

apyrgio:
The changes look good Etienne, thanks a lot. As for the signatures, I'm having a hard time retaining the original ones prior to the merge. For instance, I had to rebase this PR on top of some recently commits on `main`, so your signature has been replaced with mine. Oh well, that's just the way it goes.

I'm waiting for the CI tests to pass one last time, and then I'll merge this PR.

EtiennePerot:
Thank you for all the work on reviewing this and the related fixes!

PR #600, opened 2023-10-13
Allow user to override update checking on Linux

### LINKED ISSUE #596: [0.5.0/Ubuntu 22.04] "Check for updates" setting does not persist
## Steps to reproduce

1. Create an Ubuntu 22.04 QA environment using `qa.py`
2. Run `dangerzone` inside the container
3. Toggle the "Check for updates" checkbox
4. Restart Dangerzone

## Expected behavior

Setting persists.

## Actual behavior

The checkbox is unchecked again on restart.

deeplow:
The original intention of leaving the update checkbox in the hamburger menu was to let non-supported Linux distros (e.g. compiled from source) to check for updates. However, on Linux it ended up being disabled forcefully by default on startup.

This takes into account an overriden update checkbox.

Fixes #596


eloquence (approved):
Can confirm that this fixes the issue (tested via `qa.py`).

PR #605, opened 2023-10-13
Improve the instructions for QA step 10

### LINKED ISSUE #597: Streamline QA process for scenario 10: keeping state after update
Scenario 10 (as of 0.5.0) is not practical to test with our `dev_scripts/qa.py`.

> #### 10. Updating Dangerzone handles external state correctly.
> 
> _(Applies to Linux/Windows/MacOS. For MacOS/Windows, it requires an installer
for the new version)_
> 
> Install the previous version of Dangerzone system-wide. Open the Dangerzone
application and enable some non-default settings. Close the Dangerzone
application and get the container image for that version. For example
> 
> ```
> $ podman images dangerzone.rocks/dangerzone:latest
> REPOSITORY                   TAG         IMAGE ID      CREATED       SIZE
> dangerzone.rocks/dangerzone  latest      <image ID>    <date>        <size>
> ```

The reason for this is that we'd need access to an older dangerzone version but since the env is in a container, we can't really run root easily and install the older version.

## Solution

**We either skip this QA step on Linux platforms or we make it easier to install an older version.**

For example, we could have a dev_script called "install_older_version.sh" with setuid on the container that would download the previous release, install it, run it, changing the state and then install the current version and check if that is done properly.

apyrgio:
Clarify how can a tester install the previous version of Dangerzone in the step 10 of the QA.

Closes #597

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -158,9 +158,20 @@ should point the user to the Qubes notifications in the top-right corner:
 _(Applies to Linux/Windows/MacOS. For MacOS/Windows, it requires an installer
 for the new version)_
 
-Install the previous version of Dangerzone system-wide. Open the Dangerzone
-application and enable some non-default settings. Close the Dangerzone
-application and get the container image for that version. For example
+Install the previous version of Dangerzone system-wide:
+
+* For MacOS/Windows, use the version from the website.
+* For Linux, uninstall Dangerzone from your test environment, and install the
+  previous version using our [installation instructions](INSTALL.md). Also,
+  keep in mind the following:
+  - In order to run commands as root, execute into the container as root with
+    `podman exec -it -u root <container ID> bash`.
+  - If you encounter a GPG error, run the `dirmngr` command to create the
+    necessary directories.
+
+Open the Dangerzone application and enable some non-default settings. Close the
+Dangerzone application and get the container image for that version. For
+example:
 
 ```
 $ podman images dangerzone.rocks/dangerzone:latest
@@ -170,10 +181,11 @@ dangerzone.rocks/dangerzone  latest      <image ID>    <date>        <size>
 
 _(use `docker` on Windows/MacOS)_
 
-Install the new version of Dangerzone system-wide. Open the Dangerzone
-application and make sure that the previously enabled settings still show up.
-Also, ensure that Dangerzone reports that the new image has been installed, and
-verify that it's different from the old one by doing:
+Install the new version of Dangerzone system-wide. For Linux, copy the package
+back into the container. Open the Dangerzone application and make sure that the
+previously enabled settings still show up.  Also, ensure that Dangerzone reports
+that the new image has been installed, and verify that it's different from the
+old one by doing:
 
 ```
 $ podman images dangerzone.rocks/dangerzone:latest
--- a/dev_scripts/qa.py
+++ b/dev_scripts/qa.py
@@ -160,9 +160,20 @@ should point the user to the Qubes notifications in the top-right corner:
 _(Applies to Linux/Windows/MacOS. For MacOS/Windows, it requires an installer
 for the new version)_
 
-Install the previous version of Dangerzone system-wide. Open the Dangerzone
-application and enable some non-default settings. Close the Dangerzone
-application and get the container image for that version. For example
+Install the previous version of Dangerzone system-wide:
+
+* For MacOS/Windows, use the version from the website.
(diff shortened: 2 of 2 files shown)


apyrgio:
@eloquence You might want to take a look at this.

RELEASE.md
```diff
@@ -158,9 +158,16 @@ should point the user to the Qubes notifications in the top-right corner:
...
 
-Install the previous version of Dangerzone system-wide. Open the Dangerzone
-application and enable some non-default settings. Close the Dangerzone
-application and get the container image for that version. For example
+Install the previous version of Dangerzone system-wide:
+
+* For MacOS/Windows, use the version from the website.
+* For Linux, install the previous version using our [installation instructions](INSTALL.md).
+  In order to run commands as root, execute into the container as root with
+  `podman -it -u root <container ID> bash`.
```
eloquence, line 166: Correction: `podman exec` (repeated below)

dev_scripts/qa.py
```diff
@@ -160,9 +160,16 @@
...
 
-Install the previous version of Dangerzone system-wide. Open the Dangerzone
-application and enable some non-default settings. Close the Dangerzone
-application and get the container image for that version. For example
+Install the previous version of Dangerzone system-wide:
+
+* For MacOS/Windows, use the version from the website.
+* For Linux, install the previous version using our [installation instructions](INSTALL.md).
+  In order to run commands as root, execute into the container as root with
+  `podman -it -u root <container ID> bash`.
```
eloquence, line 168: We're orienting the tester to log into the test environment using `qa.py` here during all the previous steps:
https://github.com/freedomofpress/dangerzone/blob/f02597aa4fe2913513fc252b2a133a8e6ec77111/dev_scripts/qa.py#L648

For this scenario, I think it makes sense to spin up a new container, so no traces of the test installation remain. Perhaps we can make that more explicit at the beginning of the scenario.

Note: In order to be able to follow `INSTALL.md` on Ubuntu 22.04, I had to run `dirmngr` in the background, otherwise the `gpg` keyserver command would error out with "No dirmngr".

eloquence (changes requested):
Thanks Alex! Left a couple of comments but otherwise looks great.

dev_scripts/qa.py
apyrgio: > Note: In order to be able to follow INSTALL.md on Ubuntu 22.04, I had to run dirmngr in the background, otherwise the gpg keyserver command would error out with "No dirmngr".

I have this error too actually, and I always wondered what's the case there. I think it's because we don't install `dirmngr` in the container image, due to `--no-install-recommends`, but I'll need to take a look.

apyrgio: Actually no, the `dirmngr` package is installed, but it needs to run once before it creates some dirs. This looks to me as a container limitation, but I can add a note to the tester.

eloquence: > That's why I figured that uninstalling the Dangerzone package from the end-user environment is a reasonable compromise

As a tester, I'm still not entirely sure what the instructions are asking me to do in this scenario.

- Uninstall Dangerzone from the container in which it was installed using `qa.py` in the previous steps, install a previous version, and then upgrade? (That is what I did. I did have to make sure to keep the container opened via `qa.py` running, since the `qa.py` step itself uses `run`.)
- Prepare a new container using `env.py` and manually install Dangerzone in it, then upgrade it?
- Build Dangerzone on the _host_ environment without the use of containers? Is that what you mean with system-wide?
- Something else?

apyrgio: Sigh, sorry for the confusion. I'm referring to the first bullet (uninstall -> install previous -> upgrade). Would it make it more clear if we changed the text to:

> * For Linux, uninstall Dangerzone via `apt-get remove dangerzone` / `dnf uninstall dangerzone`, and install the previous version using our [installation instructions](INSTALL.md) [...]

eloquence: No worries, this is a complex process and I'm happy to be a guinea pig for the docs :). I'd only maybe change it to "uninstall Dangerzone from your test environment", but otherwise LGTM.

apyrgio: Great, I've incorporated this as well in 60cb83c65aedb9c267782972124f5cd6e8599124, and restructured the text a bit for clarity.

apyrgio:
Hey, quick update here. I've squashed the commits, addressed Erik's comments, and fixed a lint issue. Feel free to merge this when you have time.

eloquence (approved):
LGTM!

PR #608, opened 2023-10-17
Minor tweaks to Qubes build docs

eloquence:
- `keyring` command will only work if `python3-keyring` is installed
- fix `cp` command (`qubes` directory not included in prior command)


apyrgio (approved):
Good catch!

PR #615, opened 2023-10-23
Add instructions for adding release tag

deeplow:
Instructions only stated how to verify the release tag bug not how to make it.

A release tag should be:
  - **signed**: for verifying integrity
  - **annotated**: to make it findable by `git describe`


RELEASE.md
```diff
@@ -190,9 +190,15 @@ Before making a release, all of these should be complete:
...
 - [ ] CHANGELOG.md should be updated to include a list of all major changes since the last release
-- [ ] There must be a PGP-signed git tag for the version, e.g. for dangerzone 0.1.0, the tag must be `v0.1.0`
+- [ ] Create a PGP-signed git tag for the version, e.g. for dangerzone `v0.1.0`, the is created with:
 
-Before making a release, verify the release git tag:
+  ```
+  git tag -a -s v0.1.0
+  git push origin v0.1.0
+  ```
+  **Note**: release candidates can be suffixed by `-rcX` but should not be annotated. You would create one like this: `git tag -s v0.1.0-rc1`.
```
eloquence, line 199: Why the "should not be annotated" qualifier? AFAIK the signature is a form of annotation, so if you `-s` you'll create an annotated tag.

apyrgio: Correct, from the [Git manpage](https://git-scm.com/docs/git-tag#_description):

> Tag objects (created with -a, -s, or -u) are called "annotated" tags

So, we can just remove the `-a` flag and the note, and merge this. Other than that looks good to me.

apyrgio: @deeplow: ping about this. Are you ok with the proposed changed?

apyrgio: I've approved the PR in the meantime.

apyrgio: Sorry for the delay. Yes, I meant only the second sentence in the note. Feel free to merge, thanks!

Issue #574, opened 2023-10-04
(server) Qubes: Conversion of medium-sized documents fills up /tmp


apyrgio:
# Problem

The Qubes converter currently handles `/tmp` in an inefficient way, leading to ENOSPC situations, when a document has more than 160 US letter pages.

# Background

We have a known problem regarding the way we store pixel data in `tmpfs` (see #526), which becomes even worse in Qubes. The reason is that Qubes mounts `tmpfs` under `/tmp` this way:

```
tmpfs on /tmp type tmpfs (rw,size=1048576k,nr_inodes=35349,inode64)
```

The size of the `/tmp` dir in app and disposable qubes is therefore max 1GiB. At the same time, `pdftoppm` converts the pages at 150 DPI, which means each RGB page requires ~6.2MiB. Therefore, we can store at most ~165 pages in `/tmp`.

This limitation manifests both at the server-side, and at the client side. In the server-side, `pdftoppm` writes PPM files under `/tmp/page*`. We have a callback that writes the final RGB buffer to stdout and then removes the original PPM file, but these two processes are not synchronized, especially if the client-side does not catch up:

https://github.com/freedomofpress/dangerzone/blob/c4fdebc80d460384a3c552a09f3cae953a293c73/dangerzone/conversion/doc_to_pixels.py#L336-L342

In the client-side, we immediately write the received pixel data under `/tmp/dangerzone`, since this is where the pixels-to-pdf convert expects to find them:

https://github.com/freedomofpress/dangerzone/blob/c4fdebc80d460384a3c552a09f3cae953a293c73/dangerzone/isolation_provider/qubes.py#L132-L144

# Solution

We don't have a solid solution yet. If we fix the client-side issue, by converting the RGB files to PNG immediately, then the disposable qube can still run out of space, if the client-side does not catch up (e.g., due to the conversion). We need a solution that fixes both sides.

apyrgio:
This has been resolved by fixing #443.

PR #619, opened 2023-11-02
Solve most congestions issues in /tmp (client & Server)

### LINKED ISSUE #616: Fix /tmp Filling Up Bottlenecks (server + client)
The temporary (`/tmp`) space limitations of Qubes revealed some bottlenecks in the current architecture of Dangerzone, which is manifest themselves in large files (150+ pages). The containers version isn't as affected sync the `/tmp` file system has more space, but addressing this issue will improve the situation overall.

## The problems
As far as I can tell, there are two main issues:
  - [ ] https://github.com/freedomofpress/dangerzone/issues/526
     **Lack of space in client**: just a consequence of us waiting on all the RGB files to finally start reconstructing the file
  - [ ] https://github.com/freedomofpress/dangerzone/issues/574
   **Lack of space in server**: even though the RGB files are deleted as soon as they are sent, the reality is that the RGB files are produced much faster than the client can consume them

I tried to illustrate some of this issues in a diagram: 

dangerzone-bottlenecks drawio(4)

deeplow:
In more resourced limited systems (such as Qubes OS) converting 1000+ page documents would not be possible due to several bottlenecks.

This PR addresses several of these bottlenecks (see commits for details).

Fixes #616 #625 #574

## Regressions introduced
The containers version is slightly slower, stopping at the end of each batch for a brief moment. I don't knot yet what's causing the wait, but this will be solved with #443, once containers stream pages.

## Remaining bottlenecks
Currently, converting a 1299 page document in Qubes still fails. This is due to the compression step (the reason why this still a draft). The compression creates additional files in `/tmp` which fill it up.

### DIFF
--- a/install/linux/dangerzone.spec
+++ b/install/linux/dangerzone.spec
@@ -72,6 +72,7 @@ BuildRequires:  python3-devel
 %if 0%{?_qubes}
 # Qubes-only requirements
 Requires:       python3-magic
+Requires:       python3-pillow
 Requires:       libreoffice
 Requires:       tesseract
 # Explicitly require every tesseract model:
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,6 +1,6 @@
 [DEFAULT]
 Package3: dangerzone
-Depends3: podman, python3, python3-pyside2.qtcore, python3-pyside2.qtgui, python3-pyside2.qtwidgets, python3-pyside2.qtsvg, python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-markdown, python3-packaging
+Depends3: podman, python3, python3-pyside2.qtcore, python3-pyside2.qtgui, python3-pyside2.qtwidgets, python3-pyside2.qtsvg, python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-markdown, python3-packaging, python3-pillow
 Build-Depends: dh-python, python3, python3-setuptools, python3-stdeb
 Suite: bionic
 X-Python3-Version: >= 3.6


dangerzone/isolation_provider/container.py
```diff
@@ -304,11 +305,33 @@ def _convert_with_tmpdirs(
...
+                height = int(f.read().strip())
+            with open(rgb_filename, "rb") as rgb_f:
+                untrusted_pixels = rgb_f.read()
+                self.convert_pixels_to_png(
+                    str(pixel_dir), page, width, height, rgb_data=untrusted_pixels
+                )
+
+            os.remove(rgb_filename)
+            os.remove(width_filename)
+            os.remove(height_filename)
```
deeplow, line 328: All of this should be move to after checking if there was a non-zero exit code.

apyrgio:
I haven't finished reviewing the code, but I was thinking that we can dramatically improve the server-side batching part with [`PyMuPDF`](https://github.com/pymupdf/PyMuPDF).

Consider this snippet:

```python
import sys, fitz  # import the bindings

fname = sys.argv[1]  # get filename from command line
doc = fitz.open(fname)  # open document

for page in doc:  # iterate through the pages
    pix = page.get_pixmap(dpi=150)  # render page to an RGB image
    rgb_buf = pix.samples_mv  # Get a memory view over the pixel data
    with open("mu-%i.rgb" % page.number, "wb+") as f:
        f.write(rgb_buf)
    print(f"Page {page.number}: W: {pix.width}, H: {pix.height}")
```

This snippet:

1. Opens a document and reads every page
2. Renders each page as an RGB image **on the fly**, and immediately has access to the width and height.
    * Rendering the page to RGB on the fly means that we can ditch the unnecessary PPM -> RGB conversion step, as well as the convoluted callback logic that come with `pdftoppm`
3. Writes the pixel data to a file (but can also write them to a socket).

For the container case, we can simply iterate all the pages and write the data to the filesystem. For the Qubes case though, we can write the data immediately to the stdout, without buffering them locally. This means that we don't need any batch logic, and Unix pipes become the means of congestion control.

Also, another improvement that this snippet brings is that it can RGB files 2x faster than `pdftoppm` creates PPM file. And this does not even account the time it takes to convert PPM -> RGB.

Alas, there's a caveat: PyMuPDF is an AGPL project, and Artifex takes its GPL violations [seriously](https://opensource.stackexchange.com/a/6904). So, what do we have to concede here to make our code faster, simpler, and more secure? My understanding is that we need to license the `conversion` module under AGPL. This does not affect the application as a whole, because the code runs either in a separate Docker container, or disposable Qube.

Thoughts on this?

dangerzone/isolation_provider/base.py
```diff
@@ -101,74 +104,26 @@ def sanitize_conversion_str(self, untrusted_conversion_str: str) -> str:
...
-#         ):
-#             return False, f"Page {i} has invalid geometry"
-
-#         # Make sure the RGB file is the correct size
-#         if os.path.getsize(f"{common.pixel_dir.name}/page-{i}.rgb") != w * h * 3:
-#             return False, f"Page {i} has an invalid RGB file size"
-
-#     return True, True
+        try:
+            Image.open(ppm_data).save(png_path, "PNG")
```
apyrgio, line 127: I had tackled something similar in my `2023-10-test-enospc` branch. You don't need to go through the PPM -> PNG conversion, nor create a `bytesIO` object. You can convert it from the existing buffer with:

```
image = Image.frombuffer("RGB", (width, height), untrusted_pixels, "raw", "RGB", 0, 1)
```

dangerzone/isolation_provider/container.py
apyrgio: Just to make sure why this code is necessary in the first place:

1. The **current** containers version, which does not stream the data to the host yet, has to write all the pixels in the temporary directory. This is the worst case scenario, and by converting them to PNG after the fact does not help much.
2. The **streaming** (post-#443) containers version will piggyback on the Qubes code that converts the RGB images to PNG on the fly, which has already been added in this PR.

Still, the code seems to be necessary because it makes `pixels-to-pdf` accept only PNG images, not RGB, thereby simplifying its interface. Is this your rationale as well?

deeplow:
> I haven't finished reviewing the code, but I was thinking that we can dramatically improve the server-side batching part with [PyMuPDF](https://github.com/pymupdf/PyMuPDF).

Thanks for the feedback. That's certainly interesting proposal. I did consider some of those PDF parsing libraries, but ultimately though it wouldn't add much benefit due to their disadvantages:
 -  **we're adding another library**
 - **The PDF rendering engine change** - poppler isn't perfect, but changing it may change some things (I think). 
 - **The potential licensing issues** - I'm not sure the disposable qube counts the same as a docker image. It would probably be so if we shipped a template image, which we don't.

But the speed increase and python-native bindings are certainly. Regarding speed, the current performance seems fast enough to be acceptable and there are slower steps in the pipeline (e.g. OCR) so I don't think it's too critical at the moment. 

I am unsure if the benefits outweigh the risks. We could have a sync discussion if you have the availability.

dangerzone/isolation_provider/base.py
deeplow: Ah, nice! I though there would be some golden nugget like this, but I had difficulties finding that branch of yours. Thanks! I'll try and add it.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -112,27 +84,39 @@ async def convert(
...
-        await self.run_command(
-            args,
-            error_message="Merging pages into a single PDF failed",
-            timeout_message=(
-                "Error merging pages into a single PDF, pdfunite timed out after"
-                f" {timeout} seconds"
-            ),
-            timeout=timeout,
-        )
+        for first_page, last_page in batch_iterator(num_pages):
```
apyrgio, line 99: Question: what do we gain by batching the `pdfunite` operations?

At this point, we already have all the pages as PDF files on disk, and that is the main space hog, right? I get that with a single `pdfunite` call we need double the space, but that's a requirement for the conversion step as well, which takes the original PDF and creates a compressed one, and they both co-exist in the disk.

I have another suggestion here: using PyPDF2 instead, and perform compression online:

```python
from pypdf import PdfReader, PdfWriter

reader = PdfReader("example.pdf")
writer = PdfWriter()

for page in reader.pages:
    writer.add_page(page)

for page in writer.pages:
    # ⚠️ This has to be done on the writer, not the reader!
    page.compress_content_streams()  # This is CPU intensive!

with open("out.pdf", "wb") as f:
    writer.write(f)
 ```
 
 _(snippet taken from [the docs](https://pypdf.readthedocs.io/en/latest/user/file-size.html#lossless-compression))_
 
With PyPDF2, we can open each PDF, compress it, and immediately create the output PDF, without `pdfunite` and `ps2pdf`. We can even delete the underlying PDFs as we process them. The good thing about `PyPDF2` is that it's available in Fedora repos as well, so it will work on Qubes.

apyrgio (commented):
As you see, I have proposed some radical detours from the batching implementation, which use native Python libraries instead. I think at the end of the day, this will lead to a simpler, faster, and more secure implementation, which is worth doing now that we are at the start of the release cycle. I'd like to hear your opinion as well though.

dangerzone/conversion/pixels_to_pdf.py
deeplow: > Question: what do we gain by batching the pdfunite operations?

It's [in the commit message](https://github.com/freedomofpress/dangerzone/pull/619/commits/f7190e3876ce3ab33fb0db9b6e94dc9ff872a531). So it's not size-related but rather ulimit related. We could change the ulimit, but since I had already done the batch logic, I figured it could be reused here.

dangerzone/isolation_provider/container.py
deeplow: > Still, the code seems to be necessary because it makes pixels-to-pdf accept only PNG images, not RGB, thereby simplifying its interface. Is this your rationale as well?

Yes, that's the logic here. And it's not worth leaving the pixels-to-pdf code more complex to make #412 easier.

dangerzone/conversion/pixels_to_pdf.py
apyrgio: Oh snap. I didn't expect that Qubes had such a low ulimit (nor that `pdfunite` opens the files at the same time, jeez...).

deeplow:
> I haven't finished reviewing the code, but I was thinking that we can dramatically improve the server-side batching part with [PyMuPDF](https://github.com/pymupdf/PyMuPDF).

I have now stalled progress on this PR, since I'll be evaluating how well PyMuPDF works for us, especially regarding the need to change the code license (or at least part of it).

PR #622, opened 2023-11-16
Use PyMupPDF to solve most congestions issues in /tmp (client & Server) 

deeplow:
Alternative to #619 but with PyMuPDF. This implements changes in the doc to pixels as well as the pixels to pdf conversion.

Missing
- [ ] assess the impact of PyMuPDF (performance, security, PDF rendering impact)
- [ ] change Dangerzone license due to PyMuPDF being AGPL

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ ARG H2ORESTART_CHECKSUM=5db816a1e57b510456633f55e693cb5ef3675ef8b35df4f31c90ab9d
 RUN apk --no-cache -U upgrade && \
     apk --no-cache add \
     ghostscript \
-    graphicsmagick \
     libreoffice \
     openjdk8 \
     poppler-utils \
@@ -16,6 +15,10 @@ RUN apk --no-cache -U upgrade && \
     py3-magic \
     tesseract-ocr \
     font-noto-cjk
+
+RUN apk add g++ gcc make python3-dev py3-pip
+RUN pip install --upgrade PyMuPDF # FIXME freeze w/ hashes
+RUN apk del g++ gcc make python3-dev py3-pip
 
 # Download the trained models from the latest GitHub release of Tesseract, and
 # store them under /usr/share/tessdata. This is basically what distro packages
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -258,8 +258,6 @@ class DocumentToPixels(DangerzoneConverter):
             raise errors.MaxPagesException()
         await self.write_page_count(doc.page_count)
 
-        # Get a more precise timeout, based on the number of pages
-        timeout = self.calculate_timeout(size, doc.page_count)
         percentage_per_page = 45.0 / doc.page_count
         page_base = "/tmp/page"
         for page in doc:
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -147,14 +147,14 @@ class DocumentToPixels(DangerzoneConverter):
                 "type": "libreoffice",
             },
             # .jpg
-            "image/jpeg": {"type": "convert"},
+            "image/jpeg": {"type": None},
             # .gif
-            "image/gif": {"type": "convert"},
+            "image/gif": {"type": None},
             # .png
-            "image/png": {"type": "convert"},
+            "image/png": {"type": None},
             # .tif
-            "image/tiff": {"type": "convert"},
-            "image/x-tiff": {"type": "convert"},
+            "image/tiff": {"type": None},
+            "image/x-tiff": {"type": None},
         }
 
         # Detect MIME type
@@ -183,7 +183,7 @@ class DocumentToPixels(DangerzoneConverter):
         # Convert input document to PDF
         conversion = conversions[mime_type]
         if conversion["type"] is None:
-            pdf_filename = "/tmp/input_file"
+            doc = fitz.open("/tmp/input_file", filetype=mime_type)
         elif conversion["type"] == "libreoffice":
             libreoffice_ext = conversion.get("libreoffice_ext", None)
             # Disable conversion for HWP/HWPX on specific platforms. See:
@@ -228,24 +228,7 @@ class DocumentToPixels(DangerzoneConverter):
             #     https://github.com/freedomofpress/dangerzone/issues/494
             if not os.path.exists(pdf_filename):
                 raise errors.LibreofficeFailure()
-        elif conversion["type"] == "convert":
-            self.update_progress("Converting to PDF using GraphicsMagick")
-            args = [
-                "gm",
-                "convert",
-                "/tmp/input_file",
-                "/tmp/input_file.pdf",
-            ]
-            await self.run_command(
-                args,
-                error_message="Conversion to PDF with GraphicsMagick failed",
-                timeout_message=(
-                    "Error converting document to PDF, GraphicsMagick timed out after"
-                    f" {timeout} seconds"
-                ),
-                timeout=timeout,
-            )
-            pdf_filename = "/tmp/input_file.pdf"
+            doc = fitz.open(pdf_filename)
         else:
             raise errors.InvalidGMConversion(
                 f"Invalid conversion type {conversion['type']} for MIME type {mime_type}"
@@ -253,7 +236,6 @@ class DocumentToPixels(DangerzoneConverter):
         self.percentage += 3
 
         # Obtain number of pages
-        doc = fitz.open(pdf_filename)
         if doc.page_count > errors.MAX_PAGES:
             raise errors.MaxPagesException()
         await self.write_page_count(doc.page_count)
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -16,6 +16,7 @@ import shutil
 import sys
 from typing import Dict, List, Optional
 
+import fitz
 import magic
 
 from . import errors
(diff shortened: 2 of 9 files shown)


dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
             # The first few operations happen on a per-page basis.
             page_size = os.path.getsize(filename_base + ".rgb") / 1024**2
             total_size += page_size
             timeout = self.calculate_timeout(page_size, 1)
 
+            self.update_progress(
+                f"Converting page {page_num}/{num_pages} from pixels to PDF"
+            )
+            pixmap = fitz.Pixmap(
+                fitz.Colorspace(fitz.CS_RGB), width, height, untrusted_rgb_data, False
```
deeplow, line 57: Let's make the `untrusted_rgb_data` explicitly as the `samples` [argument](https://pymupdf.readthedocs.io/en/latest/pixmap.html#id7) (rgb data). This is security-critical since if it interprets this data as any other format, it is not ensuring the security properties of Dangerzone.

deeplow: We should also add a test to ensure that this does not accept any other format. For example, we could pass it a png and ensure that fails.

Dockerfile
```diff
@@ -7,7 +7,6 @@ ARG H2ORESTART_CHECKSUM=5db816a1e57b510456633f55e693cb5ef3675ef8b35df4f31c90ab9d
 RUN apk --no-cache -U upgrade && \
     apk --no-cache add \
     ghostscript \
-    graphicsmagick \
```
apyrgio, line 10: Do we remove only a single package? For instance, are `poppler` libraries still required? Are CJK fonts? BTW, it would be interesting measuring the impact of PyMuPDF in the resulting container image.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
-                    ),
-                    timeout=timeout,
-                )
+                try:
+                    safe_doc.insert_file(pixmap)
+                except:  # remove after PyMuPDF 22 (Document.insert_file available)
+                    blank_page = safe_doc.new_page(width=width, height=height)
+                    blank_page.insert_image(
+                        fitz.Rect(0, 0, width, height), pixmap=pixmap
+                    )
```
apyrgio, line 80: Similarly, I think we can remove this, given that PyMuPDF 1.23 exists in PyPI.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
-                width = f.read().strip()
+                width = int(f.read().strip())
             with open(height_filename) as f:
-                height = f.read().strip()
-
+                height = int(f.read().strip())
+            with open(rgb_filename, "rb") as rgb_f:
+                untrusted_rgb_data = rgb_f.read()
             # The first few operations happen on a per-page basis.
             page_size = os.path.getsize(filename_base + ".rgb") / 1024**2
```
apyrgio, line 49: The file size is now available from `len(untrusted_rgb_data)`.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
-
+                try:
+                    ocr_pdf_bytes = pixmap.pdfocr_tobytes(
+                        compress=True,
+                        language=ocr_lang,
+                        tessdata=get_tessdata_dir(),
+                    )
+                except:  # containers' verison has old API without tessdata
+                    ocr_pdf_bytes = pixmap.pdfocr_tobytes(
+                        compress=True,
```
apyrgio, line 68: How does this `compress=True` argument work with `deflate_images=True` later on? Is it redundant, can we forego one or the other?

pyproject.toml
```diff
@@ -21,6 +21,7 @@ pyxdg = {version = "*", platform = "linux"}
 requests = "*"
 markdown = "*"
 packaging = "*"
+pymupdf = "^1.23.6"
```
apyrgio, line 24: This should be necessary only when using PyMuPDF on the host, which is not the case for us, right?

apyrgio:
While running the tests, I stumbled on another issue:

```
ImportError while loading conftest '/home/user/dangerzone/tests/conftest.py'.
tests/__init__.py:8: in <module>
    from dangerzone.document import SAFE_EXTENSION
dangerzone/__init__.py:16: in <module>
    from .gui import gui_main as main
dangerzone/gui/__init__.py:28: in <module>
    from ..isolation_provider.qubes import Qubes, is_qubes_native_conversion
dangerzone/isolation_provider/qubes.py:15: in <module>
    from ..conversion.pixels_to_pdf import PixelsToPDF
dangerzone/conversion/pixels_to_pdf.py:16: in <module>
    import fitz
E   ModuleNotFoundError: No module named 'fitz'
```

We should conditionally import the Qubes isolation provider, so that we don't always import the `conversion` module, and thus `fitz`. This will also help with any other Python dependency that exists in the container, but not in the host.

dangerzone/conversion/pixels_to_pdf.py
apyrgio: Well, I guess we should pass those as non-keyword arguments. They way they are defined is a bit quirky after all:

https://github.com/pymupdf/PyMuPDF/blob/5a000bc8bda0293de74b00b33b3610f94cc68b94/fitz/fitz.i#L7761

Dockerfile
```diff
@@ -17,6 +16,10 @@ RUN apk --no-cache -U upgrade && \
     tesseract-ocr \
     font-noto-cjk
 
+RUN apk add g++ gcc make python3-dev py3-pip
+RUN pip install --upgrade PyMuPDF # FIXME freeze w/ hashes
+RUN apk del g++ gcc make python3-dev py3-pip
+
```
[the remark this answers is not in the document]
apyrgio: > Also, I added your multi-stage build Dockerfile in https://github.com/freedomofpress/dangerzone/commit/fecf02265ea900155a951cb40a9b5a250827607b to speedup the builds.

I'm semi-satisfied with the way the multi-stage build panned out, but I guess we'll be fine for now.

> Actually that was an old file that I uploaded, the corrected version was committed with https://github.com/freedomofpress/dangerzone/commit/875e0e94bc3edf5d3f5bd53ccef79528457e560d.

Looks good.  I'm not super fond of these lines:

```python
if platform.system() in ["Darwin", "Windows"]:
    CONTAINER_RUNTIME = "docker"
elif platform.system() == "Linux":
    CONTAINER_RUNTIME = "podman"
```

but we'll remove them once we can install PyMuPDF in the host.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
-                    timeout=timeout,
-                )
-
+                try:
+                    ocr_pdf_bytes = pixmap.pdfocr_tobytes(
+                        compress=True,
+                        language=ocr_lang,
+                        tessdata=get_tessdata_dir(),
+                    )
+                except:  # containers' verison has old API without tessdata
```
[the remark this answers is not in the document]
apyrgio: I agree. Thanks for the fix.

install/common/build-image.py
```diff
@@ -0,0 +1,86 @@
...
+    with open("share/image-id.txt", "w") as f:
+        f.write(image_id)
+
+
+def export_container_pip_dependencies():
+    container_requirements_txt = subprocess.check_output(
+        ["poetry", "export", "--only", "container"], universal_newlines=True
+    )
+    # XXX Export container dependencies and exclude pymupdfb since it is not needed in container
+    # in the container the --require-hashes the fact that pymupdfb is not needed
```
apyrgio, line 79: Something got jumbled here, I'm afraid :-/.

dangerzone/conversion/pixels_to_pdf.py
```diff
@@ -27,137 +29,72 @@ async def convert(
...
+                height = int(f.read().strip())
+            with open(rgb_filename, "rb") as rgb_f:
+                untrusted_rgb_data = rgb_f.read()
             # The first few operations happen on a per-page basis.
             page_size = os.path.getsize(filename_base + ".rgb") / 1024**2
             total_size += page_size
             timeout = self.calculate_timeout(page_size, 1)
 
+            self.update_progress(
+                f"Converting page {page_num}/{num_pages} from pixels to PDF"
```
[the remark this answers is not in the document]
apyrgio: I agree, it's minor.

pyproject.toml
apyrgio: Hehe, here's to adding it back again soon :clinking_glasses:. Accepting it for now.

apyrgio:
>   @apyrgio not sure I follow your exact suggestion. Do you mean importing it closer to where the code is called?
Another option would be conditionally import fitz.

I don't have a strong preference, since on-host conversion will ultimately solve this. Whatever requires the least amount of changes to revert, would be my suggestion. And yes, , my original thought was moving the import statement closer to where the Qubes isolation provider is used.

dangerzone/conversion/doc_to_pixels.py
```diff
@@ -182,7 +183,7 @@ async def convert(self) -> None:
         # Convert input document to PDF
         conversion = conversions[mime_type]
         if conversion["type"] is None:
-            pdf_filename = "/tmp/input_file"
+            doc = fitz.open("/tmp/input_file", filetype=mime_type)
```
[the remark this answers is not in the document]
apyrgio: Nice, thanks :-)

dangerzone/conversion/pixels_to_pdf.py
apyrgio: That's amazing! The minimal time overhead (in this case, 100ms per page) and the size savings in contrast to `ps2pdf` are very promising. Regarding my original question, should we disable `deflate_images` once compression is enabled for OCR?

apyrgio: You know what, all I think is that if `deflate_images=True` does not pose an extra time overhead, and the bytes are exactly the same, this would mean that the `mupdf` code auto-detects that the content stream is in DEFLATE format, and does not redo the compression. Looking into the code though, this does not seem to be the case:

https://github.com/ArtifexSoftware/mupdf/blob/master/source/pdf/pdf-write.c#L2133C2-L2144C2

On the other hand, I tried to reproduce this by doing a test on my own. I OCRed a page with `compress=True`, and then saved the pdf with `deflate_images` either set to True or False. The result does have the exact same number of bytes, so your measurements are indeed correct!

_(The SHA-256 hash is different, but that's only due to a different ID value in the PDF trailer. Other than that, the content streams are byte-to-byte equal)_

---

Worst-case scenario, turns out we actually do double conversion, but its overhead is minor compared to the overhead of the OCR. I can't prove this though, so here's what we can do here. Let's add a note and say that we enable both flags, because we have measured their impact and we have not seen a noticeable overhead. Let's even link to this discussion here, for future reference. This will simplify the code and we will be able to refer to this decision later on. Are you ok with that?

apyrgio (approved):
This is amazing work. I'm stoked for this integration! Feel free to merge.

PR #627, opened 2023-11-27
Containers Page Streaming based on PyMuPDF

### LINKED ISSUE #443: Containers: have progress streamed instead of via mounted volumes (and deprecate doc_to_pixels_qubes_wrapper.py)

deeplow:
Fixes #443, #574. Depends on #622.

Adds initial support for page streaming in containers. This is a critical move to homogenize the isolation providers code as it makes Qubes and Containers share a lot of it.


dangerzone/isolation_provider/base.py
```diff
@@ -102,6 +131,113 @@ def sanitize_conversion_str(self, untrusted_conversion_str: str) -> str:
         return armor_start + conversion_string + armor_end
 
 
+class ProcessBasedIsolationProvider(IsolationProvider):
```
apyrgio, line 134: I'd suggest merging `ProcessBasedIsolationProvider` into `IsolationProvider`, for the following reasons:

1. We don't have another type of isolation provider, so we don't benefit from the extra level of abstraction, and we hinder code-readability.
2. We can make the `Dummy` provider run closer to the way the real isolation providers will run.

dangerzone/conversion/doc_to_pixels.py
```diff
@@ -146,14 +138,14 @@ async def convert(self) -> None:
                 "type": "libreoffice",
             },
             # .jpg
-            "image/jpeg": {"type": "convert"},
+            "image/jpeg": {"type": None},
```
apyrgio, line 141: Let's not forget to create an issue for the extra file formats that PyMuPDF supports, and adding those here.

apyrgio (changes requested):
I've made a first pass of the code, and I have some minor comments. My main concern is the non-blocking read implementation, and how it will work on Windows. I'll help with figuring this out, along with streaming support on the second stage of the conversion (reserved for a next PR).

dangerzone/isolation_provider/base.py
```diff
@@ -18,11 +23,35 @@
 PIXELS_TO_PDF_LOG_END = "----- PIXELS TO PDF LOG END -----"
 
 
+def read_bytes(f: IO[bytes], size: int, timeout: float, exact: bool = True) -> bytes:
+    """Read bytes from a file-like object."""
+    buf = nonblocking_read(f, size, timeout)
+    if exact and len(buf) != size:
+        raise errors.InterruptedConversion
+    return buf
```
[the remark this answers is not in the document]
apyrgio: And that's the reason why it doesn't work:

> Note that on Windows, it only works for sockets; on other operating systems, it also works for other file types (in particular, on Unix, it works on pipes).

_(taken from the [`select`](https://docs.python.org/3/library/select.html#module-select) module)_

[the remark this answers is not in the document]
apyrgio: I have done some write-up on this subject actually: https://github.com/freedomofpress/dangerzone/issues/632

dangerzone/isolation_provider/base.py
apyrgio: We decided not to go with the UnsafeConverter yet, until it can run in all platforms.

apyrgio:
I did a second full review of the PR. Things look pretty good! I only have some minor comments:

1. Once we do our 0.6.0 release, we should inform users that the `--disable-timeouts` CLI command has been removed, since this is an API change.
2. Good thinking about testing what happens if multiple docs are converted at the same time. However, I think some problems still remain. The `percentage` and `progress_callback` attributes are instance-level, and I bet they will create issues with parallel conversions. I think in the long run, it's best to have separate converter instances per document, because they would be more easier to reason and work with. In any case, we can still keep this change and re-evaluate in the future.
3. We need an error message for a bad `conmon` version, before merging this PR. I can add this tomorrow, if you don't have time.
4. I suggest squashing commits and fixing commit messages at this point, since we are both in sync about the current state of the branch.

Finally, I have pushed some minor changes as fixups.

apyrgio:
> We need an error message for a bad conmon version, before merging this PR. I can add this tomorrow, if you don't have time.

Damn, unfortunately we don't have a nice way to find it out:

```
$ apt-cache policy conmon
conmon:
  Installed: 2.0.25+ds1-1.1+deb11u1
  Candidate: 2.0.25+ds1-1.1+deb11u1
  Version table:
 *** 2.0.25+ds1-1.1+deb11u1 100
        100 /var/lib/dpkg/status
$ conmon --version
conmon version 2.0.25
commit: unknown
```

We can't look at the `conmon --version`, because it's the same as the faulty one. At the same time, relying on `apt-cache` is a bit brittle, so I would advise against it.

apyrgio:
> Is the brittle aspect due to it not being super trivial parsing.

I'm afraid not. The reason is that we would need to future-proof our checks against newest `conmon` package patches. For instance, what if a `2.0.25+ds1-1.2` version appears? Also, what will be the naming that Ubuntu chooses for its own package?

Debian versioning is not just lexicographical ordering, so I would be hesitant to grep for strings and then use weird logic to compare them.

apyrgio:
Right, it's the false positive that I think it may bite us. It's not prohibitively difficult, but then again, it's just for Ubuntu Jammy users building from source. I think it's fine adding a line in `BUILD.md` in that case.

apyrgio:
> I am curious as to why [on the force push diff](https://github.com/freedomofpress/dangerzone/compare/bf372121233470f82960f1aa2f6882edb4509651..01e4eb0a8a0cda9ec8fd0ceb50d20cefd1d38850) [this particular line was changed](https://github.com/freedomofpress/dangerzone/compare/bf372121233470f82960f1aa2f6882edb4509651..01e4eb0a8a0cda9ec8fd0ceb50d20cefd1d38850#diff-9efd70f53282d2854b01b8df2280b506873ade281dd07b7c6d7d132bd020a443L142).

Hm, I made this change btw. See this commit: ac38d99. Could it be that you haven't pulled my changes prior to the push, since Ubuntu Jammy and Debian Bullseye fails?

apyrgio (approved):
A+ for the amazing effort @deeplow, A+++ for the fact that we did it while removing ~300 lines of code. Let's merge it!

Issue #625, opened 2023-11-24
On-host pixels to PDF conversion


deeplow:
# Background

Historically on the containers version of Dangerzone the conversion happens on a second container. This was needed since Dangerzone relied on many linux-native programs for conversion such as GraphicsMagic, ghostscript (for compression via `ps2pdf `). In Qubes the conversion from pixels to PDF already happens [on the host](https://github.com/freedomofpress/dangerzone/blob/bf38c24d99fd70ef551c1acf167f28653ee860ba/dangerzone/isolation_provider/qubes.py#L135).

# Suggestion

Allowing the second phase of the conversion to run in the host opens up several improvements for Dangerzone:

![image](https://github.com/user-attachments/assets/3784379c-6006-4fec-95c4-cd898d63b7fd)

_(partial capture, taken from https://github.com/freedomofpress/dangerzone/issues/658#issuecomment-1861366566)_

Let's explain how the on-host conversion feature brings the above improvements:

### Host container image independently

One of the reasons why the Dangerzone container image is not hosted in a container registry like DockerHub is because we don't have a way to verify its contents during image pull. Normally, this would not be an issue for Dangerzone, since it already restricts the capabilities and privileges of the container. Therefore, we _could_ treat a malicious image similarly as a container that has been pwned by a malicious document.

One problem here is that the same container image is used for the second stage of the conversion. A malicious container image in that stage would cause real damage, because its output is a PDF document, not pixels.

If we performed the pixels to PDF phase on the host instead, this would mean that the container image would be used solely for the first phase, and thus we _could_ potentially push it to an untrusted image registry.

**Benefit:** This could open the door for updating the container image independently from the app, which would be a massive security improvement for Dangerzone.

> [!NOTE]
> That was the original thinking at least. Since creating the above diagram, we now run gVisor within the container image. This means that a malicious container registry can serve an image that does not contain gVisor, and significantly weaken our defenses. The resulting container will still be restricted by our security flags, but the truth is that the original point is no longer that strong.

### Smaller container image

By performing the pixels to PDf phase on host, we can remove the dependency to the OCR models from our image, which are currently used by PyMuPDF. These models take a lot of space (roughly ~300MiB), and removing them can slim down our image considerably.

As for where they will go, the plan is to move them to the host instead. This means will either make them dependencies of the .deb/.rpm packages, or include them in the Windows/macOS installers.

[...]

PR #639, opened 2023-12-06
macos: Update our macOS entitlements

### LINKED ISSUE #638: Make our macOS entitlements stricter
Our macOS [entitlements](https://developer.apple.com/documentation/bundleresources/entitlements) are the following:

https://github.com/freedomofpress/dangerzone/blob/612ac061dec8d701798479af6104c222bc3aa981/install/macos/entitlements.plist#L3-L20

These entitlements were last updated when Dangerzone was considering using HyperKit to spawn VMs (9158d026, on 2021-06-30)

Now that it uses Docker Desktop, we can do the following changes:

1. Re-enable `com.apple.security.app-sandbox`, which offers some important protections and is a requirement for publishing Dangerzone in the MacOS App store, if we ever want to.
2. Remove `com.apple.security.hypervisor` and `com.apple.security.cs.allow-unsigned-executable-memory`, since they are Hyperkit related and lower our protections.

apyrgio:
Our entitlements were last updated when Dangerzone was considering using HyperKit to spawn VMs (9158d02, on 2021-06-30). Now that we use Docker Desktop, we can make them stricter.

Fixes #638


apyrgio:
This PR is blocked because the suggested entitlements break mounting files in the container. This will have to wait until we resolve #625, in which case, we will pass files via the container's standard streams.

almet:
Now that #625 has been merged in main, I've rebased this on top.

apyrgio:
Just tested this PR once more. I'm afraid that the recently merged on-host conversion PR does not solve this issue. I've encountered the following issues:

## Accessing the Docker socket

Opening the GUI app shows the following error message:

Screenshot From 2024-10-21 18-43-13

Note that there's indeed no Unix Domain socket there. When I attempt to specify it via the `DOCKER_HOST` envvar, I get this error:

Screenshot From 2024-10-21 18-44-35

## Accessing files from the Dangerzone CLI

Pointing the `dangerzone-cli` to a file in the host fails with "file is not found":

```console
user@macos ~ % /Applications/Dangerzone.app/Contents/MacOS/dangerzone-cli dangerzone/tests/test_docs/sample-pdf.pdf
Input file not found: make sure you typed it correctly.
user@macos ~ % ls dangerzone/tests/test_docs/sample-pdf.pdf
dangerzone/tests/test_docs/sample-pdf.pdf
```

My understanding is that sandboxed apps cannot read any file in the host, but instead have to open a browser window, and let the user choose the file on their own. See: https://developer.apple.com/documentation/security/accessing-files-from-the-macos-app-sandbox

apyrgio:
Still, even though we can't go full app sandbox here, it doesn't mean that we can't ditch some older entitlements that are no longer used. I have updated this PR to remove some that I found, and tested that the resulting .dmg works on macOS.

almet (approved):
Thanks for the investigation on this. Too bad it's not possible. Neverhtless, let's merge the removal of these old permissions :-)

PR #640, opened 2023-12-06
Bump version to 0.5.1

apyrgio:


apyrgio:
Whoops, fixed, thanks for spotting that.

PR #541, opened 2023-09-01
Enable HWP conversion on MacOS M1 again

OctopusET:
~## **This PR is not working 'yet'**~

### What they do
This PR reverts the patch that disables HWP / HWPX conversion on MacOS M1.
It does not fix conversion on Qubes OS (#494)

### Background
Previously, HWP / HWPX conversion didn't work on MacOS M1 system. #498
Because libreoffice wasn't built with Java support on alpine (aarch64).

Gratefully, the Alpine team has enabled Java support on the aarch64 system.
https://gitlab.alpinelinux.org/alpine/aports/-/commit/74d443f479df15fc57e6fde6ac02a36b24afdded

### Problem
This commit is included in `7.5.5.2-r2`, `libreoffice` package should be `7.5.5.2-r2` or higher to fix this issue.

However 3.18 doesn't have `7.5.5.2-r2` package. (https://pkgs.alpinelinux.org/package/v3.18/community/aarch64/libreoffice)
And Dangerzone image uses the `alpine:latest` image which is `3.18` as of this writing.

### Solution:
1. Wait until the update is available in the Alpine stable
2. Change Docker image `alpine:latest` to `alpine:edge` (**Currently PR is using this solution!**)
3. Install `libreoffice` package from edge repository. (I tried it, but test failed)

Big thanks to @ebandal for the H2ORestart and research on the Java support issue.

Fixes #498

PR #649, opened 2023-12-08
Enable HWP conversion on MacOS (Apple silicon CPU)

### LINKED ISSUE #498: HWP conversion fails on MacOS M1
HWP conversion was recently added in https://github.com/freedomofpress/dangerzone/pull/460 and was tested across various platforms (Windows, MacOS, Linux) through our CI. However, when running our tests on a MacOS M1 platform (which is not available through any of our CI runners), we get the following error:

```
[DEBUG] Marking doc UWYs15 as 'converting'
[INFO ] > /usr/local/bin/docker run --network none -u dangerzone --security-opt=no-new-privileges:true --cap-drop all --rm -v /var/folders/7c/49675rcx2nb0lwjms_6wmdpr0000gs/T/tmp6rzrqzt1/unsafe/input_file:/tmp/input_file:Z -v /var/folders
/7c/49675rcx2nb0lwjms_6wmdpr0000gs/T/tmp6rzrqzt1/pixels:/tmp/dangerzone:Z -e ENABLE_TIMEOUTS=1 dangerzone.rocks/dangerzone /usr/bin/python3 -m dangerzone.conversion.doc_to_pixels
[INFO ] [doc UWYs15] 0% UNTRUSTED> Installing LibreOffice extension 'h2orestart.oxt'
[INFO ] [doc UWYs15] 0% UNTRUSTED> Converting to PDF using LibreOffice                                 
[INFO ] [doc UWYs15] 3% UNTRUSTED> Calculating number of pages
[ERROR] [doc UWYs15] 3% UNTRUSTED> PDF file is corrupted
```

_(taken from the `dangerzone-cli`)_

If one tries to convert the file to PDF via Libreoffice:

```
libreoffice --headless --infilter="Hwp2002_Reader" --convert-to pdf:writer_pdf_Export <file>
```
they will get the following generic error:

```
Error: source file could not be loaded
```

Note that the exact same operation in an Intel MacOS device works.

OctopusET:
Fixes #498

See also: #541 #562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## Unreleased
 
+### Changed
+
+- Feature: Add support for HWP/HWPX files (Hancom Office) for macOS Apple Silicon devices ([issue #498](https://github.com/freedomofpress/dangerzone/issues/498), thanks to [@OctopusET](https://github.com/OctopusET))
+
 ## Dangerzone 0.5.1
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -53,8 +53,7 @@ Dangerzone can convert these types of document into safe PDFs:
 - ODF Graphics (`.odg`)
 - Hancom HWP (Hangul Word Processor) (`.hwp`, `.hwpx`)
   * Not supported on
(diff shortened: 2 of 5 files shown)


apyrgio (approved):
Looks good, thanks again @OctopusET.

PR #651, opened 2023-12-08
Improve Qubes BUILD instructions

deeplow:
This is not critical for the 0.5.0 release.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -164,7 +164,7 @@ The following instructions require typing commands in a terminal in dom0.
 
    ```
    qvm-create --class AppVM --label red --template fedora-38-dz \
-       --prop netvm=""  --prop template_for_dispvms=True \
+       --prop netvm="" --prop template_for_dispvms=True \
        --prop default_dispvm='' dz-dvm
    ```
 


BUILD.md
```diff
@@ -187,96 +189,60 @@ Overview of the qubes you'll create:
...
-1. Create the directory that will contain the Dangerzone RPC calls, if it does
-   not exist already:
-
-   ```
-   sudo mkdir -p /rw/usrlocal/etc/qubes-rpc/
-   ```
+#### In the `fedora-38-dz` template
 
-2. Move the files we copied in the previous step to their proper place:
+1. Install the `.rpm` package you just copied
```
apyrgio, line 222: I think we don't need a number for this section, as it's a single step.

BUILD.md
```diff
@@ -187,96 +189,60 @@ Overview of the qubes you'll create:
...
-# start a shell in the virtual environment
-poetry shell
 
 # run the CLI
-QUBES_CONVERSION=1 ./dev_scripts/dangerzone-cli --help
+QUBES_CONVERSION=1 poetry run  ./dev_scripts/dangerzone-cli --help
 
 # run the GUI
-QUBES_CONVERSION=1 ./dev_scripts/dangerzone
+QUBES_CONVERSION=1 poetry run  ./dev_scripts/dangerzone
```
apyrgio, line 242: Is `QUBES_CONVERSION=1` necessary anymore actually? I think not.

BUILD.md
```diff
@@ -187,96 +189,60 @@ Overview of the qubes you'll create:
...
 ```
 
-Create a .rpm:
+And when create a `.rpm` you'll need to enable the `--qubes` flag.
```
apyrgio, line 245: "when creating"

BUILD.md
```diff
@@ -187,96 +189,60 @@ Overview of the qubes you'll create:
...
-   ```
-   shutdown -h now
-   ```
+Shutdown the `fedora-38-dz` template
 
 ### Developing Dangerzone
 
-From here on, developing Dangerzone is similar as in other Linux platforms. You
-can run the following commands in the `dz` app qube:
+From here on, developing Dangerzone is similar to Fedora. The only differences
```
apyrgio, line 232: I'd suggest still mentioning that development takes place in the `dz` app qube.

apyrgio (changes requested):
The changes look good. I only left a few minor comments.

BUILD.md
```diff
@@ -147,24 +147,26 @@ Overview of the qubes you'll create:
...
    ```
 
    > :bulb: Alternatively, you can use your base Fedora 38 template in the
-   > following instructions. In that case, replace `fedora-38-dz` with
-   > `fedora-38` in the steps below.
+   > following instructions. In that case, skip this step and replace
+   > `fedora-38-dz` with `fedora-38` in the steps below.
 
-2. Create a **disposable**, offline app qube (`dz-dvm`), based on the
+2. Create a **disposable template**, offline app qube (`dz-dvm`), based on the
```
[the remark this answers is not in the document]
apyrgio: I see. Then "disposable template" is a special case of an "app qube", so it's probably redundant to mention both?

BUILD.md
apyrgio: Right, that's the only case that's necessary. Let's keep it.

BUILD.md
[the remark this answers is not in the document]
apyrgio: We discussed this offline and we reached to a better wording. Closing this discussion.

apyrgio (approved):
Thanks for the fixes, looks good!

PR #655, opened 2023-12-11
ci: Test official installation instructions

### LINKED ISSUE #653: `focal` no longer has a Release file
It looks like something went wrong with the 0.5.1 Release file generation for `focal` and it went missing:
https://github.com/freedomofpress/apt-tools-prod/pull/10
https://packages.freedom.press/apt-tools-prod/dists/

apyrgio:
Create a new GitHub Actions workflow which aims to continuously test our official installation instructions. The way we do it is the following:

1. Create two jobs, one for the Debian-based distros, and one for Fedora ones.
2. Copy the instructions from INSTALL.md into each job.
3. Create a matrix that runs the installation jobs in parallel, for each supported distro and version.

The jobs will run only on 00:00 UTC, and not on every PR, since it wouldn't make sense otherwise.

Fix #653

NOTE: The committed code does not enable runs of this CI job on pull requests. Here's the latest successful run instead: https://github.com/freedomofpress/dangerzone/actions/runs/7168694646

### DIFF
--- a/.github/workflows/check_repos.yml
+++ b/.github/workflows/check_repos.yml
@@ -1,0 +1,88 @@
+# Test official instructions for installing Dangerzone
+# ====================================================
+#
+# The installation instructions have been copied from our INSTALL.md file.
+# NOTE: When you change either place, please make sure to keep the two files in
+# sync.
+# NOTE: Because the commands run as root, the use of sudo is not necessary.
+name: Test official instructions for installing Dangerzone
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run every day at 00:00 UTC.
+
+jobs:
+  install-from-apt-repo:
+    name: "Install Dangerzone on ${{ matrix.distro}} ${{ matrix.version }}"
+    runs-on: ubuntu-latest
+    container: ${{ matrix.distro }}:${{ matrix.version }}
+    strategy:
+      matrix:
+        include:
+          - distro: ubuntu
+            version: "23.10"  # mantic
+          - distro: ubuntu
+            version: "23.04"  # lunar
+          - distro: ubuntu
+            version: "22.04"  # jammy
+          - distro: ubuntu
+            version: "20.04"  # focal
+          - distro: debian
+            version: "trixie"  # 13
+          - distro: debian
+            version: "12"  # bookworm
+          - distro: debian
+            version: "11"  # bullseye
+    steps:
+      - name: Add Podman repo for Ubuntu Focal
+        if: matrix.distro == 'ubuntu' && matrix.version == 20.04
+        run: |
+          apt-get update && apt-get -y install curl wget gnupg2
+          . /etc/os-release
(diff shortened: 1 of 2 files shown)


.github/workflows/check_repos.yml
```diff
@@ -0,0 +1,85 @@
...
+          - distro: fedora
+            version: 38
+    steps:
+      - name: Add packages.freedom.press to our YUM sources
+        run: |
+          dnf install -y 'dnf-command(config-manager)'
+          dnf config-manager --add-repo=https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo
+
+      - name: Install Dangerzone
+        run: dnf install -y dangerzone
```
eloquence, line 85: small discrepancy with `INSTALL.md`, any reason not to use `-y` in both places?

.github/workflows/check_repos.yml
```diff
@@ -0,0 +1,85 @@
+# Test official instructions for installing Dangerzone
+# ====================================================
+#
+# The installation instructions have been copied from our INSTALL.md file.
+# NOTE: When you change either place, please make sure to keep the two files in
+# sync.
```
eloquence, line 6: Maybe add a note here that because commands run as root, use of `sudo` is unnecessary?

eloquence (changes requested):
Looks great, really excited to add this to CI to catch  breakages early. :tada: Only left two small nits in review.

One thing we do in a few other repos is to have PR checklists to ensure we perform steps that might be easy to forget about. Here's the one for the SecureDrop repo for example:
https://github.com/freedomofpress/securedrop/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

With the need to keep CI and INSTALL.md in sync (while maintaining small differences that are unavoidable), this might be a good time to introduce such a checklist here as well -- e.g., there could be a checkbox in the PR template saying "If you've modified `INSTALL.md` or the `check_repos` workflow, ensure both are in sync where appropriate". But there may be other options as well and I think this can be handled separately from this PR.

.github/workflows/check_repos.yml
apyrgio: In `INSTALL.md` we explicitly ask the user to validate the signature: https://github.com/freedomofpress/dangerzone/blob/main/INSTALL.md#verifying-dangerzone-gpg-key

With `-y`, we can't do that unfortunately.

.github/workflows/check_repos.yml
apyrgio: Fixed in 70e7caaebdf69e66c6242797b9570a38b02c79f1.

eloquence (approved):

eloquence:
LGTM, thanks :)

apyrgio:
Oh, I never got to comment on your general point regarding keep the two files in sync. I agree with having a PULL REQUEST template, even if it's very short for the time being :+1:

PR #684, opened 2024-01-23
Add Dangerzone logo to Fedora build

### LINKED ISSUE #645: On Fedora (and Qubes Fedora template) the Dangerzone Application Logo is Missing
Dangerzone on Qubes (Fedora template) and Fedora are missing the Dangerzone desktop icon. We missed this when creating our own Fedora packaging from scratch (the `.spec` file).

This is the kind of issue that our semi-automated qa.py script would miss.

deeplow:
Fixes #645

### DIFF
--- a/install/linux/dangerzone.spec
+++ b/install/linux/dangerzone.spec
@@ -246,7 +246,9 @@ sed -i 's/^PySide6.*$/PySide2 = "*"/' pyproject.toml
 install -m 755 -d %{buildroot}/usr/share/
 install -m 755 -d %{buildroot}/usr/share/applications/
 install -m 755 -d %{buildroot}/usr/share/dangerzone/
-install -m 644 install/linux/* %{buildroot}/usr/share/applications/
+install -m 755 -d %{buildroot}/usr/share/pixmaps/
+install -m 644 install/linux/press.freedom.dangerzone.desktop %{buildroot}/usr/share/applications/
+install -m 644 install/linux/press.freedom.dangerzone.png %{buildroot}/usr/share/pixmaps/
 install -m 644 share/* %{buildroot}/usr/share/dangerzone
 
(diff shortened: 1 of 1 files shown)


install/linux/dangerzone.spec
```diff
@@ -244,7 +244,9 @@ sed -i 's/^PySide6.*$/PySide2 = "*"/' pyproject.toml
 install -m 755 -d %{buildroot}/usr/share/
 install -m 755 -d %{buildroot}/usr/share/applications/
 install -m 755 -d %{buildroot}/usr/share/dangerzone/
-install -m 644 install/linux/* %{buildroot}/usr/share/applications/
+install -m 755 -d %{buildroot}/usr/share/pixmaps/
+install -m 644 install/linux/press.freedom.dangerzone.desktop %{buildroot}/usr/share/applications/
+install -m 644 install/linux/press.freedom.dangerzone.png %{buildroot}/usr/share/pixmaps/
```
apyrgio, line 249: I see that while `/usr/share/pixmaps` is in the lookup path, it's not encouraged for `.png` icons. Instead, there's an [Icon Theme Specification](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#install_icons) that suggests the following:

> So, you're an application author, and want to install application icons so that they work in the KDE and Gnome menus. Minimally you should install a 48x48 icon in the hicolor theme. This means installing a PNG file in $prefix/share/icons/hicolor/48x48/apps.

In our case, the icon is 64x64, so maybe we can downscale it, and offer two versions; a 48x48 one and a 64x64 one. Later on, we can do the same for Debian as well.

apyrgio: Ok, it's not critical so let's go with that.

apyrgio (approved):

Issue #685, opened 2024-01-23
Podman can crash during standard stream I/O


apyrgio:
While working on issue #443 and PR #627, we have seen the following weird behavior (https://github.com/freedomofpress/dangerzone/pull/627#issuecomment-1892491968 and https://github.com/freedomofpress/dangerzone/pull/627#issuecomment-1892494187):

1. The client (host) may get stuck while reading page data, and time out at some point.
2. The client (host) may return immediately with a successful error code, without even getting all the pixels back.

The platforms where this behavior is consistent are:

* Ubuntu Focal (we see mostly timeouts there)
* Ubuntu Jammy (we see mostly abrupt returns)
* Debian Bullseye (we see mostly abrupt returns)

Specifically for the abrupt return bug, it seems that on the client side `read()` gets interrupted abruptly, and thus we raise `ConverterProcException()`. Then we try to find the reason for the exception, and the exit code is actually `0`! In that case, the Podman process does not exist, and the container is deleted as well.

Interestingly, with `ps` we see that the underlying conversion process still runs. With `journalctl -f`, we see that a `conmon` component attempts to print binary data to the journal.

apyrgio:
No, it's actually dependent on `conmon (>= 2.0.18~)` as you can see from `apt-cache show podman`. This means that a newer version tramps the one offered by the repos.

But yes, you're right about a) possibly unintended consequences, and b) resigning the packages. I don't like that as well.

apyrgio:
Sigh :face_exhaling:

While building the package, I just realized that this is a [known bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1014030) in Debian Bullseye. The maintainers of the package are doing the same thing we are describing in this issue (https://github.com/freedomofpress/dangerzone/issues/685#issuecomment-1908007913). They have backported the fix from **v2.0.26** into **v2.0.25**, and have created a new package called **2.0.25+ds1-1.1+deb11u1**. This package has been [accepted as a "proposed updated"](https://tracker.debian.org/news/1477995/accepted-conmon-2025ds1-11deb11u1-source-into-oldstable-proposed-updates/) for Debian Bullseye in Sun, 12 Nov 2023.

This package has not landed yet in Debian Bullseye, nor Ubuntu Jammy. We can build a Debian package out of it though and name it **2.0.25+ds1-1.1+deb11u1~fpf1**. This way, I believe that the version order will be:

```
[...]

PR #708, opened 2024-02-13
Install patched conmon from our apt-tools-prod repo

apyrgio:
The oldstable-proposed-updates Debian repo no longer offers a patched `conmon` version, since this has now been integrated in Debian Bullseye (see https://tracker.debian.org/pkg/conmon). Since we want to test as close to the end-user environment as possible, change our CI code to install `conmon` from the apt-tools-prod repo.

Refs #685


apyrgio:
Everything passed, except for the security scans. However, I just realized that our build instructions are not up-to-date with the new change. I pushed a commit to fix them (2703448d60f88be042af705aa5a04f7605b268a4).

PR #709, opened 2024-02-13
README.md bump version to 0.6.0

deeplow:


apyrgio (approved):
Nice! Let's not forget to merge it after we have released Dangerzone 0.6.0.

sudwhiwdh:
New screenshots with the capitalisation would be wonderful.

sudwhiwdh:
> @apyrgio please give me an ack for the images when you have a chance :)

The font in the header is now bold?

apyrgio:
Hm, two things here:

1. From what I see, the font in the header has been bold since at least v0.3 (2021):

    https://github.com/freedomofpress/dangerzone/blob/b4cae378741b65cb535a2235324fe7d4a80eebfd/dangerzone/gui/main_window.py#L547-L549
    
    Since then, we've switched from PySide2 to PySide6, so maybe things are rendered differently. I think this screenshot was before the change to PySide6 as well. In any case, existing users should not understand a difference.
   
2. By merging this PR, users that encounter our GitHub project will be asked to download a version that does not exist yet (0.6.0). I'll send a PR to change that part.

PR #711, opened 2024-02-13
CI: Prevent FIXUP commits

deeplow:


.github/workflows/check_push.yml
```diff
@@ -0,0 +1,19 @@
...
+        distro: debian
+        version: bookworm
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v3
+        - name: prevent fixup commits
+          run: |
+            git fetch origin
+            git status
+            git log origin/main..HEAD | grep FIXUP && exit 1 || true
```
apyrgio, line 19:
```suggestion
            git log origin/main..HEAD | grep -ie '^fixup\|^wip' && exit 1 || true
```

I suggest including both wip commits, and `fixup!` commits that Git produces.

apyrgio: Note that an even better approach is `^\s*` (match any whitespace from the start of the line), but we don't need to be pedantic here.

apyrgio (approved):
Thanks, feel free to merge.

PR #718, opened 2024-02-15
Don't show 50% duplicated progress info

### LINKED ISSUE #715: 50% shows twice in CLI conversion output
*(originally found [in 0.6.0 QA](https://github.com/freedomofpress/dangerzone/issues/704#issuecomment-1943565188))* 50% progress is repeated twice. We should have one of them be like up to 45% to avoid this repetition

noticed

deeplow:
50% would show twice in the conversion progress due to an overlap in conversion progress values. The doc_to_pixels would be from 0-50% and the pixels_to_pdf from 50%-100%.

This commit makes the first part go from 0 to 49% instead.

Fixes #715


apyrgio (approved):
Looks reasonable, thanks!

PR #738, opened 2024-03-04
Simplify QA for Scenario 10

### LINKED ISSUE #719: Move QA Scenario 10 to the CI (except windows and macOS)
Move the following scenario to the CI (except for Windows and macOS where testing the scenario is implicit and easier).

----

#### 10. Updating Dangerzone handles external state correctly.

_(Applies to Linux/Windows/MacOS. For MacOS/Windows, it requires an installer
for the new version)_

Install the previous version of Dangerzone system-wide:

* For MacOS/Windows, use the version from the website.
* For Linux, uninstall Dangerzone from your test environment, and install the
  previous version using our [installation instructions](INSTALL.md). Also,
  keep in mind the following:
  - In order to run commands as root, execute into the container as root with
    `podman exec -it -u root <container ID> bash`.
  - If you encounter a GPG error, run the `dirmngr` command to create the
    necessary directories.

Open the Dangerzone application and enable some non-default settings. Close the
Dangerzone application and get the container image for that version. For
example:

```
$ podman images dangerzone.rocks/dangerzone:latest
REPOSITORY                   TAG         IMAGE ID      CREATED       SIZE
dangerzone.rocks/dangerzone  latest      <image ID>    <date>        <size>
```

_(use `docker` on Windows/MacOS)_

Install the new version of Dangerzone system-wide. For Linux, copy the package
back into the container. Open the Dangerzone application and make sure that the
previously enabled settings still show up.  Also, ensure that Dangerzone reports
that the new image has been installed, and verify that it's different from the
old one by doing:

```
$ podman images dangerzone.rocks/dangerzone:latest
REPOSITORY                   TAG         IMAGE ID        CREATED       SIZE
dangerzone.rocks/dangerzone  latest      <different ID>  <newer date>  <different size>
```

deeplow:
Simplifies the QA process for the previous scenario 10, where the goal was to test how upgrades are handled.
- **Increase settings tests** (~100% coverage)
- **Remove requirement from testing on Linux** - was a bit cumbersome (increased test coverage should help mitigate this removal)
- **Change scenario 10 numbering to be one of the earliest scenarios.**
   When starting QA on windows, the user will have installed the previous version, so it makes sense test the new version install first. This also helps avoid some state changes introduced by having the new version for the first scenarios and then installing the old one and then moving forwards again.

Fixes #719

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -158,20 +158,50 @@ _(Only for MacOS / Windows)_
 Stop the Docker Desktop application. Then run Dangerzone. Dangerzone should
 prompt the user to start Docker Desktop.
 
-#### 3. Dangerzone successfully installs the container image
 
-_(Not for Qubes)_
+#### 3. Updating Dangerzone handles external state correctly.
+
+_(Applies to Windows/MacOS)_
+
+Install the previous version of Dangerzone, downloaded from the website.
+
+Open the Dangerzone application and enable some non-default settings.
+**If there are new settings, make sure to change those as well**.
+
+Close the Dangerzone application and get the container image for that
+version. For example:
+
+```
+$ docker images dangerzone.rocks/dangerzone:latest
+REPOSITORY                   TAG         IMAGE ID      CREATED       SIZE
+dangerzone.rocks/dangerzone  latest      <image ID>    <date>        <size>
+```
+
+Then run the version under QA and ensure that the settings remain changed.
+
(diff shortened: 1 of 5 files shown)


apyrgio (commented):
I have but a minor comment. The rest of the changes look great.

apyrgio (approved):
Thanks for the changes!

PR #743, opened 2024-03-05
dz.ConvertDev: do not teleport .pyc files

### LINKED ISSUE #723: Do not teleport compiled Python code when running dz.ConvertDev on Qubes
Our [`dz.ConvertDev`](https://github.com/freedomofpress/dangerzone/blob/5b6911af8404489187a93c185ac4b188e50b9ac6/qubes/dz.ConvertDev#L4) Qubes policy, which runs in a disposable qube, expects a Python zipfile. This zipfile is created in the dev qube where the `./dev_scripts/dangerzone*` command runs, and it contains the Python compiled files for the `dangerzone.conversion` module.

The problem here is that the disposable qube and the main qube may have different Python versions. This can happen if e.g., the dev qube is based on Fedora 38 (Python 3.11), and the disposable qube is based on Fedora 39 (Python 3.12).

deeplow:
On Qubes the conversion in dev mode would fail when converting from a Fedora 38 development qube via a Fedora 39 disposable qube. The reason was that `dz.ConvertDev` was receiving `.pyc` files, which were compiled for python 3.11 but running on python 3.12.

Unfortunately PyZipFile objects cannot send source python files, even though the documentation is a little bit [unclear on this](https://docs.python.org/3/library/zipfile.html#pyzipfile-objects).

Fixes #723

### DIFF
--- a/dangerzone/isolation_provider/qubes.py
+++ b/dangerzone/isolation_provider/qubes.py
@@ -80,16 +80,24 @@ class Qubes(IsolationProvider):
     def teleport_dz_module(self, wpipe: IO[bytes]) -> None:
         """Send the dangerzone module to another qube, as a zipfile."""
         # Grab the absolute file path of the dangerzone module.
-        import dangerzone.conversion as _conv
+        import dangerzone as _dz
 
-        _conv_path = Path(inspect.getfile(_conv)).parent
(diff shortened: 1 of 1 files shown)


apyrgio (approved):
Thanks for the fix, looks great!

PR #753, opened 2024-03-22
Bump PyMuPDF dependency to unbreak Dangerzone image build

stepnem:
@apyrgio thanks for passing this on to upstream, and kudos
to @julian-smith-artifex-com for the quick response.

With this I've successfully built the image (using
build-image.py) locally.


apyrgio:
Thanks for the PR @stepnem, I was just about to check the new PyMuPDF version as well.

apyrgio (approved):

Issue #402, opened 2023-04-28
The opened application should be recognised as a dangerzone by Fedora 38 with GNOME Desktop


sudwhiwdh:
![02](https://user-images.githubusercontent.com/58292468/235186603-132fdf5d-a4ee-4f85-8387-7db023fa8a64.png)
![07](https://user-images.githubusercontent.com/58292468/235186612-3e951184-464c-4a1e-8ea8-193192d3f1e4.png)
![01](https://user-images.githubusercontent.com/58292468/235186614-97fbe76f-5c15-42df-ad24-84c7e9e148a3.png)

Instead, python3 is displayed in various places (here Dock, Panel, Window Overview)

My wish would be that "dangerzone" is displayed together with the appropriate application icon.

apyrgio:
Right, I can reproduce this locally, but I never managed to dig deeper. Our `.desktop` file uses the "dangerzone" name everywhere, so maybe we are missing a field that Gnome understands.

sudwhiwdh:
@apyrgio Is still present in version 0.4.2.

apyrgio:
Sigh, it most probably is, since we haven't found the time to work on it. It's one of these issues that are not bugs or new features, so they stay for a bit more in the back burner. I have tagged it as `good first issue` though, and I encourage newcomers that want to assist in Dangerzone's development to take a stab at it.

deeplow:
This issue will probably be a bit about finding the proper directory to install the desktop icon.

Here's a section of [a related discussion](https://github.com/freedomofpress/dangerzone/pull/538#discussion_r1321683424):
> The location you're seeing is the one where we previously installed this icon. You're right for pointing this out, and I think this may affect https://github.com/freedomofpress/dangerzone/issues/402. Still, I'd suggest tackling it in a different PR, if you don't mind, since I bet we will go into a rabbit hole.

apyrgio:
We are still bitten by this as @deeplow found out in our recent QA for 0.5.1 (https://github.com/freedomofpress/dangerzone/issues/634#issuecomment-1845715552). Apologies, and I hope we can have a fix out in time.

apyrgio:
I think we have a fix here, which I've verified it works in Fedora 39 (see #774). If we set in our application:

```python
self.setDesktopFileName("press.freedom.dangerzone")
self.setApplicationName("dangerzone")
```

[...]

PR #774, opened 2024-04-09
Set the desktop filename of the Qt application

naglis:
Currently, the app ID of the Dangerzone GUI application when running under Wayland is `python3`, which is not very useful if one wants to automate some action related to the Dangerzone application window (e.g. to always start Dangerzone window in floating mode under Sway WM; to inspect the app ID values of currently running application windows under sway run `swaymsg -t get_tree`).

Setting the desktop filename property also sets the app ID of the application under Wayland. According to [Qt documentation](https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop), the property value should be the name of the application's .desktop file but without the extension.

Qt documentation also states:

> This property gives a precise indication of what desktop entry > represents the application and it is needed by the windowing system to retrieve such information without resorting to imprecise heuristics.

Therefore I also think that setting this property is needed to display the correct application name and icon (taken from the .desktop entry) when running under certain windowing systems (like Wayland) (see also #402, after testing locally under Gnome I believe this should fix it).

### PR questions

According to [Qt 5 docs](https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop) this property was introduced in Qt 5.7. IIUC [PySide2](https://pypi.org/project/PySide2/) is Qt 5.12+, so it should be ok to set it without wrapping it in `try/except AttributeError`? Please correct me if I am mistaken.

Also, I am not sure if perhaps [`Application`'s `__init__`](https://github.com/freedomofpress/dangerzone/blob/7c4e62954f4ee71726de0a23655c8f791f05f290/dangerzone/gui/__init__.py#L55) is the more suitable/correct place for setting it. Please let me know if it should be moved there.

### DIFF
--- a/dangerzone/gui/__init__.py
+++ b/dangerzone/gui/__init__.py
@@ -56,6 +56,21 @@ class Application(QtWidgets.QApplication):
         with open(get_resource_path("dangerzone.css"), "r") as f:
             style = f.read()
         self.setStyleSheet(style)
+
+        # Needed under certain windowing systems to match the application to the
+        # desktop entry in order to display the correct application name and icon
+        # and to allow identifying windows that belong to the application (e.g.
+        # under Wayland it sets the correct app ID). The value is the name of the
+        # Dangerzone .desktop file.
+        self.setDesktopFileName("press.freedom.dangerzone")
+
+        # In some combinations of window managers and OSes, if we don't set an
+        # application name, then the window manager may report it as `python3` or
+        # `__init__.py`. Always set this to `dangerzone`, which corresponds to the
+        # executable name as well.
+        # See: https://github.com/freedomofpress/dangerzone/issues/402
+        self.setApplicationName("dangerzone")
+
         self.original_event = self.event
 
         def monkeypatch_event(arg__1: QtCore.QEvent) -> bool:


naglis:
@apyrgio, thank you for testing. Sorry, I think I did not include all the information to make it clear.

> Can you tell me a bit about your configuration? I see that you use Sway, but what's your OS?

Sure, I am using Arch Linux.

> Have you tested if the window title also shows incorrectly with the stock window manager of your OS? If not, that's ok, I can take a look.

Arch Linux does not really have a "stock" window manager, generally you install and use whichever you prefer. I have tested it under Gnome 46 and Sway 1.9 using Wayland 1.22. Also, please note that the issue with the application name I've mentioned is not the window title, but how the application name/icon is displayed in Gnome's window switcher/Activities overview (see #402 for more information).

> Are we talking about running Dangerzone from source, or installing it from a package?

Both (note that Arch Linux does not have an official package, I am using [a modified PKGBUILD to get 0.6.0 instead of 0.5.1 from the AUR](https://aur.archlinux.org/packages/dangerzone)). Although I don't think that setting the desktop filename would have an effect on application name/icon when running Dangerzone from source while Dangerzone is not also installed as a package as Gnome would not be able to find the matching .desktop entry.

---

The issue I had initially was not being able to control how Dangerzone window is displayed when it runs under Sway. Sway is a tiling window manager, so by default, the window of an application as it starts is maximized to take up the whole screen (ignoring any window width/height preferences set by the application). I wanted Dangerzone window to be displayed in the size [as set in the application](https://github.com/freedomofpress/dangerzone/blob/94179a1d91a3adc993fdf9b32e7311901dade939/dangerzone/gui/main_window.py#L66-L72), as it does not benefit from starting in full screen mode and looks better when not maximized. To do that under Sway, one can assign certain applications, to start in [floating mode](https://wiki.archlinux.org/title/sway#Floating_windows) and not maximized. To be able to identify an application under Sway/Wayland, the `app_id` property of the window can be used (similar to using [xprop](https://www.x.org/releases/X11R7.5/doc/man/man1/xprop.1.html) to find the `class` or `wm_name` attributes under X11). When trying to find the `app_id` of Dangerzone using `swaymsg -t get_tree` I've found that it is `python3` when starting both from source and from the installed package. This value is not very useful, as it is generic and does not uniquely identify Dangerzone. To inspect Wayland window properties under Gnome, one can follow [this answer](https://askubuntu.com/a/1362816) on AskUbuntu. Here are the results I get:

| Without `setDesktopFileName` | With `setDestkopFileName` |
|--------|--------|
| Gnome Looking Glass output for Dangerzone window when not using setDesktopFilename() |  Gnome Looking Glass output for Dangerzone window when using setDesktopFilename() |

I've found that setting the desktop filename using [`setDesktopFileName`](https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop) also sets the Wayland window's app ID to that value. Since according to Qt documentation the value passed to `setDesktopFileName` should be the application's .desktop filename without the extension, this also allows to uniquely identify the application via its Wayland window's app ID.

While searching for potentially related issues on Dangerzone's issue tracker I've stumbled upon #402. I was able to reproduce it locally under Gnome and have found that setting the desktop entry filename (and thus tha app ID) to `press.freedom.dangerzone` (the Dangerzone's .desktop entry filename) allows Gnome to correctly match the application to its desktop entry and display the correct application name/icon. Here are some screenshots for what I mean (notice the icon under the Dangerzone window/the icon and the application name tooltip in the dock at the bottom):

| Without `setDesktopFileName` | With `setDestkopFileName` |
|--------|--------|
| Gnome Activity Overview when not using setDesktopFileName | Gnome Activity Overview when using setDesktopFileName |
 
Therefore, I believe that setting the desktop entry filename might also be the fix for #402. Could you test if the Dangerzone application name/icon in Gnome's Activity Overview (pressing the <kbd>Super</kbd> key) is correctly displayed on you side and, if not, if setting the desktop filename fixes it?

Here are some references to the application name/icon issue from other projects:

- https://github.com/jellyfin/jellyfin-media-player/pull/515
- https://github.com/openscad/openscad/issues/4505

Please let me know if I've missed some details or if something is not clear.

apyrgio:
Thanks you very much for the stellar bug report :star_struck: . The "looking glass" feature of Gnome was something I didn't know about. I did check on a Fedora 39 VM, and here's what I saw:

```
Dangerzone
  wmclass: __init__.py
  <untracked>
```

With your `setDesktopFileName` suggestion, the application type is now recognized:

```
Dangerzone
  wmclass: __init__.py
  app: press.freedom.dangezone.desktop
```

And here's where it gets interesting, and probably completely fixes #402. I used `setApplicationName("dangerzone")` (see [docs](https://doc.qt.io/qtforpython-5/PySide2/QtCore/QCoreApplication.html#PySide2.QtCore.PySide2.QtCore.QCoreApplication.setApplicationName)), and my window properties are finally correct:

```
Dangerzone
  wmclass: dangerzone
  app: press.freedom.dangezone.desktop
```

I'm not sure why your `wmclass` property is different, probably has to do with the way the window managers autocomplete missing properties.

In any case, I'd suggest adding the `setApplicationName("dangerzone)` call, and moving those two calls in `Application.__init__()`. Are you ok with this change?

In that case, let's also add a "Fixes #402" in the commit message.

apyrgio:
I had the same question. The Qt6 docs on [`applicationName`](https://doc.qt.io/qt-6/qcoreapplication.html#applicationName-prop) shed more light into this:

> The application name is used in various Qt classes and modules, most prominently in [QSettings](https://doc.qt.io/qt-6/qsettings.html) when it is constructed using the default constructor. Other uses are in formatted logging output (see [qSetMessagePattern](https://doc.qt.io/qt-6/qtlogging.html#qSetMessagePattern)()), in output by [QCommandLineParser](https://doc.qt.io/qt-6/qcommandlineparser.html), in [QTemporaryDir](https://doc.qt.io/qt-6/qtemporarydir.html) and [QTemporaryFile](https://doc.qt.io/qt-6/qtemporaryfile.html) default paths, and in some file locations of [QStandardPaths](https://doc.qt.io/qt-6/qstandardpaths.html). [Qt D-Bus](https://doc.qt.io/qt-6/qtdbus-index.html), [Accessibility](https://doc.qt.io/qt-6/accessible.html), and the XCB platform integration make use of the application name, too.

_(the [respective](https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop) Qt5 docs have less to say, so there may be some implementation difference)_

To me, this reads like a case for a literal `dangerzone`, not a user-facing string ("Dangerzone"). Also, another thing that caught my eye is:

> If not set, the application name defaults to the executable name.

In our case this is `dangerzone` as well.

apyrgio:
Cool, I'm gonna add this extra setting then, run some tests, and then merge your PR (unless you're already changing stuff).

naglis:
Thanks, please go ahead if you'd like to do it more quickly :+1:  If not, I'll try to update it later today/tomorrow.

apyrgio (approved):

PR #778, opened 2024-04-16
ci: Bump PySide6 version in Fedora end-user envs

apyrgio:
Our end-user Fedora environments, that we create for testing how Dangerzone would operate on a clean Fedora system, require PySide6 to be installed. This package is not available from the official Fedora repos yet.

We have a way instead to check the poetry.lock file, grab the latest PySide6 version from there, and install it from a URL. This is no longer necessary, now that PySide6 6.7.0 will soon be available in all stable Fedora releases. Since the last release maintained by FPF will be 6.6.3.1, we should pin this version in our env.py script. This way, we can bump poetry.lock independently, and let Windows/macOS users get different versions.

Refs freedomofpress/maint-dangerzone-pyside6#5

### DIFF
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -17,6 +17,7 @@ DEFAULT_DEV = False
 DEFAULT_SHOW_DOCKERFILE = False
 DEFAULT_DOWNLOAD_PYSIDE6 = False
 
+PYSIDE6_VERSION = "6.6.3.1"
 PYSIDE6_RPM = "python3-pyside6-{pyside6_version}-1.fc{fedora_version}.x86_64.rpm"
 PYSIDE6_URL = (
     "https://packages.freedom.press/yum-tools-prod/dangerzone/f{fedora_version}/%s"
@@ -309,8 +310,8 @@ class PySide6Manager:
     """Provision PySide6 RPMs in our Dangerzone environments.
 
     This class holds all the logic around checking and downloading PySide RPMs. It can
-    detect the PySide6 version that the project requires, check if an RPM is present
-    under "/dist", and download it.
+    check if the required RPM version is present under "/dist", and optionally download
+    it.
     """
 
     def __init__(self, distro_name, distro_version):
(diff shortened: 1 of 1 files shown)


dev_scripts/env.py
```diff
@@ -17,9 +17,11 @@
 DEFAULT_SHOW_DOCKERFILE = False
 DEFAULT_DOWNLOAD_PYSIDE6 = False
 
+PYSIDE6_VERSION ="6.6.3.1"
 PYSIDE6_RPM = "python3-pyside6-{pyside6_version}-1.fc{fedora_version}.x86_64.rpm"
 PYSIDE6_URL = (
-    "https://packages.freedom.press/yum-tools-prod/dangerzone/f{fedora_version}/%s"
+    #"https://packages.freedom.press/yum-tools-prod/dangerzone/f{fedora_version}/%s"
+    "https://github.com/freedomofpress/yum-tools-prod/raw/db4a1faecc4835c8f38580cb50f53f6d81bcaf40/dangerzone/f{fedora_version}/%s"
```
apyrgio, line 24: This line has been added to make the GitHub Actions use the packages in https://github.com/freedomofpress/yum-tools-prod/pull/20, and should be removed before merging this PR.

dev_scripts/env.py
```diff
@@ -322,20 +324,7 @@ def __init__(self, distro_name, distro_version):
     @property
     @functools.lru_cache
```
legoktm, line 325: This can be removed now, I think

apyrgio: ... and I should use a proper docstring. Thanks for catching this, I have a fixup in ce39cfe071afc0e693040f03e316e69bb8e9b6b5

legoktm: Your fixup seems to have ended up on https://github.com/freedomofpress/dangerzone/pull/779 instead of this PR, but it LGTM.

legoktm (approved):
LGTM, modulo the one part that needs to be reverted and the fixup commit associated with the other PR

dev_scripts/env.py
apyrgio: Ouch. I will properly squash the commits, thanks for flagging this.

dev_scripts/env.py
apyrgio: This is no longer the case.

PR #794, opened 2024-04-30
Fix Qt QAction import

naglis:
In PySide2 QAction is available [under `PySide2.QtWidgets`](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QAction.html#PySide2.QtWidgets.PySide2.QtWidgets.QAction) whereas in PySide6 it resides under [`PySide6.QtGui`](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QAction.html).

Closes #788

PR #795, opened 2024-05-01
Use `exec` instead of `exec_` for Qt dialogs

### LINKED ISSUE #595: PySide DeprecationWarning: `exec_`->`exec`
Seen in QA with Ubuntu 22.04 target during tests:

```
/home/user/dangerzone/dangerzone/gui/main_window.py:208: DeprecationWarning: 'exec_' will be removed in the future. Use 'exec' instead. update_widget.exec_()
```

Wasn't able to find any actual deprecation timeline, just complaints about lack of documentation for this deprecation :upside_down_face: https://bugreports.qt.io/browse/PYSIDE-1583

naglis:
`exec_()` is being deprecated in favor of `exec()`.

Also use [`launch()` helper method](https://github.com/freedomofpress/dangerzone/blob/8cdb2d572038cda19ab2c53b61abbfb9150fe6f5/dangerzone/gui/logic.py#L200) for `Dialog` subclasses.

Note: there is [one case of `exec_()` left](https://github.com/naglis/dangerzone/blob/828c04dfbb6dcfa3c0532358de35ca5650f38f51/dangerzone/gui/__init__.py#L190), however I've found that under PySide2 `QApplication` does not have an `exec()` method. Also, this use of `exec_` does not seem to produce warnings.

Fixes #595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## Unreleased
 
+### Fixed
+
+- Fix a deprecation warning in PySide6, thanks to [@naglis](https://github.com/naglis) ([issue #595](https://github.com/freedomofpress/dangerzone/issues/595))
+
 ## Dangerzone 0.6.1
 
 ### Added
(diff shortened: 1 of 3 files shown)


apyrgio:
Thanks a lot for the PR @naglis. Same as in #794, since we don't need to rush this PR, I'll merge it after the 0.6.1 release.

apyrgio (approved):

PR #798, opened 2024-05-03
Use `%F` field code in .desktop entry

### LINKED ISSUE #797: "Open With Dangerzone" with 2 PDFs results in 2 instances of Dangerzone being opened on Linux
Tested on Arch Linux/Dangerzone 0.6.0 (installed as package)/Gnome Shell 46.1/Nautilus 46.1. IIUC this only applies to Linux systems.

To reproduce:

```sh
mkdir /tmp/test_two_files
touch /tmp/test_two_files/{a,b}.pdf
nautilus /tmp/test_two_files
```

- <kbd>Ctrl</kbd> + <kbd>A</kbd> to select all files
  - right-click with the mouse, select "Open With...", in the dialog select "Dangerzone"
  - or (if Dangerzone is the default application to open PDF) - hit <kbd>Enter</kbd>.

[Screencast displaying opening two PDF files with Dangerzone from Nautilus and two Dangerzone instances being launched](https://github.com/freedomofpress/dangerzone/assets/827324/60957d9b-3e06-4d98-a1a6-29108f03880b)

Since Dangerzone supports converting multiple files at once, this is not ideal, as the conversion of each file has to be started separately in its Dangerzone instance. Also, if many files are opened this way, this can be overwhelming.

---

I believe this is because in its .desktop entry, [Dangerzone uses the `%u`](https://github.com/freedomofpress/dangerzone/blob/8cdb2d572038cda19ab2c53b61abbfb9150fe6f5/install/linux/press.freedom.dangerzone.desktop#L5) field code to signify what arguments it accepts. From the description for `%u` field code on the [desktop entry spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html):

>  A single URL. Local files may either be passed as file: URLs or as file path. 

The description of `%f` (equivalent of `%u` but for local paths) sheds more light on this:

> A single file name (including the path), even if multiple files are selected. **The system reading the desktop entry should recognize that the program in question cannot handle multiple file arguments, and it should should probably spawn and execute multiple copies of a program for each selected file if the program is not able to handle additional file arguments**.

So, to signify that Dangerzone can open several files at once, `%U` or `%F` field code should be used. I believe that Dangerzone should use `%F`, since `%U` is for both local files and URLs and Dangerzone does not support URLs (this can be tested by running `dangerzone 'file:///tmp/test_two_files/a.pdf'` which results in "Input file not found: make sure you typed it correctly." error).

Using `%F` in the .desktop entry results in a single Dangerzone instance being launched with both files selected:

[Screencast displaying opening two PDF files with Dangerzone from Nautilus and a single Dangerzone instance being launched](https://github.com/freedomofpress/dangerzone/assets/827324/01292703-8263-4c04-bce6-2989ad7c8e45)

naglis:
On Linux, the `%u` field code results in multiple Dangerzone instances being launched when opening multiple documents with Dangerzone from e.g. Nautilus, as `%u` signifies that the application (in this case - Dangerzone) can only open a single file/URL at once.

This changes the field code to `%F` as Dangerzone (now) supports converting multiple files at once. We use `%F` (multiple local files) instead of `%U` (multiple files and/or URLs) since Dangerzone does not support opening URLs.

See also the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html) for more information on the field codes.

Fixes #797


apyrgio:
Merged, with an additional changelog item:

> - Linux: Allow opening multiple files at once, when selecting them from the user's file manager ([issue #797](https://github.com/freedomofpress/dangerzone/issues/797))

_(I'll add proper attribution after prior to the release, sorry for missing that...)_

Also, I tested on macOS and it seems that it's bitten by a similar problem, albeit in a different way (see https://github.com/freedomofpress/dangerzone/issues/800). If the fix is trivial, I'll try to include it in the 0.6.1 release.

PR #811, opened 2024-05-22
Minor changes

almet:
A small pull request with changes I did while reading the current codebase.

I've added some comments directly on Github for places where that made sense.


dangerzone/cli.py
```diff
@@ -13,8 +13,6 @@
 from .logic import DangerzoneCore
 from .util import get_version, replace_control_chars
 
-F = TypeVar("F", bound=Callable[..., Any])
```
almet, line 16: This wasn't used in this file.

dangerzone/gui/logic.py
```diff
@@ -113,9 +112,9 @@ def _find_pdf_viewers(self) -> Dict[str, str]:
                                 "application/pdf" in desktop_entry.getMimeTypes()
                                 and "dangerzone" not in desktop_entry.getName().lower()
                             ):
-                                pdf_viewers[desktop_entry.getName()] = (
-                                    desktop_entry.getExec()
-                                )
+                                pdf_viewers[
+                                    desktop_entry.getName()
+                                ] = desktop_entry.getExec()
```
almet, line 117: This is simply a formatting change, made automatically by ruff.

dangerzone/util.py
```diff
@@ -28,21 +27,21 @@ def get_resource_path(filename: str) -> str:
     if getattr(sys, "dangerzone_dev", False):
         # Look for resources directory relative to python file
         project_root = pathlib.Path(__file__).parent.parent
-        prefix = project_root.joinpath("share")
```
almet, line 31: This might be a question of preference, and I don't have any strong opinion on the matter. This is in a separate commit, so can be reverted if you prefer the version with `.joinpath()`.

tests/gui/test_main_window.py
```diff
@@ -15,10 +14,7 @@
 from dangerzone.gui.logic import DangerzoneGui
 from dangerzone.gui.main_window import ContentWidget
 from dangerzone.gui.updater import UpdateReport, UpdaterThread
-from dangerzone.util import get_version
 
-from .. import sample_doc, sample_pdf
-from . import qt_updater as updater
```
almet, line 21: I found it weird to have the fixtures imported (because they don't need to be), but I might be missing something?

almet:
I converted this PR to a draft for now, until I have the tests running properly on my machine, so I can replicate what the CI is telling me :-)

almet:
I've put the changes for pytest fixtures in a separate PR: #813

almet:
Current linting failures are due to [a bug on mypy](https://github.com/python/mypy/issues/9457), which will mark code as unreachable if behind a boolean assert, like this:

```python
assert something is True
assert somethingelse # <-- this is considered unreachable
```

I'm considering removing the `--warn-unreachable` flag for the tests.

tests/gui/test_main_window.py
almet: Taken care of in #813

almet:
Tests are failing due the latest Alpine image. This should be fixed with #812. 

@apyrgio I'm curious about your feedback on disabling the mypy `--warn-unreachable` flag for the tests. Maybe there is a better way :-)

almet:
> I'd be more inclined to use # type: ignore [unreachable] in affected lines, since there may be an actual case where something is unreachable in our tests, and we want t the linter to flag it.

That's indeed a better solution :-) I've updated the branch accordingly, and force-pushed.

dangerzone/__init__.py
```diff
@@ -13,4 +13,4 @@
 if mode == "cli":
     from .cli import cli_main as main
 else:
-    from .gui import gui_main as main
+    from .gui import gui_main as main  # NOQA
```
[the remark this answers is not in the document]
almet: I've updated with a more specific reason

apyrgio (approved):
Thanks for the changes, they look good to me. I think once you rebase this PR on top of the `main` branch, and all the tests pass, you can merge it.

almet:
Just rebased this branch on top of latest `main`

Issue #801, opened 2024-05-08
Dangerzone segfaults on Fedora 39/40


apyrgio:
It has come to our attention that Fedora 39/40 users, who have updated their systems after Saturday, May 4th, cannot run Dangerzone. Opening the application via the desktop icon seems to do nothing, but if you open it through the CLI, you may see a segmentation fault similar to this:


```bash
$ PYTHONFAULTHANDLER=1  dangerzone                                                                                                                                                                        
Fatal Python error: Segmentation fault                                                                                 
                                                                                                                       
Current thread 0x00007fed65981b80 (most recent call first):                           
  File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1289 in create_module
  File "<frozen importlib._bootstrap>", line 813 in module_from_spec        
  File "<frozen importlib._bootstrap>", line 921 in _load_unlocked          
  File "<frozen importlib._bootstrap>", line 1331 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1360 in _find_and_load                       
  File "/usr/lib64/python3.12/site-packages/shiboken6/__init__.py", line 27 in <module>
  File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed     
  File "<frozen importlib._bootstrap_external>", line 995 in exec_module        
  File "<frozen importlib._bootstrap>", line 935 in _load_unlocked                
  File "<frozen importlib._bootstrap>", line 1331 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1360 in _find_and_load
  File "/usr/lib64/python3.12/site-packages/PySide6/__init__.py", line 64 in _setupQtDirectories
  File "/usr/lib64/python3.12/site-packages/PySide6/__init__.py", line 124 in <module>
  File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 995 in exec_module
  File "<frozen importlib._bootstrap>", line 935 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1331 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1360 in _find_and_load
  File "/usr/lib/python3.12/site-packages/dangerzone/gui/__init__.py", line 20 in <module>
  File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed
[...]

Issue #493, opened 2023-07-31
GUI tests fail on Ubuntu Focal


apyrgio:
GUI tests have started to fail on Ubuntu Focal, once we added an unrelated test in our test suite. And by unrelated, we mean that creating even a dummy `tests/isolation_provider/test_base.py` test that returns immediately makes them error out with:

```
[...]
tests/gui/test_main_window.py::test_default_menu PASSED                  [ 83%]
tests/gui/test_main_window.py::test_no_update PASSED                     [ 84%]
tests/gui/test_main_window.py::test_update_detected PASSED               [ 85%]
tests/gui/test_main_window.py::test_update_error Fatal Python error: Segmentation fault

Thread 0x00007f49c5d42700 (most recent call first):
  File "/usr/lib/python3.8/traceback.py", line 359 in extract
  File "/home/circleci/.cache/pypoetry/virtualenvs/dangerzone-3aSsmiER-py3.8/lib/python3.8/site-packages/exceptiongroup/_formatting.py", line 96 in __init__
  File "/home/circleci/.cache/pypoetry/virtualenvs/dangerzone-3aSsmiER-py3.8/lib/python3.8/site-packages/exceptiongroup/_formatting.py", line 179 in __init__
  File "/usr/lib/python3.8/traceback.py", line 103 in print_exception
  File "/usr/lib/python3.8/logging/__init__.py", line 626 in formatException
  File "/usr/lib/python3.8/logging/__init__.py", line 676 in format
  File "/usr/lib/python3.8/logging/__init__.py", line 929 in format
  File "/usr/lib/python3.8/logging/__init__.py", line 1085 in emit
  File "/usr/lib/python3.8/logging/__init__.py", line 1187 in emit
  File "/usr/lib/python3.8/logging/__init__.py", line 954 in handle
  File "/usr/lib/python3.8/logging/__init__.py", line 1661 in callHandlers
  File "/usr/lib/python3.8/logging/__init__.py", line 1599 in handle
  File "/usr/lib/python3.8/logging/__init__.py", line 1589 in _log
  File "/usr/lib/python3.8/logging/__init__.py", line 1475 in error
  File "/usr/lib/python3.8/logging/__init__.py", line 1481 in exception
  File "/home/circleci/project/dangerzone/gui/updater.py", line 294 in check_for_updates
  File "/home/circleci/project/dangerzone/gui/updater.py", line 300 in run

Current thread 0x00007f49d6b55740 (most recent call first):
[...]

PR #813, opened 2024-05-23
tests: remove fixture imports in the tests

almet:
Fixtures should find their way by themselves, as mentioned in the [pytest docs](https://docs.pytest.org/en/4.6.x/fixture.html#conftest-py-sharing-fixture-functions).

> You don’t need to import the fixture you want to use in a test,
> it automatically gets discovered by pytest. The discovery of fixture
> functions starts at test classes, then test modules, then conftest.py
> files and finally builtin and third party plugins.

The tests are currently failing, I believe because of the way they are run. I'm exploring how to use [pytest xdist](https://pytest-xdist.readthedocs.io/en/stable/) to have them run in separate processes.


almet:
> The tests are currently failing, I believe because of the way they are run. I'm exploring how to use [pytest xdist](https://pytest-xdist.readthedocs.io/en/stable/) to have them run in separate processes.

It turns out the tests weren't able to import the fixtures because they were not defined in the correct `conftest.py` file, where pytest is looking for them. I moved them (and the other utilities that were lying in the `__init__.py` files, which hopefully should be enough for it to work.

almet:
*I believe the current tests are failing because of #801, so it is not related to these changes.*

I would be interested to have more information about why the tests were run in separate processes, as mentioned in https://github.com/freedomofpress/dangerzone/issues/493. 

Especially, I would like to try to reproduce the issue, as the changes on how we load and use the fixtures _might_ fix it.

apyrgio:
The tests pass locally in my machine. However, if I specifically try to run `pytest -v tests/gui/` (at least in Ubuntu Focal), it fails. Are you seeing something similar?

almet:
> The tests pass locally in my machine. However, if I specifically try to run pytest -v tests/gui/ (at least in Ubuntu Focal), it fails. Are you seeing something similar?

I can't reproduce, (un)fortunately, Here are the comments I'm running and the output:

```bash
./dev_scripts/env.py --distro ubuntu --version focal build-dev
./dev_scripts/env.py --distro ubuntu --version focal run --dev bash -c "cd dangerzone; poetry run pytest -v tests/gui"
```

apyrgio:
Interesting, I rebuilt my Ubuntu Focal dev environment, and I can't reproduce it as well. It could be due to the PySide6 / Python3 update. Let's hope we'll never have to find out :slightly_smiling_face: 

Feel free to merge, thanks!

apyrgio (approved):

Issue #633, opened 2023-12-05
Defense in Depth - Traceless Sanitization


apyrgio:
# Security Concern

One aspect of the sanitization that Dangerzone has not targeted yet is avoid leaving traces of the converted file on disk. Depending on your threat model, this may be troublesome for two reasons:

1. It allows an attacker to uncover converted files, by looking for [magic numbers ](https://en.wikipedia.org/wiki/Magic_number_(programming)) on the disk (see [extundelete](https://extundelete.sourceforge.net/) project).
2. It can nullify a [plausible deniability](https://en.wikipedia.org/wiki/Plausible_deniability) defense, if there is proof that a sensitive file was at some point stored in the computer.

# Current Situation

Dangerzone uses Linux commands (`libreoffice`, `gm`, `pdftoppm`, `tesseract`) for the various stages of the file conversion. Most of these commands rely on a file to work. Dangerzone is passing files from command to command using two locations:

1. Temporary directories created on the host, which are then mounted into the container (not applicable to Qubes)
2. The `/tmp` directory in the container.

> [!NOTE]
> A few notes on the `/tmp` dir of a container. This directory is not guaranteed to be backed by `tmpfs`, unless you pass a flag like `--tmpfs` / `--mount type=tmpfs`. Even in that case, this functionality may not be available across platforms. For example, Docker states that tmpfs mounts are [only available on Linux](https://docs.docker.com/storage/tmpfs/#limitations-of-tmpfs-mounts). Also, WSL used to [emulate `tmpfs` on disk](https://www.reddit.com/r/bashonubuntuonwindows/comments/69zez8/comment/dhb49z0/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).
>
> Note that there are contradicting accounts on whether WSL2 supports tmpfs on RAM:
> * https://github.com/docker/for-win/issues/12401#issuecomment-989262473
> * https://stackoverflow.com/questions/64401190/can-i-use-docker-tmpfs-in-wsl2-for-running-docker-containers-on-ram#comment117009551_64838516
>
> In any case, we have to verify the avaiability of `tmpfs` mounts on each platform separately.

# Upcoming Improvements

The file passing implementation will drastically change for two reasons:
1. We are working on integrating PyMuPDF with Dangerzone (#622) , meaning that most of the file conversion steps **within the sandbox** can take place in memory, instead of invoking commands.
[...]

PR #815, opened 2024-05-23
Add design document for gVisor integration

apyrgio:
Add a design document for the gVisor integration, based on the discussions we had in #590. Also, add a few more design documents that are scavenged from other GitHub issues / dirs.


docs/developer/gvisor.md
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+```
+
+## Security considerations
+
+* gVisor does not have an official release on Alpine Linux. The developers
+  provde gVisor binaries from a GCS bucket. In order to verify the integrity of
+  these binaries, they also provide a SHA-512 hash of the files.
```
almet, line 278: Like you, I'm a bit concerned that we are downloading the gVisor binaries directly from an external URL, which might be compromised and offer a nice attack vector when people are building their containers, especially as we are lowering the security of the outer container.

I'm not sure what type of practical solutions there are. Here one option I can think of:

The gVisor team is already already using [a GPG signing key](https://gvisor.dev/archive.key) for the apt repository, so we might use the signed binaries and check them against this key? 🤔

almet (approved):
Thanks, the overall document is really clear, even for a newcomer to containers (with a bit of research), so I guess it's taking the good stance 👍

docs/developer/gvisor.md
apyrgio: Yeah, I'm afraid that this signing key is used to sign the release files for the APT repo. I'm afraid the actual binaries are not signed with this key.

Ultimately, we'll probably have to go down the same way we do for the PyPI packages, that is trust on first use (TOFU).

apyrgio: Yeah, the Debian repo is a nice way forward. It so happens that we want to switch from the Alpine container image to the Debian image, so using the Debian repo will definitely fit in that case.

apyrgio:
I've added some small fixups that cover some points we discussed with @EtiennePerot, and the fact that we no longer disable seccomp when using Docker on Windows / macOS.

docs/developer/gvisor.md
```diff
@@ -0,0 +1,293 @@
...
+   ```
+   /usr/bin/python3 -m dangerzone.conversion.doc_to_pixels
+   ```
+
+2. _(outer container)_ The entrypoint code then creates an OCI config for
+   `runsc` with the following properties:
+   * Use UID/GID 1000 in the **inner** container image.
+   * Run the command we detected on step 1.
+   * Enable only the following Linux capabilities in the **inner** container
+     image: `CAP_SETUID` and `CAP_SETGID`. These capabilities can be removed
```
EtiennePerot, line 91: This is now outdated with commit 2cc67c6af0f09468d6f1b1cd2d51441eb3f5d43c

docs/developer/gvisor.md
```diff
@@ -0,0 +1,293 @@
...
+     once https://github.com/google/gvisor/issues/9918 is fixed, but the
+     important thing is that they won't be passed to the conversion process.
+   * Limit the number of open files to 4096.
+   * Use the `/dangerzone-image/rootfs` directory as the root path for the
+     **inner** container.
+   * Mount a gVisor view of the `procfs` hierarchy under `/proc` , and then
+     mount `tmpfs` in the `/dev`, `/sys` and `/tmp` mount points. This way, no
+     host-specific info may leak to the **inner** container.
```
EtiennePerot, line 99: gVisor doesn't mount any host filesystem by default. If the mounts are removed from the OCI config, these directories will simply not exist in the inner container. The mounts in the OCI config for `/proc`, `/dev`, `/sys` etc are mostly just because most Linux applications expect these mounts to exist, rather than being about hiding host-specific info. (Nonetheless, this bullet point is right that there's no leak)

docs/developer/gvisor.md
EtiennePerot: > If you really really wanted, you could fetch the package from the apt repository using `apt download ...`, which verifies the signature, unpack the deb and extract the binary from it. I'm not sure if that provides any more meaningful trust though depending on how gvisor is signing their repository.

The gVisor release process pushes the same binary to both places. The `apt` solution is a bit better because the `runsc` binary is signed as part of the gVisor release process, as opposed to the just hash-based solution which could be susceptible to HTTPS MitM during the Dangerzone image build process.

Alternatively you can always build gVisor from source.

docs/developer/environments.md
```diff
@@ -0,0 +1,133 @@
...
+more nuanced, for two reasons:
+
+1. Wayland support is not that mature on Linux, so we need to
+   [set some extra environment variables](https://github.com/mviereck/x11docker/wiki/How-to-provide-Wayland-socket-to-docker-container). To simplify things, we will target
+   X11 / XWayland hosts, which are the majority of the Linux OSes out there.
+2. Sharing the Unix Domain socket does not allow the client to talk to the
+   display server, for security reasons. In order to allow the client, we need
+   to mount a magic cookie stored in a file pointed at by the `$XAUTHORITY`
+   envvar. Else, we can use `xhost`, which is considered slightly more dangerous
+   for multi-user environments.
```
EtiennePerot, line 87: By the way, I'm trying to build something new in this area, involving securely running GUI applications. :)

docs/developer/gvisor.md
EtiennePerot: I realize now that this rewording was probably due to [my comment here](https://github.com/freedomofpress/dangerzone/commit/8641b66b0db634d1b6b849f9047a93671d7c5a13#r142338345). Sorry if I had emphasized the wrong thing :) I think this bullet point in the .md file is fine as-is, no need to change it.

docs/developer/gvisor.md
apyrgio: Nice catch, I'll remove it.

docs/developer/gvisor.md
apyrgio: Got it. I was not aware that the directories are not created, but I guess I should have expected it.

Ok, so I'm leaving this bullet point as is. To be honest, I was itching to also write that "conversely, no container data persist to the host", due to the usage of `tmpfs`. This would be the final blow to https://github.com/freedomofpress/dangerzone/issues/633. Still, because `tmpfs` is backed by swap, I don't think we can safely make this assumption.

In the future, we can consider mounting `ramfs` on the LibreOffice dir only (currently `/home/dangerzone`), since it's under our control, and we create a single file there.

docs/developer/environments.md
apyrgio: That's exciting, and notoriously difficult (if one considers the Wayland saga). Check out what the Qubes folks have been doing there, if you haven't done so already: https://www.qubes-os.org/doc/gui/

docs/developer/gvisor.md
apyrgio: > as opposed to the just hash-based solution which could be susceptible to HTTPS MitM (somehow) during the Dangerzone image build process.

Honestly, I'm less worried about an HTTPS attack, and more worried about a credential leak that would lead to a compromised GCS bucket. Especially given the fact that the checksum is served by the same bucket. That's why having a signature gives me more peace of mind.

For the last part (compiling gVisor from source), does the gVisor team offer signed tarballs, or signed Git tags, perhaps?

apyrgio: @legoktm I see that Debian is lagging behind gVisor's official releases: https://tracker.debian.org/pkg/golang-gvisor-gvisor. I also see that a user has pointed them to the latest gVisor release: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1070301

Anw, this is just an fyi. It would be nice having a recent gVisor in the official repos.

EtiennePerot: > For the last part (compiling gVisor from source), does the gVisor team offer signed tarballs, or signed Git tags, perhaps?

No, and even if it did, it would probably not be done in a way that meaningfully adds security (i.e. it would probably just be a bot blindly signing every commit, rather than individual contributors signing their own commits). The way the development workflow happens is just... too convoluted (to be generous) to make this work :(

apyrgio:
@almet I'm squashing the fixups. Are we good to merge?

almet:
Sure, go ahead ✅

Issue #674, opened 2024-01-15
Migrate from CircleCI to GitHub actions


apyrgio:
We want us to have all of our workflows in a single place. GitHub actions has proven stable for the time being, and is free, so it's best to move to that.

PR #821, opened 2024-05-28
build: bump CircleCI runners

### LINKED ISSUE #681: Update CircleCI runners
The Linux image we are using for the CircleCI runners has been deprecated (read the [official announcement](https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177)). Full deprecation will take place on September 30th, 2024, and intermittent outages (brownouts) are expected to start happening from February 26th.

By that date, we either need to bump the image of our runners (we have done so in the past without much trouble, or migrate to GitHub actions (#674).

almet:
Try using `:current` as specified in their documentation.

Fixes #681

If this simple approach doesn't cut it, I'm not sure it's worth digging further, as we plan to move away from CircleCI in the future, as mentioned in #674

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
 
   build-container-image:
     machine:
-      image: ubuntu-2004:202111-01
+      image: ubuntu-2004:current
     steps:
       - checkout
       - run: *install-podman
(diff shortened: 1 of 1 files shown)


apyrgio:
It seems that the `convert-test-docs` CI test is failing. Still, I think it's more worth doing the migration to GitHub actions (#674), than battling the remaining CircleCI issues. Therefore, I would suggest closing this PR as "unplanned".

PR #826, opened 2024-06-03
Fix some minor lint issues

apyrgio:
This should fix some minor lint issues that had slipped in the `main` branch. There are two CI issues remaining, but they will be fixed after #818 is merged.


almet (approved):
LGTM

PR #838, opened 2024-06-19
Use `LF` line-ending for `gvisor_wrapper/entrypoint.py`

almet:
Otherwise, Dangerzone images built on Windows might not be able to work, as they cannot load the entrypoint file, triggering an error.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+*     text=auto
+*.jpg -text
+*.gif -text
+*.png -text


almet:
One test is failing: 

```
FAILED tests/test_cli.py::TestExtraFormats::test_hancom_office[sample-hwp.hwp] - assert 1 == 0
 +  where 1 = <CLIResult SystemExit(1)>.exit_code
```

---

I was surprised the CI didn't catch this error (the image built from windows not being functional).

The current CI tests do not build the images, which makes the CI faster, but also doesn't catch some errors, like this one.

I see multiple good reasons (:polar_bear: :ice_cube:, :hourglass_flowing_sand:) for using `DUMMY_CONVERSION = 1` in some builds, but it might also prove useful to build the image from windows and/or macOS once in a while.

.gitattributes
```diff
@@ -0,0 +1 @@
+dangerzone/gvisor_wrapper/entrypoint.py    eol=lf
```
apyrgio, line 1: I was thinking more in the lines of `* text=auto`, which is arguably the most sensible choice and won't bite us for other files as well. Are you ok with that?

almet: Doing it for every file might be okay.

I just wonder how that would play with `.jpg`, `.gif` and `.png`. In case we don't want to change these ones we could add

```
*.jpg           -text
*.gif           -text
*.png           -text
```

apyrgio: Hm, I thought that Git could distinguish between text and binary files. You're right though, I've seen `.gitattributes` doing what you propose, and it's probably better than what we have now.

almet: I updated the PR accordingly in [`cb87c45` (#838)](https://github.com/freedomofpress/dangerzone/pull/838/commits/cb87c456bcb8f14e00a506f99a761883a70e22a5)

apyrgio (approved):

apyrgio:
Thanks for the change Alexis. Hopefully we won't have to deal with it again. Feel free to merge!

PR #840, opened 2024-06-20
Drop support for Fedora 38

almet:
Fedora 38 is EOL since 21 May 2024, so this removes the specific branches we had checking for it, and updates the related instructions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -327,33 +327,6 @@ jobs:
             ./dev_scripts/env.py --distro fedora --version 39 run --dev \
                 bash -c 'cd dangerzone; poetry run make test'
 
-  ci-fedora-38:
-    machine:
-      image: ubuntu-2004:202111-01
-    steps:
-      - checkout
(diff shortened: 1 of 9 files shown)


apyrgio:
I see two places that still refer to Fedora 38:

```
git grep -i fedora | grep 38
BUILD.md:   > :bulb: Alternatively, you can use your base Fedora 38 template in the
[...]
dev_scripts/qa.py:class QAFedora38(QAFedora):
```

The first should change to refer to Fedora 40. What we're saying to the user there is that they can install Dangerzone on their base Fedora 40 template, instead of a separate one. For the second reference, we should just remove it.

almet:
:bookmark: Hey, thanks for catching this up. Mentally bookmarking the `-i` flag and the `grep` on the version number for later reference.

apyrgio (approved):

apyrgio:
Oh nice, you found another instance of Fedora 38 as well. Thanks for the changes, feel free to merge.

PR #843, opened 2024-06-20
Update screenshots

almet:
Two things to note:

- The new screenshots are smaller than the previous ones
- They've been taken from a Linux machine, where the previous ones were from a macOS machine

Generally speaking, the only thing that seem to change is the version number, (so we might want to just close this PR instead).


apyrgio:
Hm, I think that the borders of the screenshots were not captured properly: https://github.com/freedomofpress/dangerzone/tree/2024-06-update-screenshots. Any idea why?

Also, in general macOS has more aesthetically pleasing fonts and colours, so a screenshot from there would be much better. If we don't have one at hand, then that's ok.

Finally, did you manage to find the document trove that deeplow had used for the previous screenshots? Amazing!

almet:
> Hm, I think that the borders of the screenshots were not captured properly: [`2024-06-update-screenshots`](https://github.com/freedomofpress/dangerzone/tree/2024-06-update-screenshots?rgh-link-date=2024-06-20T14%3A23%3A27Z). Any idea why?

Probably because I did adjust it manually, will try again :-)

> Also, in general macOS has more aesthetically pleasing fonts and colours, so a screenshot from there would be much better. If we don't have one at hand, then that's ok.

That's true! I don't have my macOS device with me, but I can do it later!

> Finally, did you manage to find the document trove that deeplow had used for the previous screenshots? Amazing!

I actually copied the filenames only :upside_down_face:

apyrgio:
> I actually copied the filenames only 🙃

Damn, it never crossed my mind to do that. Ingenious...

apyrgio:
> Generally speaking, the only thing that seem to change is the version number, (so we might want to just close this PR instead).

Sigh, I just saw what you wrote here, sorry. You're right, the new drag-and-drop interface is not evident from the screenshots. I think it's fine to close the PR, if the version number is the only difference.

almet:
Yep, that makes sense. At least we caught some other problems along the way :-)

PR #845, opened 2024-06-25
Provide a seccomp policy on docker if using "old" version of runc

almet:
That's what I have, @apyrgio. I renamed the "runtime" references to "container_engine" in the hopes of making it clearer, and added a "get_runtime_version" static method.

Not much, but that's a start :-)


apyrgio:
I have sent a PR (#847) which does things a bit differently than this one. The most important distinction is that it does not use the `runc` runtime version for compatibility checks, so we don't need to refactor our code to change references from "container runtime" to "container engine".

To be honest though, this is a fair point, we should do it at some time.

PR #847, opened 2024-06-26
Use a custom seccomp policy for older Docker Desktop releases

### LINKED ISSUE #846: Handle seccomp policies that don't include ptrace()
# Problem

Our recent gVisor integration (#590) requires allowlisting the `ptrace(2)` system call in the **outer** container, in order to spawn the **inner** container with `runsc`. Nowadays, this is the default [1], but we have encountered systems that don't allow this system call, and thus Dangerzone cannot run in them, at least out of the box.

Affected systems are:
* Ubuntu Focal (via OpenSUSE's repo), with Podman version 3.4.2
* Ubuntu Jammy, with Podman version 3.4.4
* Debian Bullseye, with Podman version 3.0.1
* Older Docker Desktop releases, e.g., with `runc` version 1.1.5

# Background

Before explaining how we plan to fix this issue, we'll give some background on the `ptrace(2)` system call.

First of all, why is this syscall dangerous in the first place? The main reason is that a malicious process can use it in order to escalate its privileges, or thwart some system protections. A real-life example is  [CVE-2019-2054](https://nvd.nist.gov/vuln/detail/CVE-2019-2054). This CVE is the reason why `ptrace(2)` is not allowed in Linux kernels < 4.8, but it's not the only ptrace-related CVE that has been reported.

In order to control the scope of `ptrace(2)` system call, the Linux kernel offers the following mechanisms:

1. The `CAP_SYS_PTRACE` [Linux capability](https://man7.org/linux/man-pages/man7/capabilities.7.html). If this capability is enabled, then the process can have full tracing capabilities, such as tracing other processes that it has not started. If this capability is not granted, then the usage of `ptrace(2)` is still allowed, but restricted through the mechanisms listed below.
2. Disabling the system call (or arguments to it) via a seccomp policy. For instance:
    - Docker originally had disabled `ptrace(2)` in their seccomp policy, and then [re-enabled](https://github.com/moby/moby/commit/1124543ca8071074a537a15db251af46a5189907) it for kernels >= 4.8.
    - Podman similarly [lifted](https://github.com/containers/buildah/issues/3833) this restriction a few years later. 
    - Containerd [did so](https://github.com/containerd/containerd/pull/6810) around the same time.
3. The YAMA Linux Security Module [`ptrace_scope`](https://www.kernel.org/doc/html/v4.15/admin-guide/LSM/Yama.html) setting. This setting controls the behavior of `ptrace(2)` system-wide. In the Linux platforms we support, the default seems to be `1`, i.e., allow `ptrace(2)` only for processes that the parent has direct relationship with (e.g., child processes).

[1] See Podman's [seccomp policy](https://github.com/containers/common/blob/ac141149be3478f8b0cab7c54eddf4dbec52bd34/pkg/seccomp/seccomp.json#L314), Docker's [seccomp policy](https://github.com/moby/moby/blob/ff1e2c0de72a1bbbe4cdbe1558da57d327899df5/profiles/seccomp/default.json#L416-L425), and `containerd`'s [seccomp policy](https://github.com/containerd/containerd/blob/c673fcc9af3ec4a44ac112bd1e8e6479862a10af/contrib/seccomp/seccomp_default.go#L490-L504).

apyrgio:
We are aware that some Docker Desktop releases before 25.0.0 ship with a seccomp policy which disables the `ptrace(2)` system call. In such cases, we opt to use our own seccomp policy which allows this system call. This seccomp policy is the default one in the latest releases of Podman, and we use it in Linux distributions where Podman version is < 4.0.

Fixes #846

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -56,7 +56,12 @@ class Container(IsolationProvider):
         """
         # Get the Docker/Podman version, using a Go template.
         runtime = Container.get_runtime_name()
-        cmd = [runtime, "version", "-f", "{{.Client.Version}}"]
+        if runtime == "podman":
+            query = "{{.Client.Version}}"
+        else:
+            query = "{{.Server.Version}}"
+
+        cmd = [runtime, "version", "-f", query]
         try:
             version = subprocess.run(
                 cmd, capture_output=True, check=True
@@ -104,6 +109,11 @@ class Container(IsolationProvider):
           - This particular argument is specified in `start_doc_to_pixels_proc()`, but
             should move here once #748 is merged.
         """
+        # This file has been copied as is [1] from the official Podman repo. See:
+        #
+        # [1] https://github.com/containers/common/blob/d3283f8401eeeb21f3c59a425b5461f069e199a7/pkg/seccomp/seccomp.json
+        seccomp_json_path = get_resource_path("seccomp.gvisor.json")
+        custom_seccomp_policy_arg = ["--security-opt", f"seccomp={seccomp_json_path}"]
         if Container.get_runtime_name() == "podman":
             security_args = ["--log-driver", "none"]
             security_args += ["--security-opt", "no-new-privileges"]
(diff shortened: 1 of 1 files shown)


dangerzone/isolation_provider/container.py
```diff
@@ -56,7 +56,12 @@ def get_runtime_version() -> Tuple[int, int]:
         """
         # Get the Docker/Podman version, using a Go template.
         runtime = Container.get_runtime_name()
-        cmd = [runtime, "version", "-f", "{{.Client.Version}}"]
+        if runtime == "podman":
+            query = "{{.Client.Version}}"
+        else:
+            query = "{{.Server.Version}}"
```
almet, line 62: At least on my machine, `podman version -f "{{.Client.Version}}"` returns the same than `podman version -f "{{.Server.Version}}"`.

I might not understand why we check for the "Server" version for Docker and the "Client" version for podman.

almet (approved):
I added a quick comment about the Server VS Client versions, but other than that it seems fine to me.

dangerzone/isolation_provider/container.py
apyrgio: I'm kind of hesitant to use `{{.Server.Version}}` for Podman, since it's not part of the JSON output:

```
$ podman version -f json | python3 -m json.tool
{
    "Client": {
        "APIVersion": "5.1.1",
        "Version": "5.1.1",
        "GoVersion": "go1.22.3",
        "GitCommit": "bda6eb03dcbcf12a5b7ae004c1240e38dd056d24-dirty",
        "BuiltTime": "Wed Jun  5 01:12:10 2024",
        "Built": 1717539130,
        "OsArch": "linux/amd64",
        "Os": "linux"
    }
}
```

I guess they internally satisfy this version check to match Docker's but I'm not sure if we can rely on it forever.

PR #851, opened 2024-06-26
WIP: Install PyMuPDF via the prebuilt Python wheels

### LINKED ISSUE #850: Install PyMuPDF via the prebuilt Python wheels for musl
Starting from PyMuPDF 1.24.6, the developers have [added](https://github.com/pymupdf/PyMuPDF/releases/tag/1.24.6) support for prebuilt musl-based wheels:

> Add musllinux x86_64 wheels to release.

This change can significantly improve our Dockerfile, and specifically these two lines:

https://github.com/freedomofpress/dangerzone/blob/e7e3430ca11ef768485b2d27e069ea76e9923fec/Dockerfile#L10-L11

Instead of installing build tools (`g++`, `make`, etc.) and building PyMuPDF from source, `pip install` will install the necessary wheels. This will greatly improve the build times of our container image.

There's one problem though: There is still no support for prebuilt musl-based wheels for aarch64 architectures. Since our Dockerfile is the same between our two architectures, we can't easily do different actions per architecture. For example, the pre-built wheels also create the following path: `/usr/lib/python3.12/site-packages/PyMuPDFb.libs/`. The regular wheels do not include this path, and therefore our Dockerfile would have to conditionally copy it, depending on the architecture:

https://github.com/freedomofpress/dangerzone/blob/e7e3430ca11ef768485b2d27e069ea76e9923fec/Dockerfile#L64-L65

This is still an open problem, and we have to wait until the PyMuPDF devs have a solution for it.

apyrgio:
Change our Dockerfile and image build script to install PyMuPDF from the prebuilt Python wheels for musl.

Note that this work is incomplete, because it doesn't handle the case of aarch64 wheels, which are still not prebuilt for musl.

Closes #850

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,8 @@ FROM alpine:latest as pymupdf-build
 ARG REQUIREMENTS_TXT
 
 # Install PyMuPDF via hash-checked requirements file
+RUN apk --no-cache add py3-pip
 COPY ${REQUIREMENTS_TXT} /tmp/requirements.txt
-RUN apk --no-cache add linux-headers g++ linux-headers gcc make python3-dev py3-pip clang-dev
 RUN pip install -vv --break-system-packages --require-hashes -r /tmp/requirements.txt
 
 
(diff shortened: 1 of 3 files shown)


apyrgio:
Closing this issue in favor of https://github.com/freedomofpress/dangerzone/pull/902

PR #855, opened 2024-07-01
container: Handle case where `docker kill` hangs

### LINKED ISSUE #854: The `docker kill` command may hang during container termination
During container termination, especially when we want to terminate the conversion on the client-side, we may attempt to kill the container with the `docker|podman kill` command first:

https://github.com/freedomofpress/dangerzone/blob/4ea0650f42ba8a0de556ef63184102fb26aa5e68/dangerzone/isolation_provider/container.py#L304-L317

Then, once this command has returned, we check the status of the conversion process, and may kill it forcefully:

https://github.com/freedomofpress/dangerzone/blob/4ea0650f42ba8a0de556ef63184102fb26aa5e68/dangerzone/isolation_provider/base.py#L233-L250

The problem here is that we have seen that for certain files (e.g., `no-pages-types.pdf` from our [test set](https://github.com/freedomofpress/dangerzone-test-set)) the `docker kill` command may never return. We don't have an explanation for this, but it seems to be a limitation from the Docker's side.

In order to circumvent this issue, we have to add a timeout to the `docker|podman kill` command, and move forward with the rest of our termination logic, if the timeout expires.

apyrgio:
We have encountered several conversions where the `docker kill` command hangs.  Handle this case by specifying a timeout to this command. If the timeout expires, log a warning and proceed with the rest of the termination logic (i.e., kill the conversion process).

Fixes #854

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -15,6 +15,9 @@ from ..util import get_tmp_dir  # NOQA : required for mocking in our tests.
 from ..util import get_resource_path, get_subprocess_startupinfo
 from .base import PIXELS_TO_PDF_LOG_END, PIXELS_TO_PDF_LOG_START, IsolationProvider
 
+TIMEOUT_KILL = 5  # Timeout in seconds until the kill command returns.
+
+
 # Define startupinfo for subprocesses
 if platform.system() == "Windows":
     startupinfo = subprocess.STARTUPINFO()  # type: ignore [attr-defined]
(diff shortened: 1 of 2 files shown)


almet:
Hey,

The linter just wants some spacing to be less stressed, let's give it that.

Other that that, it looks good to me, feel free to merge!

apyrgio:
Yeap, I've just pushed a fixup commit for that (was waiting for the CircleCI runner to cache the Dangerzone image).

almet (approved):

PR #858, opened 2024-07-05
Ignore CVE-2024-5535 from our security scans

apyrgio:
We believe that Dangerzone is not affected by CVE-2024-5535 for the following reasons:

1. This CVE affects applications that make network calls. The Dangerzone container does not perform any such calls, and has no access to the internet.
2. The OpenSSL devs have marked this issue as low severity.


almet (approved):
Thanks for the torough comment, I like how this is documented. Let's merge this!

Issue #773, opened 2024-04-09
Replace stdeb with a different way to produce Debian packages


apyrgio:
Dangerzone has been using [stdeb](https://github.com/astraw/stdeb/) since the very beginning (early 2020). At that time, stdeb was still maintained, but this is not the case anymore. It hasn't seen a new release since October 2020 (see the [PyPI releases](https://pypi.org/project/stdeb/#history) page). Also, it seems to not work with Python 3.12, since it throws this error:

```python
Traceback (most recent call last):
  File "/root/project/setup.py", line 34, in <module>
    setuptools.setup(
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 107, in setup
    return distutils.core.setup(**attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 185, in setup
    return run_commands(dist)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 1233, in run_command
    super().run_command(command)
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 988, in run_command
    cmd_obj.run()
  File "/usr/lib/python3/dist-packages/stdeb/command/sdist_dsc.py", line 30, in run
    debinfo = self.get_debinfo()
              ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/stdeb/command/common.py", line 197, in get_debinfo
    debinfo = DebianInfo(
              ^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/stdeb/util.py", line 802, in __init__
    check_cfg_files(cfg_files, module_name)
  File "/usr/lib/python3/dist-packages/stdeb/util.py", line 733, in check_cfg_files
    cfg = ConfigParser.SafeConfigParser()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'configparser' has no attribute 'SafeConfigParser'. Did you mean: 'RawConfigParser'?
```

[...]

PR #860, opened 2024-07-08
ci: Disable Debian Trixie builds

apyrgio:
Disable building packages in Debian Trixie, since it's Python version has changed to 3.12, which is not compatible with `stdeb`.

Refs #773


almet:
Thanks, let's disable them for now. I believe it will put some more pressure on the `.deb` packaging situation to replace `stdeb`.

PR #869, opened 2024-07-16
dev_scripts: Download FPF's PySide6 RPM only for Fedora 39

apyrgio:
Download the FPF-maintained python3-pyside6 RPM [1] only when we build an end-user environment for Fedora 39. Else, from Fedora 40 onwards, we can use the official `python3-pyside6` RPM.

Refs freedomofpress/maint-dangerzone-pyside6#5

[1] https://packages.freedom.press/yum-tools-prod/dangerzone/f39/python3-pyside6-6.7.1-1.fc39.x86_64.rpm

### DIFF
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -638,26 +638,27 @@ class Env:
             package_dst = build_dir / package
             install_cmd = "dnf install -y"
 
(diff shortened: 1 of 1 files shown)


legoktm (approved):
🎉

PR #875, opened 2024-07-23
Various improvements in our release instructions and scripts

apyrgio:
Make the following improvements to our release instructions and scripts:
1. Add a way to check if the installed Python version is the expected one.
2. Install the Thunar file manager in our Dangerzone container environments.
3. Do not use the builder cache by default, when building the Dangerzone container image. Add instead the `--use-cache` flag, when the developer wants to expedite the process for non-release builds.


legoktm (approved):
LGTM

apyrgio:
For this particular change, I'll wait for our CI to become green again before I merge. Thanks for the review Kunal!

Issue #878, opened 2024-07-25
Handle cases when LibreOffice hangs


apyrgio:
When running Dangerzone against our large test set, we found that some files (e.g.,  [`fdo78883.docx`](https://github.com/freedomofpress/dangerzone-test-set/blob/main/all_documents/fdo78883.docx) and [`ofz21168-1.doc`](https://github.com/freedomofpress/dangerzone-test-set/blob/main/all_documents/ofz21168-1.doc)) make LibreOffice 7.6 hang.

We opened a [bug report](https://bugs.documentfoundation.org/show_bug.cgi?id=162057) for these files, but until the underlying issue is solved, we need a way to detect such hangs, and stop the conversion.

apyrgio:
Re-introducing timeouts for the **whole document** is a solution I'd personally like to avoid. They have bitten us a lot in the past (#749), they are arbitrary (documents with many pages lead to very large timeout times), and we have recently decided to ditch them altogether (#687).

What makes more sense to me is the following:
* Get the number of pages in the document. Here's a way to do so: https://askubuntu.com/questions/305633/how-can-i-determine-the-page-count-of-odt-doc-docx-and-other-office-documents
* Spin up an UNO server (https://github.com/unoconv/unoserver). This server is responsible for loading a document, and listening for API requests.
* Send API requests to the UNO server, and ask to convert the document a **single page** at a time.
   - See some [supported export parameters](https://help.libreoffice.org/latest/en-US/text/shared/guide/pdf_params.html?DbPAR=SHARED) that LibreOffice offers by default, and the [blog post](https://vmiklos.hu/blog/pdf-convert-to.html) by the LibreOffice dev that added those. The `PageRange` option is of interest to us.
    - The UNO server project provides a command-line client ([`unoconvert`](https://github.com/unoconv/unoserver?tab=readme-ov-file#unoconvert)), but maybe we can send these API requests programmatically.
* Set a timeout for each API request. Since at the API request level we know we're dealing with a signle document page, we can set a sensible timeout (e.g., 3 minutes). Anecdotally, converting a .docx of ~2000 pages took in my laptop 18 minutes, so this timeout is more than reasonable.

Some extra benefits of this approach:
* `unoserver` has an option to send files back as binary data, instead of writing them to the filesystem. This will help with #633.
[...]

PR #879, opened 2024-07-25
Make PyMuPDF always log to stderr

### LINKED ISSUE #877: Make PyMuPDF always log to stderr
# Problem

PyMuPDF has made the design decision to log all messages by default to stdout (see https://github.com/pymupdf/PyMuPDF/issues/3135). This is a problem for us for two reasons:

First, in the document to pixels phase, we assume that the stdout of the conversion process contains only the pixels of the rasterized document (plus some page/size info). If a library that we use also logs to stdout, then this will corrupt our pixel stream, and our conversion will fail.

This is actually biting us in practice. We have a malformed file ([`no-pages-types.pdf`](https://github.com/freedomofpress/dangerzone-test-set/blob/main/all_documents/no-pages-types.pdf)) in our [large test set](https://github.com/freedomofpress/dangerzone-test-set/) that causes PyMuPDF to log this error to stdout:
    
```
MuPDF error: format error: non-page object in page tree
```

> [!TIP]
> You can reproduce it with:
> ```
> $ python3 -c 'import fitz; list(fitz.Document("no-pages-types.pdf").pages())' 2>/dev/null   
> ```

The error above is mixed with the pixel stream, and Dangerzone fails.

Second, in the pixels to PDF phase, we had a case (https://github.com/freedomofpress/dangerzone/issues/700) where PyMuPDF was logging to stdout, from which we expect to read conversion progress reports in JSON format. We circumvented this using Python's `contextlib.redirect_stdout()`, but we lose error messages this way.

# Solution

A solution to the above problem that works for all configurations is to take advantage of the environment variables that PyMuPDF devs have [added](https://github.com/pymupdf/PyMuPDF/issues/3135#issuecomment-1992625724) to control logging. An important detail here is that we have to use these environment variables in our `dangerzone.conversion.*` modules, before we load the `fitz` module, which is ugly, but doable.

apyrgio:
PyMUPDF logs to stdout by default, which is problematic because we use the stdout of the conversion process to read the pixel stream of a document.

Make PyMuPDF always log to stderr, by setting the following environment variables: `PYMUPDF_MESSAGE` and `PYMUPDF_LOG`.

Also, set a 5 minute timeout for document processing, when we run our large tests.
    
Fixes #877
Refs #878

### DIFF
--- a/dangerzone/conversion/doc_to_pixels.py
+++ b/dangerzone/conversion/doc_to_pixels.py
@@ -3,6 +3,17 @@ import os
 import sys
 from typing import Dict, Optional
 
+# XXX: PyMUPDF logs to stdout by default [1]. The PyMuPDF devs provide a way [2] to log to
+# stderr, but it's based on environment variables. These envvars are consulted at import
+# time [3], so we have to set them here, before we import `fitz`.
+#
+# [1] https://github.com/freedomofpress/dangerzone/issues/877
+# [2] https://github.com/pymupdf/PyMuPDF/issues/3135#issuecomment-1992625724
+# [3] https://github.com/pymupdf/PyMuPDF/blob/9717935eeb2d50d15440d62575878214226795f9/src/__init__.py#L62-L63
+os.environ["PYMUPDF_MESSAGE"] = "fd:2"
+os.environ["PYMUPDF_LOG"] = "fd:2"
+
+
 import fitz
 import magic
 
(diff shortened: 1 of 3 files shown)


legoktm (approved):
LGTM based on what upstream enabled, but setting environment variables at runtime to modify the environment is pretty hacky and it would be nice if they offered something better :/

apyrgio:
Yeap, it's too hacky for my liking, and I've opened an issue for that. Anw, we have to live with it...

Thanks for the review!

Issue #880, opened 2024-07-26
SELinux in enforcing mode breaks nested gVisor container


apyrgio:
# Problem

While debugging a Dangerzone 0.7.0 bug report on a Fedora 40 system (#871), we found out that SELinux in enforcing mode breaks the **inner** gVisor container that runs within the **outer** Podman container.

## Original error

```
$ dangerzone-cli ~/Downloads/sample.pdf
[...]
[DEBUG] Marking doc qyU8c4 as 'converting'
[INFO] > /usr/bin/podman run --log-driver none --security-opt no-new-privileges --cap-drop all --cap-add SYS_CHROOT --network=none -u dangerzone --rm -i --name dangerzone-doc-to-pixels-qyU8c4 --userns nomap dangerzone.rocks/dangerzone /usr/bin/python3 -m dangerzone.conversion.doc_to_pixels
[ERROR] [doc qyU8c4] 0% Unspecified error
[DEBUG] Marking doc qyU8c4 as 'failed'
```

## Output from Podman container with gVisor debug logging enabled

```
$ /usr/bin/podman run -e RUNSC_DEBUG=1 --log-driver none --security-opt no-new-privileges --cap-drop all --cap-add SYS_CHROOT --network=none -u dangerzone --rm -i --name dangerzone-doc-to-pixels-qyU8c4 --userns nomap dangerzone.rocks/dangerzone /usr/bin/python3 -m dangerzone.conversion.doc_to_pixels

8< ... snip ... >8

I0725 14:19:04.811776      20 main.go:197] **************** gVisor ****************
I0725 14:19:04.812445      20 boot.go:258] Setting product_name: "Standard PC (Q35 + ICH9, 2009)"
W0725 14:19:04.813081      20 specutils.go:129] noNewPrivileges ignored. PR_SET_NO_NEW_PRIVS is assumed to always be set.
I0725 14:19:04.813105      20 chroot.go:91] Setting up sandbox chroot in "/tmp"
W0725 14:19:04.813312      20 chroot.go:108] Failed to copy /etc/localtime: open /etc/localtime: no such file or directory. UTC timezone will be used.
I0725 14:19:04.813335      20 chroot.go:36] Mounting "/proc" at "/tmp/proc"
W0725 14:19:04.813394      20 util.go:64] FATAL ERROR: error setting up chroot: error remounting chroot in read-only: permission denied
error setting up chroot: error remounting chroot in read-only: permission denied
[...]

PR #882, opened 2024-07-30
dev_scripts: Handle Dangerzone packages with patch level != 1

apyrgio:
Update our `env.py` script to auto-detect the correct Dangerzone package name. This is useful when building an end-user environment, i.e., a container image where we copy the respective Dangerzone .deb/.rpm package and install it via a package manager.

To achieve this, we replace the hardcoded patch level (`-1`) in the package name with a glob character (`*`). Then, we check in the respective build directory if there's exactly one match for this pattern. If yes, we return the full path. If not, we raise an exception.

Note that this limitation was triggered when we were building RPM packages for the 0.7.0 hotfix release.

Refs #880

### DIFF
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -411,6 +411,28 @@ class Env:
         """Create an Env class from CLI arguments"""
         return cls(distro=args.distro, version=args.version, runtime=args.runtime)
 
+    def find_dz_package(self, path, pattern):
+        """Get the full path of the Dangerzone package in the specified dir.
+
+        There are times where we don't know the exact name of the Dangerzone package
+        that we've built, e.g., because its patch level may have changed.
+
+        Auto-detect the Dangerzone package based on a pattern that a user has provided,
+        and fail if there are none, or multiple matches. If there's a single match, then
+        return the full path for the package.
+        """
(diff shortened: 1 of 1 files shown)


rocodes:
An additional comment: When reviewing https://github.com/freedomofpress/yum-tools-prod/pull/25, I understood that this wasn't just a failing CI check that was blocking package builds, but that it's a preflight release check for built artifacts as well.

https://github.com/freedomofpress/yum-tools-prod/blob/97aed85442670aa4e91b504631c1f2f0a279b105/.github/workflows/check-packages.yml#L49-L52 

In that case, I would say that we should clarify that it's also used there, and I would be less inclined to be lax about version globbing - I think it's a fair restriction for pre-deployment CI to require an exact package version match somewhere.  But it's not something I want to block anyone on for this specific release.

apyrgio:
Fair point, I was debating whether we should a `--patch-level` argument as well. I'll open an issue for this, and we can reconsider this stance. For the time being, I'll merge this PR to unblock the `yum-tools-prod` one.

Thanks a lot for the thorough review, and I'd be happy to hear how we can improve this (admittedly overly complex) script.

PR #896, opened 2024-08-08
Ignore CVE-2024-5175 from our security scans

### LINKED ISSUE #895: Ignore CVE-2024-5171 from security scans
Our security scanner has [detected](https://github.com/freedomofpress/dangerzone/security/code-scanning/483) that the `aom-libs` package that we ship in our container has a critical vulnerability (CVE-2024-5171).

## Who uses this library?

This is a library that renders AV1 codecs. From the original [Chromium thread](https://issues.chromium.org/issues/332382766), it seems that the vulnerable function calls are:
* `aom_img_alloc`
* `aom_img_alloc_with_cb`
* `aom_img_wrap`
* `aom_img_alloc_with_border`

In our case, it seems that a GStreamer library (`libgstaom.so`) does use these function calls:

```console
$ readelf -Ws /usr/lib/gstreamer-1.0/libgstaom.so | grep aom_img
    26: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND aom_img_free
    41: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND aom_img_alloc
```

More specifically, it uses the vulnerable `aom_img_alloc` function in the **encode** path (see  [here](https://github.com/GStreamer/gst-plugins-bad/blob/ca8068c6d793d7aaa6f2e2cc6324fdedfe2f33fa/ext/aom/gstav1enc.c#L793)).

LibreOffice is a consumer of GStreamer, and therefore uses this library when converting documents to PDF:

> On the LibreOffice side all we require is to be able to extract a key frame from the video to provide a preview image, so e.g. on Linux the support depends on what gstreamer plugins you have installed.

_(taken from https://vmiklos.hu/blog/pdf-video-export.html)_

## Is Dangerzone affected?

Dangerzone is not affected.

GStreamer uses the vulnerable `aom_img_alloc` function in the `gstav1enc.c` file, which implements the **encoding** of videos to AV1 streams.  On the other hand, LibreOffice needs to **decode** the AV1 stream in order to generate a thumbnail. So, their paths don't cross.

We can verify this by invoking LibreOffice with the following [GStreamer envvars](https://gstreamer.freedesktop.org/documentation/gstreamer/running.html.):

```console
$ GST_DEBUG=*:6 libreoffice --headless --convert-to pdf --outdir . test.odp
[...]
0:00:00.192904920 359295 0x7525f0002690 LOG         codecparsers_av1 gstav1parser.c:819:gst_av1_parser_identify_one_obu: identify obu type is 6                                                                                               
0:00:00.192912986 359295 0x7525f0002690 LOG                 av1parse gstav1parse.c:1430:gst_av1_parse_handle_one_obu:<av1parse1> parsing the obu frame, result is 0
0:00:00.193099180 359295 0x7525f0002690 DEBUG               av1parse gstav1parse.c:744:gst_av1_parse_update_src_caps:<av1parse1> Update src caps video/x-av1, stream-format=(string)obu-stream, alignment=(string)tu, codec_data=(buffer)81040c000a0b00000024cf7f0dbfff3008, width=(int)960, height=(int)540, framerate=(fraction)25/1, pixel-aspect-ratio=(fraction)1/1, chroma-format=(string)4:2:0, bit-depth-luma=(uint)8, bit-depth-chroma=(uint)8, parsed=(boolean)true, profile=(string)main
0:00:00.195891100 359295 0x7525f0002690 DEBUG                 av1dec gstav1dec.c:142:gst_av1_dec_init:<GstAV1Dec@0x7525b400ead0> gst_av1_dec_init
[...]
```

We see that `gstav1enc.c` is not present in the logs. Instead, we see logs from `gstav1parse,c`, `gstav1parser.c` and `gstav1dec.c`, which belong to the **decode** path.

## Next steps

We believe that we can ignore CVE-2024-5171 from our security scans, by adding a reference in `.grype.yaml`.

apyrgio:
Ignore CVE-2024-5175 from our security scans, because Dangerzone is not affected by it. Our assessment follows:

The affected library, `libaom.so`, is linked by GStreamer's `libgstaom.so` library. The vulnerable `aom_img_alloc` function is only used when **encoding** a video to AV1. LibreOffce uses the **decode** path instead, when generating thumbnails.

Closes #895

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -43,3 +43,16 @@ ignore:
   #      > typically be under attacker control making active exploitation
   #      > unlikely.
   - vulnerability: CVE-2024-5535
+  # CVE-2024-5171
+  # =============
+  #
+  # NVD Entry: https://nvd.nist.gov/vuln/detail/CVE-2024-5171
+  # Verdict: Dangerzone is not affected. The rationale is the following:
+  #
+  # The affected library, `libaom.so`, is linked by GStreamer's `libgstaom.so`
+  # library. The vulnerable `aom_img_alloc` function is only used when
+  # **encoding** a video to AV1. LibreOffce uses the **decode** path instead,
+  # when generating thumbnails.
+  #
+  # See also: https://github.com/freedomofpress/dangerzone/issues/895
+  - vulnerability: CVE-2024-5171
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -2,46 +2,6 @@
 # latest release of Dangerzone, and offer our analysis.
 
 ignore:
-  # CVE-2023-7104
-  # =============
-  #
-  # NVD Entry: https://nvd.nist.gov/vuln/detail/CVE-2023-7104
-  # Verdict: Dangerzone is not affected. The rationale is the following:
-  #
-  #   1. This CVE affects malicious/corrupted SQLite DBs.
-  #   2. Databases can be loaded either via LibreOffice Calc or Base. Files for
-  #      the latter are not a valid input to Dangerzone.
-  #   3. Based on the LibreOffice Calc guide [1], users can only refer to
-  #      external databases, not embed them in a spreadsheet.
-  #   4. The actual CVSS score for this vulnerability is High, according to
-  #      NIST, not Critical.
-  #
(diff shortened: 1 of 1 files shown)


legoktm (approved):
LGTM

Issue #226, opened 2022-10-17
Defense in Depth - SELinux


apyrgio:
Parent issue: https://github.com/freedomofpress/dangerzone/issues/221

If an attacker manages to escape the container, we want to restrict them as much as possible. One way to do so is by using SELinux, which is also mentioned in the CIS Docker advisories ([official link](https://www.cisecurity.org/benchmark/docker), [mirror](https://edu.anarcho-copy.org/GNU%20Linux%20-%20Unix-Like/Docker/CIS_Docker_Benchmark_v1.3.1_PDF.pdf)).

There are several issues that make the adoption of an SELinux policy by Dangerzone a bit difficult though:
1. Generating an SELinux policy is notoriously difficult.
2. Enforcing an SELinux policy typically requires using Fedora/CentOS hosts, which is not the case for Docker Desktop on Windows/MacOS, which uses Alpine Linux (see https://gitlab.alpinelinux.org/alpine/aports/-/issues/8606)

Once we circumvent these issues, we should check if Dangerzone can take advantage of enforcing SELinux on the host.

deeplow:
A user reported [here](https://github.com/freedomofpress/dangerzone/issues/157#issuecomment-1416668001) having issues running dangerzone on a SELinux-enabled system.

apyrgio:
As part of fixing https://github.com/freedomofpress/dangerzone/issues/880, we apply the `container_engine_t` label to the **outer** container. This action is basically a no-op for most hosts, except those that use SELinux in enforcing mode.

Ultimately, if a Linux host does not use SELinux, there's no way for Dangerzone to "enable" it. This is an administrative action, and one that may break the workflow of lots of users. As for Windows and macOS, the same issue applies, plus the fact that we have even less control on the Linux VM.

Given that with gVisor, the **inner** container does not have the capability to open files (due to gVisor's strict seccomp filter), we believe that we have a similar protection as SELinux, in the rare case of container escapes. If we also consider that Dangerzone can utilize SELinux in enforcing mode, then we think it's safe to close this issue.

EtiennePerot:
> Given that with gVisor, the **inner** container does not have the capability to open files (due to gVisor's strict seccomp filter)

[...]

PR #898, opened 2024-08-10
Disable gVisor's DirectFS feature

EtiennePerot:
DirectFS is enabled by default in gVisor to improve I/O performance, but comes at the cost of enabling the `openat(2)` syscall (with [severe restrictions](https://gvisor.dev/blog/2023/06/27/directfs/), but still). As Dangerzone is not performance-sensitive, and that it is desirable to guarantee for the document conversion process to not open any files (to mimic some of what SELinux provides), might as well disable it by default.

See #226.

### DIFF
--- a/dangerzone/gvisor_wrapper/entrypoint.py
+++ b/dangerzone/gvisor_wrapper/entrypoint.py
@@ -142,6 +142,9 @@ runsc_argv = [
     "--rootless=true",
     "--network=none",
     "--root=/home/dangerzone/.containers",
+    # Disable DirectFS for to make the seccomp filter even stricter,
+    # at some performance cost.
+    "--directfs=false",
 ]
 if os.environ.get("RUNSC_DEBUG"):
     runsc_argv += ["--debug=true", "--alsologtostderr=true"]


apyrgio (approved):

apyrgio:
Thanks for the catch Etienne. I had seen this feature in a gVisor blog post, but I didn't have time to look to deep into it. I've taken a look at the rest of the [gVisor flags](https://github.com/google/gvisor/blob/e0643b8ed582cc549272e7788860a5dd4636c06d/runsc/config/flags.go#L35), just to make sure we don't miss an important option, but I don't see something else we can tweak.

I'll merge this PR as soon as I run some benchmarks to ensure that the performance impact is indeed minimal for our use case (I don't expect something different to be honest).

apyrgio:
I managed to run a quick benchmark for this PR. I ran our test suite on our 26 sample documents and took the following measurements:
* Elapsed time (in seconds) with DirectFS **disabled** (3 runs)
* Elapsed time (in seconds) with DirectFS **enabled** (3 runs)

The command I used is:

```
time poetry run pytest -v tests/test_cli.py::TestCliConversion::test_formats
```


Here are the results:

|            | Without DirectFS | With DirectFS |
| ---------|--------------------------| -------------------- |
| Run 1 |84.7s                      | 81.7s |
| Run 2 | 84.8s                     | 80.8s |
| Run 3 | 84.0s                     | 80.6s |
| Average | 84.5s                | 81.0s |
| stdev | 0.36s                    | 0.48s |

For this particular dataset, we notice a 5.3% speedup with DirectFS disabled. Because the second container is involved in this benchmark, which doesn't use gVisor, we can assume that the speedup is even larger. This would match the 12% speedup in the workloads that were measured in the [gVisor blog post](https://gvisor.dev/blog/2023/06/27/directfs/).

I think we can live with this performance penalty, if it means more security benefits in the case of a container escape.

apyrgio:
Oh, btw, tested on Docker Desktop on macOS and it works as well. That's to be expected, since we remove syscalls, not add them.

Issue #745, opened 2024-03-07
Research/spec for independent container updates


harrislapiroff:
#698 is a feature with potential for a lot of unforeseen issues in implementation. Let's write up a document the explores proposed architecture, UI, security concerns, etc. before we get to the work of implementing it.

apyrgio:
Adding a comment real quick. We had a discussion with Alexis, and we agreed that the way we use image IDs now is a bit problematic, given some [recent developments](https://github.com/freedomofpress/dangerzone/issues/933). It's worth reconsidering them here as well.

apyrgio:
We have internally researched this subject, and we have an overview of how we want to implement this feature in https://github.com/freedomofpress/dangerzone/issues/1006. More concrete implementation details will be added to that issue. Closing this one.

PR #901, opened 2024-08-28
Replace stdeb in favor of modern Debian packaging tools

### LINKED ISSUE #394: Notice when installing from packages.freedom.press
Running `sudo apt update` on Ubuntu 20.04 currently displays the following notice with our recommended packaging configuration:

```
N: Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'https://packages.freedom.press/apt-tools-prod focal InRelease' doesn't support architecture 'i386'
```

### LINKED ISSUE #621: Instruction to add repository in Ubuntu 23.10, missing arch parameter
After apt update, the error message is displayed about i386

`N: Le fichier configuré « main/binary-i386/Packages » ne sera pas pris en compte car le dépôt « https://packages.freedom.press/apt-tools-prod mantic InRelease » ne prend pas en charge l'architecture « i386 »`

missing "arch=amd64"

```
source /etc/os-release
echo deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg] \
    https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main \
    | sudo tee /etc/apt/sources.list.d/fpf-apt-tools.list

### LINKED ISSUE #773: Replace stdeb with a different way to produce Debian packages

### PR DESCRIPTION
**Pull request by almet**
This removes stdeb in favor of more modern Debian packaging tools (see #773 for the rationale).

As a result, we can build `.deb` packages on debian trixie.

I've tested manually that the produced debian image is working for the following distributions :

- ubuntu-22.04
- ubuntu-23.10
- ubuntu-24.04
- debian-bullseye
- debian-bookworm
- debian-trixie

 ubuntu-20.04 is currently failing because of the dependency to podman that it can't install. I'm not sure yet why.

Fixes #773
Fixes #394
Fixes #621

### CODE REVIEW

File: debian/changelog [hunk 1]
```diff
@@ -0,0 +1,5 @@
+dangerzone (0.7.0) bionic; urgency=low
```
almet, line 1: I've committed the file produced by `stdeb` at first, but I'm not sure why "bionic" is mentioned here, and if we should keep this changelog as is.

debian/compat
```diff
@@ -0,0 +1 @@
+10
```
almet, line 1: The compat level was set to 9 when generated by stdeb, I took the liberty to bump it to see if it works to silence deprecation warnings.

install/linux/build-deb.py
```diff
@@ -2,19 +2,15 @@
 # -*- coding: utf-8 -*-
```
almet, line 2: This whole file could potentially be replaced by a few commands. Maybe we should do it? :roll_eyes: 

```bash
rm deb_dist && rm dist
dpkg-buildpackage
mkdir deb_dist
mv ../*.deb deb_dist
```

debian/control
```diff
@@ -0,0 +1,16 @@
+Source: dangerzone
+Maintainer: Freedom of the Press Foundation <info@freedom.press>
+Section: python
+Priority: optional
+Build-Depends: dh-python, python3-setuptools, python3-all, debhelper (>= 9), dh-python, python3, python3-setuptools
```
legoktm, line 5: * dh-python and python3-setuptools are listed twice
* you don't need both python3-all and python3, just the former is fine

debian/rules
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/make -f
+
+# This file was automatically generated by stdeb 0.10.0 at
```
legoktm, line 3: You can delete these comments :)

debian/changelog
legoktm: It's supposed to be what distribution you're targeting. For our purposes it doesn't really matter, because our repository software ignores this field (well, we tell it too). In SecureDrop Client we just write "unstable" and [have logic](https://github.com/freedomofpress/securedrop-client/blob/main/scripts/fixup-changelog.sh) to embed the distro name in the version. But I don't think you need that since y'all use one deb across all distro versions.

legoktm (commented):
Nice!!! Few notes inline.

You don't need an empty `debian/install` file, if there's nothing else to add into the package, just delete it.

debian/changelog
almet: Thanks, I've set it to "unstable" as well then. Yep for now it doesn't seem needed to generate different .deb files

almet:
Thanks for the comments. I've added a few fixup commits addressing them.

almet:
> ubuntu-20.04 is currently failing because of the dependency to podman that it can't install. I'm not sure yet why.

This seems in fact expected, and we have [a script](https://github.com/freedomofpress/dangerzone/blob/main/install/linux/install-podman-ubuntu-focal.sh) to install podman on focal

almet:
I've been curious if the current `.deb` file we provide to our debian / ubuntu users requires podman, and it does:

```
Depends: <snip> podman <snip>
```

I guess it "just" needs to be installed for this to work properly (?)

almet:
Actually, the reason for the failure was that we were receiving a `503` error when installing podman from opensuse repositories, so completely orthogonal to our changes here.

debian/source/options
```diff
@@ -0,0 +1,4 @@
+compression = "gzip"
```
apyrgio, line 1: I think we have a great opportunity to significantly reduce our build times here.

The main reason our .deb (and .rpm) builds are taking a long time is because they try to compress all files. And by all files, I also mean the `container.tar.gz` file, which is already compressed at the highest level. This file is basically responsible for the inflated size of the Debian package, and we don't gain something by re-compressing it.

My suggestion is to add here a `compression-level = "fast"` option (taken from https://manpages.debian.org/stretch/dpkg-dev/dpkg-source.1.en.html#Generic_build_options). I believe the change in size will be minimal (~1-2 MBs), but the change in build time will be massive. If this holds, we can also do something similar for RPM packages in a subsequent PR.

install/linux/build-deb.py
apyrgio: Hm, that would be a pass from me. If we want to change this script later on (as we have done already with the `--distro` argument), it's best to have a script written in Python than one in Bash / Posix shell.

Also, we'd have to `set -e` in the above Shell script, and thus make the `rm` commands not fail, etc.

debian/control
```diff
@@ -0,0 +1,15 @@
+Source: dangerzone
+Maintainer: Freedom of the Press Foundation <info@freedom.press>
+Section: python
+Priority: optional
+Build-Depends: dh-python, python3-setuptools, python3-all, debhelper (>= 9)
```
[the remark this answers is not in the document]
almet: Good catch! Added in a fixup commit.

debian/control
```diff
@@ -0,0 +1,15 @@
...
+Standards-Version: 4.5.1
+Homepage: https://github.com/freedomofpress/dangerzone
+Rules-Requires-Root: no
+
+Package: dangerzone
+Architecture: all
```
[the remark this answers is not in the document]
almet: Thanks, this seem indeed to be the case. I pushed this in a fixup commit.

almet:
Thanks for your comments,

> First, it seems that `pybuild` attempts to find and clean `.pyc` files under the `dev_scripts` directory. This directory may not be empty if we have created dev environments for Dangerzone

You are correct, and it's annoying to me as well. It's actually debian helper who's using `find` to look for files, and doesn't get access to some of them, so it's bailing out. I'm trying to find a solution to not include the `dev_scripts` folder in the `find` commands, and also disable cleaning files within it.

> Second, if you notice in the above logs, it seems that `pybuild` attempts to use `setup.py` (`python3.11 setup.py clean`), instead of consulting `pyproject.toml`. 

I agree 100% with you that using `pyproject.toml` rather than `setup.py` is the path forward. I've been actually involved in the early discussions about distutils2 back in the days. It represents something to get rid of this `setup.py` file whenever I can. 

But... I'm unsure if we should be able to build the debian packages on all the platforms we support. If it's the case (and I expect it is), I believe it's better to wait until ubuntu focal is deprecated, like you mentioned.

Nevertheless, I will give a try to `pybuild-plugin-pyproject` and report my findings, probably in a separate issue.

debian/source/options
almet: :sad-trombone: Unfortunately, it actually doesn't seem to make much difference (on my machine at least) : 

|  | time (real) | time (user) | time (sys) | Package size |
|--------|--------|--------|--------|--------|
| `compression-level = "fast"` | 1m29.722s | 5m52.673s | 0m8.686s | 658M |
| Default value | 1m30.911s | 6m2.769s | 0m8.471s | 658M |

almet: They mention it in their docs actually: Note that this option will not cause existing tarballs to be recompressed, it only affects new files

debian/control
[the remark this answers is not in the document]
apyrgio: Right, and that's a good thing actually. Our apt-tools-prod repo is already configured as an `amd64` only repo, so I think all the pieces are there. I don't see something else we need to change.

debian/source/options
apyrgio: Sad trombone indeed. Thanks for taking a look at this Alexis.

_Weird, then what is it compressing for 6 minutes straight..._

apyrgio:
>  I'm trying to find a solution to not include the dev_scripts folder in the find commands, and also disable cleaning files within it.

Cool, cool. I think that if you enable `DH_VERBOSE=1`, you may get a bit more logs. I believe that the offending command is a `python setup.py clean` command, and I think we can pass arguments to that via `pybuild`.

> But... I'm unsure if we should be able to build the debian packages on all the platforms we support. If it's the case (and I expect it is), I believe it's better to wait until ubuntu focal is deprecated, like you mentioned.

I agree, ultimately that's the sensible thing to do, provided that we make the `setup.py` (`distutils`) method work. If not, then `pyproject` is back on the menu I guess...

almet:
> I believe that the offending command is a `python setup.py clean` command, and I think we can pass arguments to that via `pybuild`. Maybe we can even override the clean function from our own `setup.py`.

It's actually some of the deb helper `make` targets which contain some `find` commands which are trying to read inside some folders they don't have access to. The only way to tweak seems to be by overriding them using `*_override` targets in `d/rules`.

I'm still trying to find where the original ones are :eyes:

legoktm:
> The only drawback is that `pybuild-plugin-pyproject` is not available in Ubuntu Focal. In that case, we _may_ backtrack until Ubuntu Focal is deprecated.

You can probably try something like `Build-Depends: pybuild-plugin-project | python3`; see https://www.debian.org/doc/debian-policy/ch-relationships.html#relationships-between-source-and-binary-packages-build-depends-build-depends-indep-build-depends-arch-build-conflicts-build-conflicts-indep-build-conflicts-arch for the nuances on how alternative build dependencies work.

----

> The only way to tweak seems to be by overriding them using `*_override` targets in `d/rules`.

See https://manpages.debian.org/bookworm/debhelper/dh_auto_clean.1.en.html#DESCRIPTION, especially:
> This is intended to work for about 90% of packages. If it doesn't work, or tries to use the wrong clean target, you're encouraged to skip using dh_auto_clean at all, and just run make clean manually.

almet:
> You can probably try something like `Build-Depends: pybuild-plugin-project | python3`; 

Following this path would mean we would still need to rely on `setup.py` for ubuntu focal.

Rather than having to maintain multiple paths for this, it might be easier to continue relying on `setup.py` for the time being, and move to `pyproject.toml` + `pybuild-plugin-project` it as soon as we drop support for ubuntu focal, which will happen after 02 April 2025  (the end of Maintenance & Security Support).

I've opened an issue to follow this: #911

---

Some notes on the current progress:

1. I've removed `python3-all`, as it doesn't seem needed anymore to build the debian packages. It seems that the reason we needed it was due to [a `stdeb` bug](https://github.com/astraw/stdeb/issues/153).

2. I've removed `podman` from the build dependencies. It's actually a requirement to build the image, but not a requirement to build the packages themselves. Also, because we're currently [considering providing the images for the users to consume](#745 ), I'm not sure if it's worth adding the dependency now to remove it later on.

3. The debian package currently builds on all platforms :nerd_face:

.circleci/config.yml
```diff
@@ -5,25 +5,12 @@ aliases:
...
-      apt-get update
-      apt-get install -y python-all
-
   - &install-dependencies-deb
     name: Install dependencies (deb)
     command: |
       export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true
       apt-get update
-      apt-get install -y dh-python python3 python3-stdeb
+      apt-get install -y dh-python python3 python3-setuptools dpkg-dev debhelper
```
legoktm, line 13: If you're in the DZ checkout, you can just run `apt build-dep .` instead of needing to duplicate the Build-Requires here.

BUILD.md
```diff
@@ -59,7 +59,7 @@ Install dependencies:
 
 ```sh
 sudo apt install -y podman dh-python build-essential fakeroot make libqt6gui6 \
```
legoktm, line 61: fakeroot can be removed now that `Rules-Requires-Root: no` is set (unless something else needs it?)

apyrgio (approved):

apyrgio:
I don't have anything else to comment on. Kunal has some nice comments, but other than that, the end result looks great to me. Feel free to merge once Kunal's comment are addressed!

.circleci/config.yml
almet: Thanks! Just did.

BUILD.md
almet: Good call. :+1:

legoktm:
Woooo, congrats!!  🥳 🎆 🎊

PR #902, opened 2024-08-28
Use PyMuPDF wheels for non-ARM architectures.

almet:
This removes the need to build the PyMuPDF project by ourselves, but only when on non-ARM architectures since the wheels for these are not provided yet.

Changes the `Dockerfile` and `build-image.py` script, introducing a new `ARCH` flag to conditionally build the wheels.

Supersedes #851

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,23 @@
 # Build PyMuPDF
 
 FROM alpine:latest as pymupdf-build
-
+ARG ARCH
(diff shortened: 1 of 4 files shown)


apyrgio (approved):

apyrgio:
Nice, the intermediate image layers are even slimmer with the `ARCH` build argument. Let's merge it once the CI is green.

PR #904, opened 2024-08-29
Rename make lint-apply to make format

almet:
We are periodically running this command to get the CVEs on track, and it is easier to have it available as `make security-scan`

Note: maybe it's worth also removing the `/tmp/container.tar` before running the command?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ mypy: mypy-host mypy-tests ## check type hints with mypy
 lint: lint-black lint-isort mypy ## check the code with various linters
 
 .PHONY: lint-apply
-lint-apply: lint-black-apply lint-isort-apply ## apply all the linter's suggestions
+format: lint-black-apply lint-isort-apply ## apply all the linter's suggestions
 
 .PHONY: test
 test:


Makefile
```diff
@@ -66,6 +66,11 @@ test-large: test-large-init  ## Run large test set
 	python -m pytest --tb=no tests/test_large_set.py::TestLargeSet -v $(JUNIT_FLAGS) --junitxml=$(TEST_LARGE_RESULTS)
 	python $(TEST_LARGE_RESULTS)/report.py $(TEST_LARGE_RESULTS)
 
+security-scan: ## Look for CVEs using the grype tool
+	gunzip --keep -c ./share/container.tar.gz > /tmp/container.tar
+	docker pull anchore/grype:latest
+	docker run --rm -v /tmp/container.tar:/container.tar anchore/grype:latest /container.tar
+
```
apyrgio, line 73: I'm fine with merging this command make action, but I'm afraid it's a bit too strict to be generally useful:
1. It needs to take into account that we use `podman` in Linux systems (well, I happen to have Docker installed as well, but I don't really use it)
2. It needs to work on a file other than `share/container.tar.gz`. This file is usually the last tarball we've compiled in our repo, whereas we typically run this command on the latest **released** tarball (maybe somewhere in our Downloads folder, or some place else)

What do you think?

almet: Actually, I'm at the moment running the `grype` tool directly without using docker / podman at all; and as a result it's less useful to me to have this specific make target that I thought it would be.

Let's just close this (and just keep the `s/lint-apply/format` change.

apyrgio (approved):

apyrgio:
Thanks, feel free to merge :slightly_smiling_face:

legoktm:
(I didn't see this until it was merged so sorry for the late bikesheddy comment; in SecureDrop we've standardized on `make fix` - see https://github.com/freedomofpress/securedrop-tooling/issues/11; which covers non-formatting tools like Ruff and also is slightly shorter ;))

PR #906, opened 2024-08-30
Fix wrong container runtime detection on Linux

### LINKED ISSUE #212: Wrong container runtime detection on Linux
## Description

Running Dangerzone on a Linux environment that does not have Podman properly configured, throws an error about Docker not running:

> Dangerzone Requires Docker Desktop
>
> Docker is installed but isn't running.
>
> Open Docker and make sure it's running in the background.

 This is unexpected because the newest Dangerzone release uses Podman instead of Docker.

## Steps to Reproduce

OS: Ubuntu 22.04 (containerized via Docker)
Release: 0.3.2

* Create an Ubuntu container with no special privileges, but with X11 access.
* Install Dangerzone in it.
* Make sure that `podman images ls` fails.
* Run `dangerzone` within that container.
* See the wrong error message in the GUI.

## Expected Behavior

Inform the user that Podman is not properly configured. Ideally, explain what is the command that failed, either in the UI or in the logs.

almet:
This is still a work in progress, not ready to be merged.

Fixes #212

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -10,14 +10,18 @@ from typing import List, Optional
 # FIXME: See https://github.com/freedomofpress/dangerzone/issues/320 for more details.
 if typing.TYPE_CHECKING:
     from PySide2 import QtCore, QtGui, QtSvg, QtWidgets
-    from PySide2.QtWidgets import QAction
+    from PySide2.QtCore import Qt
+    from PySide2.QtWidgets import QAction, QTextEdit
 else:
     try:
         from PySide6 import QtCore, QtGui, QtSvg, QtWidgets
+        from PySide6.QtCore import Qt
         from PySide6.QtGui import QAction
+        from PySide6.QtWidgets import QTextEdit
     except ImportError:
         from PySide2 import QtCore, QtGui, QtSvg, QtWidgets
-        from PySide2.QtWidgets import QAction
+        from PySide2.QtWidgets import QAction, QTextEdit
+        from PySide2.QtCore import Qt
 
 from .. import errors
 from ..document import SAFE_EXTENSION, Document
@@ -402,6 +406,28 @@ class WaitingWidget(QtWidgets.QWidget):
         super(WaitingWidget, self).__init__()
 
 
+class TracebackWidget(QTextEdit):
+    """Reusable component to present tracebacks to the user.
+
+    By default, the widget is initialized but does not appear.
+    You need to call `.set_content("traceback")` on it so the
+    traceback is displayed.
+    """
+
+    def __init__(self) -> None:
+        super(TracebackWidget, self).__init__()
+        # Error
+        self.setReadOnly(True)
+        self.setVisible(False)
+        self.setProperty("style", "traceback")
+        # Enable copying
+        self.setTextInteractionFlags(Qt.TextSelectableByMouse)
+
+    def set_content(self, error: str) -> None:
+        self.setPlainText(error)
+        self.setVisible(True)
+
+
 class WaitingWidgetContainer(WaitingWidget):
     # These are the possible states that the WaitingWidget can show.
     #
@@ -434,10 +460,13 @@ class WaitingWidgetContainer(WaitingWidget):
         self.buttons = QtWidgets.QWidget()
         self.buttons.setLayout(buttons_layout)
 
+        self.traceback = TracebackWidget()
+
         # Layout
         layout = QtWidgets.QVBoxLayout()
         layout.addStretch()
         layout.addWidget(self.label)
+        layout.addWidget(self.traceback)
         layout.addStretch()
         layout.addWidget(self.buttons)
         layout.addStretch()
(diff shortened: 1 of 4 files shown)


almet:
Updated, this now looks like this:

image

This still needs to be more consistent between the different OSes, and probably display the error on other platforms.

dangerzone/gui/main_window.py
```diff
@@ -459,38 +482,75 @@ def check_state(self) -> None:
...
+            if platform.system() == "Linux":
+                message = (
+                    "<strong>Dangerzone requires Podman</strong><br><br>"
+                    "Podman is installed but isn't running.<br>"
+                    "It could mean that it is not configured properly."
+                )
+                if error:
+                    message += "<br><br>The following error was caught when trying to run <code>podman image list</code>"
+
+                    code_snippet = f"<pre>{error.decode()}</pre>"
```
EtiennePerot, line 533: This may need HTML escaping.

dangerzone/gui/main_window.py
```diff
@@ -459,38 +482,75 @@ def check_state(self) -> None:
...
+                )
             self.buttons.show()
         elif state == "not_running":
-            self.label.setText(
-                "<strong>Dangerzone Requires Docker Desktop</strong><br><br>Docker is installed but isn't running.<br><br>Open Docker and make sure it's running in the background."
-            )
+            if platform.system() == "Linux":
+                message = (
+                    "<strong>Dangerzone requires Podman</strong><br><br>"
+                    "Podman is installed but isn't running.<br>"
```
EtiennePerot, line 527: Perhaps this is nitpicking, but there is no such thing as "Podman is running" in the same sense as "Docker is running". Docker works with a daemon/client model, where `dockerd` runs persistently in the background and `docker` is a client to it. Podman by contrast has no background daemon. Every invocation of the `podman` command-line does everything it needs to do on the spot and then disappears with no persistent presence on the system.

This "`not_running`" status is derived from the failure of running `podman image list` above, which in this case would indicate that something is wrong about Podman's setup, but not that it is not running.

So I suggest changing this message to something like "Dangerzone requires Podman. Podman is installed but cannot run properly. See errors below."

almet: Hey @EtiennePerot, thanks for the comments.

You're correct and thanks for the explanation. That explains why trying to stop "a podman service" doesn't make the following podman commands fail.

I've updated the error message accordingly.

dangerzone/gui/main_window.py
almet: Thanks for the comment. The code now uses `setPlainText` instead, so no HTML is involved.

almet:
This now looks like this:

image

almet:
While reviewing the issues that [should be included in 0.8.0](https://github.com/orgs/freedomofpress/projects/11/views/16?query=sort%3Aupdated-desc+is%3Aopen), I found about https://github.com/freedomofpress/dangerzone/issues/193, which is about displaying errors to the end user. 

I'm now thinking that the code introduced here to display the logs might be put in a sort of component, to be reused on other parts of the client UI.

dangerzone/gui/main_window.py
```diff
@@ -402,6 +406,38 @@ def __init__(self) -> None:
...
+        self.setStyleSheet(
+            """
+        QTextEdit {
+            font-family: Consolas, Monospace;
+            font-size: 12px;
+            background-color: #fff;
+            color: #000;
+            padding: 10px;
+        }
+        """
```
apyrgio, line 430: Suggestion, let's move this to `dangerzone.css`, since it's kind of long and CSS related. A simple way to do this is to set a property to that widget:

```python
self.error_text.setProperty("style", "container_error")
```

and then add the following in `dangerzone.css`:

```diff
diff --git a/share/dangerzone.css b/share/dangerzone.css
index 45fdc2b..23f3bab 100644
--- a/share/dangerzone.css
+++ b/share/dangerzone.css
@@ -48,3 +48,11 @@ QLabel.version {
     font-size: 20px;
     padding-bottom: 5px;  /* align with 'dangerzone' font */
 }
+
+QTextEdit[style="container_error"] {
+    font-family: Consolas, Monospace;
+    font-size: 12px;
+    background-color: #ffffff;
+    color: #000000;
+    padding: 10px;
+}
```

I've tested already and it works.

dangerzone/gui/main_window.py
```diff
@@ -448,49 +487,83 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
...
+                    "Podman is installed but cannot run properly. See errors below"
+                )
+                if error:
+                    self.traceback.set_content(error)
+
+                self.label.setText(message)
+
+            else:
+                self.label.setText(
+                    (
```
apyrgio, line 554: The extra parentheses are not needed, I believe.

dangerzone/gui/main_window.py
```diff
@@ -448,49 +487,83 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
...
             self.buttons.show()
         else:
             self.label.setText(
-                "Installing the Dangerzone container image.<br><br>This might take a few minutes..."
+                (
+                    "Installing the Dangerzone container image.<br><br>"
+                    "This might take a few minutes..."
+                )
             )
             self.buttons.hide()
```
apyrgio, line 568: Besides hiding the buttons, we may need to hide the error box (`self.error_text`) as well. Else, if a user fixes a broken Podman installation, the error box will remain while the container image gets installed.

Piping the installation progress of the container image would be the ideal solution here, but it's an overkill for this feature.

apyrgio (commented):
Looks good. I have a few small suggestions, and then we're good to merge.

dangerzone/gui/main_window.py
almet: Thanks, removed.

dangerzone/gui/main_window.py
almet: Good catch, I've made the change.

dangerzone/gui/main_window.py
almet: Awesome, thanks for the comment. I've updated the code and used the "traceback" CSS class name instead.

apyrgio (approved):

apyrgio:
Thanks Alexis, feel free to merge, I guess once the CI is working once more.

PR #907, opened 2024-09-02
Switch from CircleCI runners to Github actions

### LINKED ISSUE #674: Migrate from CircleCI to GitHub actions

almet:
This is still pretty much a WIP. I'm pushing it here mainly to see how the CI is reacting.

Fixes #674


almet:
Here are the changes I did in this PR, which is up for discussion:

"dev" environments (`dangerzone.rocks/build/distro-version`) are now built daily, and pushed to the [github container registry](https://ghcr.io), so that the CI can reuse them during the day, without having to build them.

I'm currently using the ghcr, but I'm wondering if using a cache instead would be more beneficial, because we could invalidate the cache more easily, and rebuild the dev images on the fly if needed.

The dangerzone container image is also built daily, cached and uploaded to the ghcr.

"end user" (`dangerzone.rocks/distro-version`) images are currently not uploaded because they require a `.deb` or `.rpm` to be built, which is part of the CI pipeline.

While doing these changes, the `dangerzone.rocks` prefix shown to be cumbersome, as it is taking the place of the container registry. As a result, I now changed this prefix to `dangerzone` instead, only for the "dev" and "end-user" images.

We might want to follow the same lead for the main container image `dangerzone.rocks/dangerzone`.

With these changes, the CI is now taking ~9mn to complete all the tests! :racehorse:

almet:
I'm currently using the ghcr for the dev images, but I'm leaning towards getting it (almost) out of the picture, as I'm not sure it makes more sense than using a cache with `podman save` / `podman load -i`.

This last option will make it easier to invalidate the cache if for instance the dependencies change. But maybe I'm missing the main point why you wanted to have a look at the ghcr in the first place?

apyrgio:
Damn, super exciting work Alexis! I'll dig into it soon.

---

On the subject of caching, I'm wondering how is the invalidation story different for caches, instead of container images? I'm definitely missing context here since I have theorized about this working, but I didn't implement it as you did.

In any case, I suggested pushing to a container registry instead of creating a tarball for the following reasons:
1. We can re-use these container images in other CI actions (think `apt-tools-prod` / `yum-tools-prod`). We can even give them a check locally in case of a CI error, without waiting for image rebuilds.
2. Pushing the Dangerzone container image to GHCR may help with independent container updates.
3. It will speed up image uploads/downloads, in case some layers are already present (think the base Debian/Fedora layers, and maybe some extra ones if only a later portion of the Dockerfile changes)."
4. Artifacts like Tesseract data (soon to come) won't be evicted in case of lots of dev container images.

That's mostly it, but I'd like to know your experience so far.

.github/workflows/ci.yml
```diff
@@ -201,15 +283,88 @@ jobs:
...
+          if [ "${{ matrix.version }}" = "bullseye" ]; then
+            echo "modified_version=${{ matrix.version }}-backports" >> $GITHUB_ENV
+          else
+            echo "modified_version=${{ matrix.version }}" >> $GITHUB_ENV
+          fi
+
+      - name: Install dev env from GHCR
+        run: |
+          podman pull \
+            ${{ env.IMAGE_REGISTRY }}/dangerzone/build/${{ matrix.distro }}:${{ env.modified_version }}
```
apyrgio, line 349: You've probably have alluded to this already, but what happens here if the `env.py` script changes, and creates the dev environments slightly different for a PR?

I think we should hash the resources that shape our dev images (much as you have done already for the container image), and make this hash past of the container image tag. To make things simpler, I'd suggest adding a `--sync` flag in our `env.py` script, which will attempt to pull the image, build it if not already built, and then push it to GHCR.

.github/workflows/ci.yml
```diff
@@ -4,10 +4,64 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: debian:bookworm
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dev. dependencies
+        run: |-
+          apt-get update
+          apt-get install -y git make python3 python3-poetry --no-install-recommends
+          poetry install --no-ansi --only lint,test
```
jkarasti, line 36: Is the `--no-ansi` flag here needed anymore? The issue mentioned in the circleci config got fixed in cleo 2.1 which was included in poetry 1.7 released last November.

apyrgio: The quick answer here is "probably not". But I'm more interested in how you stumbled on this. Did it mess with the logs in your CI runs?

jkarasti: It didn't. Just happened to notice it while poking around the linters used in the Dangerzone repo.

Once I'm done with the windows installer, I'm planning on replacing black and isort with ruff for code formatting and linting as my next small project in the dangerzone codebase. Mostly since theres the issue about unused imports still open and I noticed Securedrop made the change a while ago and thought why not do that here too? A good excuse to play around with ruff, something I've wanted to do for a while now, and Dangerzone repo benefits from faster linting with potential to enable a ton of linting tools in one with little effort.

Though, first I'll open an issue proposing the change in a couple of days.

dev_scripts/env.py
```diff
@@ -575,8 +621,26 @@ def run(
...
-    def build_dev(self, show_dockerfile=DEFAULT_SHOW_DOCKERFILE):
+    def pull_image_from_registry(self, image):
+        process = subprocess.run(self.runtime_cmd + ["pull", IMAGES_REGISTRY + image])
+        return process.returncode == 0
+
+    def push_image_to_registry(self, image):
+        process = subprocess.run(
+            self.runtime_cmd + ["push", image, IMAGES_REGISTRY + image]
+        )
+        return process.returncode == 0
```
apyrgio, line 632: I'd probably add `check=True` here, and capture the stderr, to learn about errors.

.github/workflows/ci.yml
```diff
@@ -4,10 +4,64 @@ on:
...
+            share/image-id.txt
+      - name: Build Dangerzone image
+        run: |-
+          if [ -f "share/container.tar.gz" ]; then
+            echo "Already cached, skipping"
+          else
+            sudo pip3 install poetry
+            python3 ./install/common/build-image.py
+          fi
+
```
apyrgio, line 64: I propose we borrow the exact same logic as in `build.yml`, else changes in the container image (e.g. in a PR) will not be cached immediately.

.github/workflows/ci.yml
```diff
@@ -57,56 +126,58 @@ jobs:
...
       - name: Build Dangerzone .deb
         run: |
-          ./dev_scripts/env.py --distro ${{ env.distro }} \
-              --version ${{ env.version }} \
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} \
+              --version ${{ matrix.version }} \
               run --dev --no-gui ./dangerzone/install/linux/build-deb.py
 
       - name: Upload Dangerzone .deb
+        if: matrix.distro == 'debian' && matrix.version == 'trixie'
```
apyrgio, line 155: Nit, the reference distro where we build Debian packages is Debian bookworm:

```suggestion
        if: matrix.distro == 'debian' && matrix.version == 'bookworm'
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
+name: Build dev environments
+on:
+  push:
+  schedule:
+    - cron: "0 0 * * *" # Run every day at 00:00 UTC.
+env:
+  IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
+  REGISTRY_USER: ${{ github.actor }}
+  REGISTRY_PASSWORD: ${{ github.token }}
```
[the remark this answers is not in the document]
almet: Should be fixed by 6a441ea9c4b96c42f19e514516b906d4c4a006ed, thanks.

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Login to GHCR
+        run: |
+          echo ${{ github.token }} | podman login ghcr.io -u USERNAME --password-stdin
+
+      - name: Push To GHCR
+        run: |
+          podman push \
+            dangerzone.rocks/dangerzone \
+            ${{ env.IMAGE_REGISTRY }}/dangerzone/dangerzone
```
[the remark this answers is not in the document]
almet: Thanks, this is fixed in 4244af1d2054205f0600cbb1b80ce7ee71192036

.github/workflows/ci.yml
almet: Thanks for the remark, this is fixed in 510c169b0baa3ae4a3dc99d480c7836779c81c81

dev_scripts/env.py
```diff
@@ -56,8 +58,9 @@
 # FIXME: Maybe create an enum for these values.
 DISTROS = ["debian", "fedora", "ubuntu"]
 CONTAINER_RUNTIMES = ["podman", "docker"]
-IMAGE_NAME_BUILD_DEV_FMT = "dangerzone.rocks/build/{distro}:{version}"
-IMAGE_NAME_BUILD_FMT = "dangerzone.rocks/{distro}:{version}"
+IMAGES_REGISTRY = "ghcr.io/freedomofpress/"
+IMAGE_NAME_BUILD_DEV_FMT = "v2/dangerzone/build-dev/{distro}-{version}:{hash}"
+IMAGE_NAME_BUILD_ENDUSER_FMT = "v2/dangerzone/end-user/{distro}-{version}:{hash}"
```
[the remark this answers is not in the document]
almet: It is actually added in the `pull_image_registry` and `push_image_registry` methods (which are doing the podman/docker pull/push), so I believe we should be covered here.

dev_scripts/env.py
almet: Thanks, fixed in 0ae59f3b6d63b55cd8fc107188ada6df5f250edb

.github/workflows/ci.yml
almet: Great! This is fixed in 9077dacdadc8f5614a8accfe74c193b8ec7be5e3

dev_scripts/env.py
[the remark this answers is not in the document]
apyrgio: Right, I'm talking about our local runs of `env.py build | build-dev` though. It's weird not having a hostname, that's why I'm saying this.

dev_scripts/env.py
```diff
@@ -322,6 +340,34 @@ def dz_version():
...
+    """
+    hash_obj = hashlib.new("sha256")
+    for path in file_paths:
+        with open(path, "rb") as file:
+            file_data = file.read()
+            hash_obj.update(file_data)
+
+    if include_current_date:
+        current_date = date.today().strftime("%Y-%m-%d")
+        hash_obj.update(current_date.encode())
```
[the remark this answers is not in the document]
apyrgio: Hm, any particular reason you added the date in the name, and not the tag of the image?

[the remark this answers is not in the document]
almet: That's actually an error. Fixed in a4e14e495f8bccc1eb00b7a7aa9215cc1bfb1107.

dev_scripts/env.py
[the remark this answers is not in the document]
almet: Thanks for clarifying. It should always be the full name, you're right! 

Updated in 3d4d89b6978af2418b1d4ea4b5cb2c690b0ffcaa

.github/workflows/ci.yml
apyrgio: Eeek, you're right!

dev_scripts/env.py
```diff
@@ -306,14 +309,29 @@ def distro_build(distro, version):
...
     """Get the container image for the dev variant of a Dangerzone environment."""
-    return IMAGE_NAME_BUILD_DEV_FMT.format(distro=distro, version=version)
+    hash = hash_files(
+        get_files_in("dev_scripts")
+        + [
+            git_root() / "pyproject.toml",
+            git_root() / "poetry.lock",
+        ],
+        include_current_date=True,
+    )
```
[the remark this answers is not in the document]
almet: Done in [`09aaee2` (#907)](https://github.com/freedomofpress/dangerzone/pull/907/commits/09aaee2dbc980749cb53f5f1fd74e456fd0bfad0)

apyrgio (approved):

PR #914, opened 2024-09-09
ci: Update our CircleCI machines to Ubuntu 22.04

apyrgio:
Update our CircleCI machines for specific tests (Debian Bookworm and Fedora 40). It seems that the newest Podman version (5.2.1+), when creating a container using the `--userns nomap` triggers a permission denied error in older kernels. E.g.:

    Error: crun: cannot stat `/tmp/storage-run-1000/containers/overlay-containers/d00932f2600df7b0d8f4cc78e2346487ec92bfd17307127f3ae8d4e5bbc7887b/userdata/hosts`: Permission denied: OCI permission denied

The solution that works for us is to update the machine image from Ubuntu 20.04 to Ubuntu 22.04.

PR #916, opened 2024-09-10
added try excepts for unhandled exceptions

### LINKED ISSUE #776: Unhandled exceptions occurring inside function passed to `executor.map` are swallowed
Currently, if an unhandled exception occurs inside [`convert_doc`](https://github.com/freedomofpress/dangerzone/blob/7c4e62954f4ee71726de0a23655c8f791f05f290/dangerzone/logic.py#L75) when running `dangerzone-cli`, the exception is swallowed (traceback is not logged, the exit code is not changed).

This manifests in tests with dummy conversion - e.g. if [`shutil.copy`](https://github.com/freedomofpress/dangerzone/blob/7c4e62954f4ee71726de0a23655c8f791f05f290/dangerzone/isolation_provider/dummy.py#L65) raises an exception (e.g. `src` and `dst` being the same path), the `CLIResult` is still reported as a success due to the swallowed exception. [Here](https://github.com/naglis/dangerzone/commit/0d214c14d3c9045b9eeb7764fdedb55df4d9a5a3#diff-4e8715c7a425ee52e74b7df4d34efd32e8c92f3e60bd51bc2e1ad5943b82032eR222) is a demo test case.

IIUC during actual (not dummy) execution it would hide potential exceptions occurring [here](https://github.com/freedomofpress/dangerzone/blob/7c4e62954f4ee71726de0a23655c8f791f05f290/dangerzone/isolation_provider/base.py#L69-L70) or [here](https://github.com/freedomofpress/dangerzone/blob/7c4e62954f4ee71726de0a23655c8f791f05f290/dangerzone/isolation_provider/base.py#L90-L94).

I believe this is because the generator returned by `executor.map` is not consumed -  from [`ThreadPoolExecutor` docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map):

> If a _fn_ call raises an exception, then that exception will be raised when its value is retrieved from the iterator.

Here is a simplified form showing the issue:

```python
>>> from concurrent.futures import ThreadPoolExecutor
>>>
>>> def do_work(i):
...     print(f"do_work({i})")
...     1 / 0
...
>>> with ThreadPoolExecutor(max_workers=4) as executor:
...     _ = executor.map(do_work, range(4))
...
do_work(0)
do_work(1)
do_work(2)
do_work(3)
>>>
```

amnak613:
### Description
-what: added a try catch to the areas in the code mentioned in the pr
-why: will help any unhandled errors that are being swallowed within the code

### Test Case

- [x] Ensure all tests pass

Closes #776

### DIFF
--- a/dangerzone/logic.py
+++ b/dangerzone/logic.py
@@ -65,11 +65,17 @@ class DangerzoneCore(object):
         self, ocr_lang: Optional[str], stdout_callback: Optional[Callable] = None
     ) -> None:
         def convert_doc(document: Document) -> None:
-            self.isolation_provider.convert(
-                document,
-                ocr_lang,
-                stdout_callback,
-            )
+            try:
+                self.isolation_provider.convert(
+                    document,
+                    ocr_lang,
+                    stdout_callback,
+                )
+            except Exception as e:
+                log.exception(
+                    f"Unexpected error occurred while converting '{document}'"
+                )
+                document.mark_as_failed()
 
         max_jobs = self.isolation_provider.get_max_parallel_conversions()
         with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -19,6 +19,7 @@ from strip_ansi import strip_ansi
 from dangerzone.cli import cli_main, display_banner
 from dangerzone.document import ARCHIVE_SUBDIR, SAFE_EXTENSION
 from dangerzone.isolation_provider.qubes import is_qubes_native_conversion
+from dangerzone.util import get_resource_path
 
 from .conftest import for_each_doc, for_each_external_doc
 
@@ -218,6 +219,20 @@ class TestCliConversion(TestCliBasic):
         output_filename = str(Path(temp_dir) / uncommon_filename)
         result = self.run_cli([sample_pdf, "--output-filename", output_filename])
         result.assert_success()
+
+    ### Test method for swallowed exception
+    def test_output_filename_same_file_dummy_fails(self) -> None:
+        resource_path = get_resource_path("dummy_document.pdf")
+        # Using the same filename for both input and output should fail.
+        result = self.run_cli(
+            [
+                resource_path,
+                "--output-filename",
+                resource_path,
+                "--unsafe-dummy-conversion",
+            ]
+        )
+        result.assert_failure()
 
     def test_output_filename_new_dir(self, sample_pdf: str) -> None:
         output_filename = str(Path("fake-directory") / "my-output.pdf")


dangerzone/isolation_provider/base.py
```diff
@@ -88,8 +88,11 @@ def convert(
             log.exception(
                 f"An exception occurred while converting document '{document.id}'"
             )
-            self.print_progress(document, True, str(e), 0)
-            document.mark_as_failed()
+            try:
+                self.print_progress(document, True, str(e), 0)
+                document.mark_as_failed()
+            except Exception as e:
+                log.exception(f"Error occured while marking document as failed '{document.id}'")
```
apyrgio, line 95: Adding a try...except block here is not necessary, since the plan is to catch exceptions even higher up. I would ditch this change.

dangerzone/logic.py
```diff
@@ -64,16 +64,20 @@ def clear_documents(self) -> None:
...
+                    ocr_lang,
+                    stdout_callback,
+                )
+                        
+            max_jobs = self.isolation_provider.get_max_parallel_conversions()
+            with concurrent.futures.ThreadPoolExecutor(max_workers=max_jobs) as executor:
+                executor.map(convert_doc, self.documents)
+        except Exception as e:
+            print(e)
+            log.exception(f"Error occured while converting doc")
```
apyrgio, line 80: ... but not that high up. The original issue is that `executor.map()` swallows exceptions by design. This means that we don't gain something by wrapping `executor.map()` in a try...except block. Instead, we need to catch exceptions within the `convert_doc()` function, which is the highest an exception from the conversion may reach.

Something like :

```suggestion
      def convert_doc(document: Document) -> None:
          try
                self.isolation_provider.convert(
                    document,
                    ocr_lang,
                    stdout_callback,
                )
            except Exception:
                log.exception(f"Unexpected error occurred while converting '{document}'")
                document.mark_as_failed()
```

apyrgio (changes requested):
Hey @amnak613! Thanks a lot for your first contribution in the project. It seems that the CI tests are failing, so I've suggested a way to improve your solution and make them work. I think once you fix it, we'll be good to merge.

dangerzone/logic.py
amnak613: Thanks I updated the code so it is contained within the convert_doc function and the unit tests are passing

dangerzone/isolation_provider/base.py
amnak613: Removed this change

amnak613:
@apyrgio All tests are passing and was able to resolve some of the linter errors. I'm having some issues trying to figure out how to get rid of the last lint error on test_cli.py

tests/test_cli.py
```diff
@@ -219,6 +218,22 @@ def test_output_filename_uncommon(
         result = self.run_cli([sample_pdf, "--output-filename", output_filename])
         result.assert_success()
 
+    ###
```
almet, line 221: nitpick: you can remove this comment

almet (commented):
Thanks for this contribution, this is appreciated!

About the linting error you see, you should be able to fix it by running `make format` (which will apply `isort`, which reorders the imports alphabetically).

We changed the name of this utility recently, so be sure to get the latest changes from the `main` branch by running `git rebase main`.

This should also make the CI happier about Fedora 40 and Debian Trixie, [which we fixed last week](#914)

apyrgio (approved):

apyrgio:
Thanks a lot for the changes @amnak613. I agree with @almet, with some minimal changes the PR will be good to merge. If you want to, we can make those ourselves and just merge the PR.

tests/test_cli.py
amnak613: I'll fix this up, I can go ahead and add the changes and then merge the pr

amnak613:
@apyrgio @almet Updated and added changes once the pr has all the approvals will merge

apyrgio (approved):

apyrgio:
Squashed the commits and verified that the CI passed when the branch was last updated (now the CI fails, but that's due to a CircleCI brown-out). Thanks for the changes @amnak613.

PR #921, opened 2024-09-17
Update README.md to use / instead of /about.html

DustinAlandzes:
Hey!

I noticed the link to the official site is broken:
image

so I changed it so that it goes to / instead of about.html:
image


almet:
Thanks for taking the time to propose these changes Dustin! The actual link is https://dangerzone.rocks/about/, so I updated your PR to reflect this and pushed it in https://github.com/freedomofpress/dangerzone/pull/923 as I don't have push access to your fork.

Closing this one now in favor of the new one.

PR #924, opened 2024-09-18
docs: Update gVisor design doc

apyrgio:
Update the gVisor design doc, to better reflect the current state of the gVisor integration. More specifically, the following have changed since this design doc was merged:

* We have dropped the need for the `SETFCAP` capability.
* We have added the SELinux label `container_engine_t` to the outer container.


almet (approved):
Looks good, thanks!

PR #926, opened 2024-09-19
Always use our own seccomp policy as a default.

### LINKED ISSUE #908: All conversions fail with "Unspecified error", using Orbstack
Trying to use Dangerzone 0.7.0 on some files recently, everything fails with "Unspecified error".

I'm using Dangerzone with Orbstack (https://orbstack.dev/) rather than Docker Desktop, so perhaps there might be some configuration issue or something? Not sure how to generate logs or anything but let me know if there's anything I can provide. I used to use Dangerzone & Orbstack without any issues but perhaps there's been a breaking change somewhere recently on either end.

almet:
As per Etienne Perot's comment on #908:

> Then it seems to me like it would be easy to simply apply this seccomp profile under all container runtimes (since there's no reason why the same image and the same command-line would call different syscalls under different container runtimes).

As mentioned in the comments there, we might want to tighten the default seccomp policy, to have more control on what we accept or not.

Fixes #908

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -109,38 +109,30 @@ class Container(IsolationProvider):
         * Set the `container_engine_t` SELinux label, which allows gVisor to work on
           SELinux-enforcing systems
           (see https://github.com/freedomofpress/dangerzone/issues/880).
+        * Set a custom seccomp policy for every container engine, since the `ptrace(2)`
+          system call is forbidden by some.
 
         For Podman specifically, where applicable, we also add the following:
         * Do not log the container's output.
-        * Use a newer seccomp policy (for Podman 3.x versions only).
         * Do not map the host user to the container, with `--userns nomap` (available
           from Podman 4.1 onwards)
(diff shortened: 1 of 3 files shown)


dangerzone/isolation_provider/container.py
```diff
@@ -126,21 +126,13 @@ def get_runtime_security_args() -> List[str]:
...
         else:
             security_args = ["--security-opt=no-new-privileges:true"]
-            # Older Docker Desktop versions may have a seccomp policy that does not
-            # allow `ptrace(2)`. In these cases, we specify our own. See:
-            # https://github.com/freedomofpress/dangerzone/issues/846
-            if Container.get_runtime_version() < (25, 0):
-                security_args += custom_seccomp_policy_arg
+
+        # We specify a custom seccomp policy, because on some container engines, the
+        # default policy might now allow `ptrace(2)`.
```
EtiennePerot, line 133: typo: `now` -> `not`

almet: Thanks, fixed.

PR #931, opened 2024-09-26
Sign Dangerzone executables and installer with SHA256

jkarasti:
Something I noticed while working on the windows installer is that it and executables shipped with dangerzone are still using SHA1 as the signature algorithm so let's fix that.

The first commit switches to SHA256 as recommended by the signtool.exe [docs](https://learn.microsoft.com/en-us/windows/win32/seccrypto/signtool).

A further improvement here could be changing `/t http://time.certum.pl/` to `/tr http://time.certum.pl/` and adding `/td sha256` but I have no idea if the Certum time server supports RFC 3161 time stamp protocol, it's not [documented](https://support.certum.eu/en/signing-the-code-using-tools-like-signtool-and-jarsigner-instruction/) at least. Might be something to take into consideration at some point as not using `/td` is going to cause an error in a future version of signtool, though there's no timeline provided for the change.

The second commit adds an additional step that runs `signtool verify` to verify the signature of the files after signing. Something I picked up from the docs for code signing provided by Certum. Not really a mandatory step, but better safe than sorry, especially with something like code signing.

I obviously don't have access to the Dangerzone code signing certificate, so this is very much a shot in the dark as to whether it works as intended or not.

### DIFF
--- a/install/windows/build-app.bat
+++ b/install/windows/build-app.bat
@@ -2,12 +2,14 @@ REM delete old dist and build files
 rmdir /s /q dist
 rmdir /s /q build
 
-REM build the exe
+REM build the gui and cli exe
 python .\setup-windows.py build
 
 REM code sign dangerzone.exe
-signtool.exe sign /v /d "Dangerzone" /a /n "Freedom of the Press Foundation" /fd sha1 /t http://time.certum.pl/ build\exe.win-amd64-3.12\dangerzone.exe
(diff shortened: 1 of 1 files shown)


apyrgio:
Heads up, I just managed to actually test the changes. Better late than never I guess.

The changes look great, and I can verify that `signtool.exe verify` instructions return the appropriate message ("Successfully verified: ..."). Merging this PR now.

apyrgio (approved):

Issue #928, opened 2024-09-23
Latest gVisor release breaks Dangerzone


apyrgio:
It seems that the latest gVisor release ([`release-20240916.0`](https://github.com/google/gvisor/releases/tag/release-20240916.0)) is breaking the development branch of Dangezone. I believe this is the case due to the following:

1. Our tests have been [failing](https://github.com/freedomofpress/dangerzone/actions/runs/10932011706/job/30348083281) since 5 days ago, when the newest binaries for gVisor were released.
2. I have rebuilt the v0.7.0 image for Dangerzone, which should run properly on all of our supported platforms, and I get an error as well.

The gVisor error we're seeing when running the container by hand is this:

```
I0923 13:05:11.356492       1 main.go:201] **************** gVisor ****************
W0923 13:05:11.358402       1 boot.go:266] Not setting product_name: open /sys/devices/virtual/dmi/id/product_name: no such file or directory
I0923 13:05:11.358522       1 boot.go:279] Setting host-shmem-huge: "never"
W0923 13:05:11.359241       1 specutils.go:129] noNewPrivileges ignored. PR_SET_NO_NEW_PRIVS is assumed to always be set.
I0923 13:05:11.359297       1 chroot.go:92] Setting up sandbox chroot in "/tmp"
W0923 13:05:11.359386       1 chroot.go:109] Failed to copy /etc/localtime: open /etc/localtime: no such file or directory. UTC timezone will be used.
I0923 13:05:11.359425       1 chroot.go:37] Mounting "proc" at "/tmp/proc"
W0923 13:05:11.359477       1 util.go:64] FATAL ERROR: error setting up chroot: error mounting proc in chroot: error mounting "proc" at "/tmp/proc": mount("proc", "/tmp/proc", 15) failed: operation not permitted
error setting up chroot: error mounting proc in chroot: error mounting "proc" at "/tmp/proc": mount("proc", "/tmp/proc", 15) failed: operation not permitted
D0923 13:05:11.360661      12 sandbox.go:1330] Destroying sandbox "dangerzone"
```

You can find the full error report here: [gvisor.log](https://github.com/user-attachments/files/17098118/gvisor.log). I'll give it a check with the previous gVisor release as well.

apyrgio:
Yeap, building Dangerzone with the 2024-08-26 gVisor release works. The change I made was:

```diff
diff --git a/Dockerfile b/Dockerfile
index 9988660..872e634 100644
--- a/Dockerfile
[...]

PR #932, opened 2024-09-26
Preparations for the on-host conversion PR

apyrgio:
Make some preparations for the on-host conversion PR:
* Add a temporary fix for our gVisor breakage (see #928)
* Improve our `.gitattributes` file, so that Python files on Windows retain their LF line endings.
   - This broke caching in our CI for the tessdata bundle.
* Make dummy tests faster (general quality of life improvement)
* Add an Intel macOS runner (for completeness)
* Major improvements in termination logic (figured them out while working on the on-host conversion PR)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,15 @@ jobs:
         run: poetry run .\install\windows\build-app.bat
 
   macOS:
-    runs-on: macos-latest
+    name: "macOS (${{ matrix.arch }})"
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+          - runner: macos-latest  # CPU type: Apple Silicon (M1)
+            arch: arch64
+          - runner: macos-13  # CPU type: Intel x86_64
+            arch: x86_64
     env:
       DUMMY_CONVERSION: 1
     steps:
--- a/dangerzone/isolation_provider/base.py
+++ b/dangerzone/isolation_provider/base.py
@@ -1,5 +1,8 @@
 import contextlib
 import logging
+import os
+import platform
+import signal
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -30,7 +30,15 @@ jobs:
...
-    runs-on: macos-latest
+    name: "macOS (${{ matrix.arch }})"
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+          - runner: macos-latest  # CPU type: Apple Silicon (M1)
+            arch: arch64
+          - runner: macos-13  # CPU type: Intel x86_64
+            arch: x86_64
```
almet, line 41: :100:

dangerzone/isolation_provider/qubes.py
```diff
@@ -67,6 +67,7 @@ def start_doc_to_pixels_proc(self, document: Document) -> subprocess.Popen:
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             stderr=stderr,
+            start_new_session=True,
```
almet, line 70: I believe we should add the commit message here as a comment for clarity.

```suggestion
            # Start the conversion process in a new session, so that we can later on
            # kill the process group, without killing the controlling script.
            start_new_session=True,
```

tests/isolation_provider/test_dummy.py
```diff
@@ -25,19 +25,20 @@ def start_doc_to_pixels_proc(self, document: Document) -> subprocess.Popen:
...
         )
 
-    def terminate_doc_to_pixels_proc(
-        self, document: Document, p: subprocess.Popen
-    ) -> None:
-        p.terminate()
-
 
 @pytest.fixture
 def provider_wait() -> DummyWait:
```
almet, line 33: Reading "provider_wait" here didn't made it clear to me on first sight. Maybe we could name this `blocked_provider` :thinking: ?

almet (approved):
Wooo! Lot of great things here. Thanks for removing these sleeps and unblocking the situation regarding the kills with timeouts!

Added a few comments. Other that that, feel free to merge!

dangerzone/isolation_provider/base.py
```diff
@@ -209,6 +212,35 @@ def terminate_doc_to_pixels_proc(
         """Terminate gracefully the process started for the doc-to-pixels phase."""
         pass
 
+    def _signal_process_group(self, p: subprocess.Popen, signo: int) -> None:
```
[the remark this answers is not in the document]
apyrgio: Yeah, we do have a `util.py`, and we could move it there, but there are a few assumptions that are tied with the isolation provider:

1. The fact that we don't raise an error if the process no longer exists. In a neutral setting, we would have to inform the caller about it, but it doesn't make sense for the doc-to-pixels process.
2. The fact that we don't kill the process group on Windows, and we fallback to killing the process. That's a choice that makes sense for the doc-to-pixels process, but I'm not sure we can generalize it :-/

dangerzone/isolation_provider/qubes.py
apyrgio: Makes sense, did so in dfc3dfd.

tests/isolation_provider/test_dummy.py
apyrgio: Good point. Actually, we will fully remove this in a PR I'll send soon, so let's keep it as is for now, to avoid merge conflicts.

dangerzone/isolation_provider/base.py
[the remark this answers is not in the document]
almet: Thanks for the explanations.

If you don't think it's worth generalizing, I suggest removing it from the class and having it as a first level function of the python module rather than being a method.

If we were to generalize it, naming would be important, and parameters to these methods might also be useful (like `fallback_to_process` and `raise_on_error`, for instance) so that it's clear what it's done.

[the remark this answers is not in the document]
apyrgio: Ok, then let's go with turning the methods into module-level functions: 0d55aea63552c4bfdd079348730b473374bd559f

almet:
Everything seems alright on this PR. Feel free to merge after a rebase + squash of the commits :-)

PR #940, opened 2024-10-07
Vendor PyMuPDF

apyrgio:
This PR allows our build scripts to:
* Install PyMuPDF in a special location (`dangerzone/vendor`) so that it can be used by Dangerzone in a subsequent PR (see https://github.com/freedomofpress/dangerzone/pull/748)
* Override the .deb building phase and vendor PyMuPDF, just for Debian.

Refs #625


install/linux/vendor-pymupdf.py
```diff
@@ -0,0 +1,43 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--dest",
+        default=DZ_VENDOR_DIR,
+        help="The destination directory for the vendored packages (default: ./dangerzone/vendor)",
+    )
+    args = parser.parse_args()
+
+    print(">>> Getting PyMuPDF deps as requirements.txt", file=sys.stderr)
```
almet, line 20: Why using `stderr` here? 🤔

install/linux/vendor-pymupdf.py
```diff
@@ -0,0 +1,43 @@
...
+    cmd = [
+        "python3",
+        "-m",
+        "pip",
+        "install",
+        "--no-cache-dir",
+        "--no-compile",
+        "-t",
+        args.dest,
+        "-r",
```
almet, line 34: It would be easier to read if using the long-form arguments, like `--requirement` and `--target`

install/linux/vendor-pymupdf.py
```diff
@@ -0,0 +1,43 @@
...
+        default=DZ_VENDOR_DIR,
+        help="The destination directory for the vendored packages (default: ./dangerzone/vendor)",
+    )
+    args = parser.parse_args()
+
+    print(">>> Getting PyMuPDF deps as requirements.txt", file=sys.stderr)
+    cmd = ["poetry", "export", "--only", "container"]
+    container_requirements_txt = subprocess.check_output(cmd)
+
+    print(f">>> Vendoring PyMuPDF under '{args.dest}'", file=sys.stderr)
```
almet, line 24: I've just read https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program which explains why we shouldn't use pip internal library, maybe it's worth dropping a line about this here?

install/linux/vendor-pymupdf.py
apyrgio: Well, I prefer to print log message in `stderr`, that's why I mention it explicitly. But I can replace this with logigng statements (which also log in `stderr`), if it feels ugly.

dangerzone/__init__.py
```diff
@@ -1,6 +1,13 @@
 import os
 import sys
 
+try:
+    from . import vendor  # type: ignore [attr-defined]
+
+    sys.path.insert(0, vendor.__path__[0])
```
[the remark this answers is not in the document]
apyrgio: Nice catch. Actually, I can't use `logging` there, since it's not initialized yet. Still, I can use `print(..., file=sys.stderr)`. Check it out in 3892dec4.

debian/rules
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/make -f
 export PYBUILD_NAME=dangerzone
 export DEB_BUILD_OPTIONS=nocheck
+#export PYBUILD_INSTALL_ARGS=--install-lib=/usr/lib/python3/dist-packages
+#export PYTHONDONTWRITEBYTECODE=1
+#export DH_VERBOSE=1
```
apyrgio: Yikes, I just realized these options are disabled. I think they should be enabled actually, but I'll make a cross-check with #748, once it works again. In the meantime, I've enabled them in 3892dec4.

install/linux/vendor-pymupdf.py
apyrgio: Sure thing, added in 3892dec4.

install/linux/vendor-pymupdf.py
apyrgio: I agree, added in 3892dec4.

install/linux/vendor-pymupdf.py
```diff
@@ -0,0 +1,46 @@
...
+import argparse
+import subprocess
+import sys
+from pathlib import Path
+
+DZ_VENDOR_DIR = Path("./dangerzone/vendor")
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
almet, line 12: I've seen in other parts of the code that we're using click, and I'm wondering why it's not the case here? 🤔

install/linux/vendor-pymupdf.py
almet: I think I would prefer it if we were to use the logger functions of python instead. By default, the [`LastResort` handler](https://docs.python.org/3/library/logging.html#logging.lastResort) outputs to `stderr`, but it's also possible to configure it to behave differently.

install/linux/vendor-pymupdf.py
apyrgio: Where I can, I try not to introduce external libraries for our dev scripts, so that we don't need enter a Poetry environment just to run them.

dangerzone/__init__.py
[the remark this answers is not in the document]
apyrgio: Also, check out 68c516f, which logs to `stderr`.

install/linux/vendor-pymupdf.py
almet: Thanks for clarifying this. I would like to challenge this, if that's open for discussion?

I believe using the same tooling across the project would be beneficial in terms of general maintenance, as it would require from the dev team to only know one tool, rather than two, dig deeper in their docs etc.

That being said, that's the current status of the code, and I don't think it's worth blocking the work here, so let's discuss that the next time we meet!

apyrgio: Sure, let's discuss when we meet. I get what you're saying, but I always kinda believed that `argparse`, being in stdlib, is ubiquitous. So, it feels to me more like the default option, instead of "a different option". Also, given that this is a dev script, I figured it's best not to complicate things more. We now have lots of dev scripts though, so let's see what makes sense in the long run.

almet: I agree about `argparse` being seen as the default option. It's part of the batteries included with python. I've been using it extensively in the past, and I mostly like it. I also can see why and how `click` enhances this situation (and I've also used it on different projects).

To me the question seems to be mainly about which one we decide on using, and sticking with one rather than having two options depending on the code base. My preference is to use tools like `click` and `requests` because they solve real problems of their stdlib counterparts.

I would like to know more about why you prefer not relying on dependencies for the dev scripts. (I might be missing an obvious reason why)

almet: Closing this to not block the merge. Discussion is happening here: https://github.com/freedomofpress/dangerzone/issues/946

dangerzone/__init__.py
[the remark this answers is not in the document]
apyrgio: Well, we _can_ initialize the `logging` module ourselves, but it's a separation of concerns thing. The `dangerzone/__init__.py` module is part of the Dangerzone library. Libraries tend to log, but not setup the logging infrastructure. The responsibility for this falls to the frontends (e.g., `dangerzone-cli`).

If a library decides to setup logging at import time, then any consumer of this library that wants to setup logging as well will need to build on top of what this library did. This is confusing, so libraries tend not to do that. Does this make more sense?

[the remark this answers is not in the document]
almet: I'm not a fan of logging to `stderr` without going trough a logger at all, because it can pollute the space there.

I'm starting to think we could actually do this `sys.path` changes inside a `vendorized/pymupdf.py` folder living in our code tree, in which case that would be the one we import. This has the following advantages:

1. It is more explicit. We look in the vendorized path, we can use loggers there. 
2. We do not add the `vendor` path for all the libraries that we need, but one by one, potentially configuring them the way we want in their specialized modules.
3. It solves the issue about logging, because by the time this would be imported, it will be configured.

What do you think?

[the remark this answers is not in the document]
apyrgio: We discussed this out-of-band, and landed on adding a `logger.debug()` statement there, without configuring logging ourselves. If the caller has done so, they will see a debug message about vendored packages. If not, they will see nothing. This change can be found in f4073ff013513c2e8b87f57cbba854771819f1b5.

apyrgio:
I have rebased on top of the latest `main` branch and squashed all reviewed fixup commits. I've added one more fixup commit where we switch from a print to log statement, which is awaiting for your review Alexis. Other than that, it's good to go.

almet (approved):
Thanks, let's merge!

PR #939, opened 2024-10-03
Use github issue templates

almet:
This adds two issue templates, one as a form and the other as some text already filled in.

Here is what it looks like:

image

---

Bug report:

image

---

Feature request:

image

---

Is there any other input that could be valuable here?

Note that we could also have a specific template for "Conversion failed" issues, where we could add some more specific information about the docker / podman setup, and ask the user to run specific tasks (like add SC_DEBUG to the docker run command line to get more logs back).

I am curious what you think on this, and if you think it could be valuable? 🤔 

Edit: I realize that these screencaps are not up to date. If you want to have a look at what it looks like now for the form, look at [this link instead](https://github.com/freedomofpress/dangerzone/blob/920-gh-issue-templates/.github/ISSUE_TEMPLATE/bug_report.yml)

Issue #442, opened 2023-06-13
Have a better approach to detecting 


deeplow:
The current way that the code detects if we're running in debug mode is not ideal:
```python
getattr(sys, "dangerzone_dev", False):
```
It hijacks the `sys` module to add something unrelated. To cite @apyrgio:

> This has started to become so common, that we probably need to move it under a common module.

PR #941, opened 2024-10-07
Add a `--debug` flag to the CLI to help retrieve more logs.

almet:
While adding Github Issue Templates on this repository (#939) we found that the commands we require the user to enter in their terminals can be quite complex. Some of them might require some bash-fu, for instance to define the location of the custom seccomp profile we are using, as it differs depending the OS.

This is a proposal to add a `--debug` flag to the `dangerzone-cli` to simplify the process of generating proper logs.

When the flag is set:

- The `RUNSC_DEBUG=1` environment variable is added to the outer container ;
- the stderr from the outer container is attached to the exception, and displayed to the user on failures.

Info: tests are currently failing and I didn't put too much effort in making them pass, as this is here mainly to see if it could make sense to add this in the first place.

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -42,6 +42,12 @@ def print_header(s: str) -> None:
     type=click.UNPROCESSED,
     callback=args.validate_input_filenames,
 )
+@click.option(
+    "--debug",
+    "debug",
+    flag_value=True,
+    help="Run Dangerzone in debug mode, to get logs from gVisor.",
+)
 @click.version_option(version=get_version(), message="%(version)s")
 @errors.handle_document_errors
 def cli_main(
@@ -50,6 +56,7 @@ def cli_main(
     filenames: List[str],
     archive: bool,
     dummy_conversion: bool,
+    debug: bool,
 ) -> None:
     setup_logging()
 
@@ -58,7 +65,7 @@ def cli_main(
     elif is_qubes_native_conversion():
         dangerzone = DangerzoneCore(Qubes())
     else:
-        dangerzone = DangerzoneCore(Container())
+        dangerzone = DangerzoneCore(Container(debug=debug))
 
     display_banner()
     if len(filenames) == 1 and output_filename:
--- a/dangerzone/isolation_provider/base.py
+++ b/dangerzone/isolation_provider/base.py
@@ -5,7 +5,9 @@ import platform
 import signal
 import subprocess
 import sys
+import threading
 from abc import ABC, abstractmethod
+from io import BytesIO
 from typing import IO, Callable, Iterator, Optional
 
 import fitz
@@ -18,7 +20,6 @@ from ..util import get_tessdata_dir, replace_control_chars
 
 log = logging.getLogger(__name__)
 
-MAX_CONVERSION_LOG_CHARS = 150 * 50  # up to ~150 lines of 50 characters
 DOC_TO_PIXELS_LOG_START = "----- DOC TO PIXELS LOG START -----"
 DOC_TO_PIXELS_LOG_END = "----- DOC TO PIXELS LOG END -----"
 
@@ -75,9 +76,9 @@ def read_int(f: IO[bytes]) -> int:
     return int.from_bytes(untrusted_int, "big", signed=False)
 
 
-def read_debug_text(f: IO[bytes], size: int) -> str:
-    """Read arbitrarily long text (for debug purposes), and sanitize it."""
-    untrusted_text = f.read(size).decode("ascii", errors="replace")
+def sanitize_debug_text(text: bytes) -> str:
+    """Read all the buffer and return a sanitized version"""
+    untrusted_text = text.decode("ascii", errors="replace")
     return replace_control_chars(untrusted_text, keep_newlines=True)
 
 
@@ -86,11 +87,15 @@ class IsolationProvider(ABC):
     Abstracts an isolation provider
     """
 
-    def __init__(self) -> None:
-        if getattr(sys, "dangerzone_dev", False) is True:
+    def __init__(self, debug: bool = False) -> None:
+        self.debug = debug
+        if self.should_capture_stderr():
             self.proc_stderr = subprocess.PIPE
         else:
             self.proc_stderr = subprocess.DEVNULL
+
+    def should_capture_stderr(self) -> bool:
+        return self.debug or getattr(sys, "dangerzone_dev", False)
 
     @abstractmethod
     def install(self) -> bool:
@@ -327,7 +332,11 @@ class IsolationProvider(ABC):
         timeout_force: int = TIMEOUT_FORCE,
     ) -> Iterator[subprocess.Popen]:
         """Start a conversion process, pass it to the caller, and then clean it up."""
+        # Store the proc stderr in memory
+        stderr = BytesIO()
         p = self.start_doc_to_pixels_proc(document)
+        stderr_thread = self.start_stderr_thread(p, stderr)
+
         if platform.system() != "Windows":
             assert os.getpgid(p.pid) != os.getpgid(
                 os.getpid()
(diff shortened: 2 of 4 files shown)


apyrgio (commented):
That's a nice step towards getting debug logs from users without asking them to run weird Docker/Podman commands. I've commented on a few things that we can improve, but overall I like where this is going.

dangerzone/cli.py
```diff
@@ -42,6 +42,7 @@ def print_header(s: str) -> None:
     type=click.UNPROCESSED,
     callback=args.validate_input_filenames,
 )
+@click.option("--debug", "debug", flag_value=True)
```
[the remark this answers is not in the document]
almet: 💯 See this in 55b170b69680ab5e19a9b4694f7717cd56abd80c

dangerzone/cli.py
```diff
@@ -58,7 +60,7 @@ def cli_main(
     elif is_qubes_native_conversion():
         dangerzone = DangerzoneCore(Qubes())
     else:
-        dangerzone = DangerzoneCore(Container())
+        dangerzone = DangerzoneCore(Container(debug=debug))
```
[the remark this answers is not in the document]
almet: Thanks, I somehow thought we could only rely on the `--debug` flag to kinda replace the `sys.dangerzone_dev` setting to know if we want to output debug logs.

I've updated the code to reflect the two scenarii then (in 12fa83043811304a5c9b6cb6c1d878be6e177778)

almet:
This seems correlated to #442.

dangerzone/conversion/errors.py
```diff
@@ -18,11 +18,20 @@ class ConversionException(Exception):
...
             self.error_message = error_message
+        self.logs = logs
         super().__init__(self.error_message)
 
+    def __str__(self):
+        msg = f"{self.error_message}"
+        if self.logs:
+            msg += f" {self.logs}"
+        return msg
+
```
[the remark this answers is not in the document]
apyrgio: I think that container errors are now reported on the `main` branch. Give it a look and let me know if now.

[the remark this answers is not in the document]
almet: This PR will need some revisit at some point, as the main point is to add `RUNSC_DEBUG=1` info to the output, making it possible to run by end users without having to do a lot of copy/paste.

When I tried to rebase this on top of the latest main, it seemed that the run was now blocked, and I didn't investigated further yet.

apyrgio:
We had a discussion with Alexis. He mentioned that adding the `--debug` flag makes the Dangezone CLI hang. The reason behind this bug is the fact that we read from `stderr` only after the conversion is done. However, with `RUNSC_DEBUG=1`, gVisor logs to `stderr` early, and thus the container hangs, because we don't read from the pipe.

almet:
I've rebased this branch on top of the latest `main` (actually replayed the changes manually) and it shows that it hangs.

almet:
This is currently blocking, because the docker process writes to `stderr` but the pipe is never called by the other side, resulting in a process hang (the buffer is full and so it stops there).

Our conversion process is currently running in a thread pool, 

After a discussion with @apyrgio this morning on the matter, there seem to be two different outcomes for this:

- One "involved" way where we change all of the caller code to be async;
- We start a thread reading stderr

I went with the second option for now, but the first one could also make sense.

dangerzone/isolation_provider/base.py
```diff
@@ -86,11 +88,20 @@ class IsolationProvider(ABC):
...
         else:
             self.proc_stderr = subprocess.DEVNULL
+        self.stderr_queue: Queue = Queue()
+
+    def should_capture_stderr(self) -> bool:
+        return self.debug or getattr(sys, "dangerzone_dev", False)
+
+    @staticmethod
+    def is_runtime_available() -> bool:
+        return True
```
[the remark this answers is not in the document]
almet: Yes it is. Removed.

dangerzone/isolation_provider/container.py
```diff
@@ -177,14 +181,22 @@ def exec_container(
...
             + command
         )
         args = [container_runtime] + args
-        return self.exec(args)
+        args_str = " ".join(shlex.quote(s) for s in args)
+        log.info("> " + args_str)
+
+        process = self.exec(args)
+        # Start stderr reader thread, attaching it to the process
+        self.start_stderr_thread(process)
```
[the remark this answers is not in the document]
almet: Good idea! Updated the code to reflect this.

dangerzone/isolation_provider/base.py
```diff
@@ -355,3 +366,23 @@ def doc_to_pixels_proc(
...
+
+    def _stream_stderr(self, stderr: IO[bytes], queue: Queue) -> None:
+        """Read stderr in a separate thread to avoid blocking"""
+        try:
+            for line in stderr:
+                queue.put(line)
+                if self.debug:
+                    log.debug(line.decode().strip())
+        except (ValueError, IOError) as e:
+            log.debug(f"Stderr stream closed: {e}")
```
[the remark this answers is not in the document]
almet: It is a simpler approach, thanks. I've updated the code to reflect this, let me know what you think!

dangerzone/isolation_provider/base.py
```diff
@@ -343,15 +353,30 @@ def doc_to_pixels_proc(
...
 
-            # Read the stderr of the process only if:
-            # * Dev mode is enabled.
-            # * The process has exited (else we risk hanging).
-            if getattr(sys, "dangerzone_dev", False) and p.poll() is not None:
-                assert p.stderr
-                debug_log = read_debug_text(p.stderr, MAX_CONVERSION_LOG_CHARS)
+            if self.should_capture_stderr():
+                self.stderr.seek(0)
+                debug_log = read_debug_text(self.stderr, MAX_CONVERSION_LOG_CHARS)
```
apyrgio, line 358: Hm, if the stderr thread is still writing to the `BytesIO` buffer (e.g., due to a race), then seeking to `0` risks overwriting the first bytes in the buffer. Also, we may get incomplete logs, without us realizing, which was not the case before (because we use `p.poll() is not None`.

I propose we make some minor adjustments to the above logic:
1. Check if the thread is alive after waiting for 1 second (`.join(timeout=1)` and `.is_alive()`)
2. Grab the `BytesIO` buffer with `BytesIO.getvalue()`, which doesn't require seeking.
3. Adjust `read_debug_text()` to sanitize the `bytes` buffer of (2), instead of reading the `BytesIO` object.
4. In case the thread is still alive, add an `(incomplete)` indication to the `----- DOC TO PIXELS LOG START -----` message.

How does that sound, do you have a better idea?

dangerzone/isolation_provider/container.py
```diff
@@ -154,7 +154,7 @@ def exec(
             args,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
-            stderr=self.proc_stderr,
+            stderr=subprocess.PIPE,
```
apyrgio, line 157: I think we need to reinstate the `self.proc_stderr` value, else we will always log to a pipe, even if we don't capture stderr.

dangerzone/logic.py
```diff
@@ -1,6 +1,7 @@
 import concurrent.futures
 import json
 import logging
+from io import StringIO
```
apyrgio, line 4: I think this can be removed.

dangerzone/isolation_provider/base.py
```diff
@@ -86,11 +88,17 @@ class IsolationProvider(ABC):
...
-    def __init__(self) -> None:
-        if getattr(sys, "dangerzone_dev", False) is True:
+    def __init__(self, debug: bool = False) -> None:
+        self.debug = debug
+        if self.should_capture_stderr():
             self.proc_stderr = subprocess.PIPE
         else:
             self.proc_stderr = subprocess.DEVNULL
+        # Store the proc stderr in memory
+        self.stderr = BytesIO()
```
apyrgio, line 98: Kind of a minor note for now, but will be important in the future: I'd suggest not storing objects in the `IsolationProvider` instance that are affected by the conversion. It's not important for now, but when we will have multiple conversions in flight, then it won't work.

The `BytesIO` object in particular can be moved to the `doc_to_pixels_proc()` context manager, and get passed to the stderr thread.

apyrgio (commented):
Thanks a lot for the changes. I think we're :pinching_hand: close to merging it. I've left a few comments, but overall I like the new approach.

Oh, let's also have an issue to ask people to use the `--debug` flag, once we have a release out.

dangerzone/isolation_provider/base.py
almet: Thanks for mentioning this. I even think that we could use composition rather than inheritance in the future, to make that impossible :-)

I did the changes you mentioned in 3b961e6b9009779cd56609459eda84c1bb56469f

dangerzone/isolation_provider/base.py
apyrgio: Nice, thanks a lot for the change. I have one important comment, and then some nits that you can ignore if you want to:

I think that `MAX_CONVERSION_LOG_CHARS` will actually truncate the debug output of gVisor. It amounts to 150 lines of 50 chars roughly, but I think gVisor's debug logs are more than that. Now that we have a non-blocking way to read the stderr, I think we can drop the `MAX_CONVERSION_LOG_CHARS` constant altogether, since it no longer serves the original purpose.

Nits:
* I'd maybe rename `read_debug_text()` to `sanitize_debug_text()`.
* It would be nice to know if the debug log is incomplete, without scrolling to the end.

apyrgio: > About MAX_CONVERSION_LOGS_CHARS, I felt exactly the same so thanks for the feedback! Dropped now and renamed the function in https://github.com/freedomofpress/dangerzone/commit/8532688bab6b292c4690a751de1547ffd9d7363c

Great, thanks!

> About the "incomplete" log, I feel that having in the end is enough as it's where we would expect to have some more info. I might be missing some practical information you have. Do you have anything specific in mind?

Nothing fancy, something like:

```python
DOC_TO_PIXELS_LOG_START = "----- DOC TO PIXELS LOG START -----"
DOC_TO_PIXELS_LOG_START_INCOMPLETE = "----- DOC TO PIXELS LOG START (incomplete) -----"

log_header = DOC_TO_PIXELS_LOG_START
if stderr_thread.is_alive():
    log_header = DOC_TO_PIXELS_LOG_START_INCOMPLETE

log.info(
    "Conversion output (doc to pixels)\n"
    f"{log_header}\n"
    f"{debug_log}"  # no need for an extra newline here
    f"{DOC_TO_PIXELS_LOG_END}"
)
```

But I'll leave it to you, really.

apyrgio (approved):

dangerzone/isolation_provider/base.py
almet: I actually went with removing the `DOC_TO_PIXELS_LOG_{START,STOP}` intermediate variables as they seemed superfluous, and added code with the same outcome as what you're proposing.

You can read that in 26c3d096710ba49424a8438c8c4f33c30dab261a

apyrgio: Heh, you're right, we can just remove them! Fix looks good, I think you can dump the dollar sign `$` though.

almet: Good catch. Oh my god JS, stay where you are :see_no_evil:

almet:
Okay, just rebased. Will merge as soon as the CI is green.

PR #942, opened 2024-10-07
Prevent user from using illegal characters in output filename

apyrgio:
Add some checks in the Dangerzone GUI and CLI that will prevent a user from mistakenly adding illegal characters in the output filename.

Original PR: https://github.com/freedomofpress/dangerzone/pull/834

### DIFF
--- a/dangerzone/document.py
+++ b/dangerzone/document.py
@@ -1,8 +1,10 @@
 import enum
 import logging
 import os
+import platform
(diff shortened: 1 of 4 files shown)


almet (approved):
Adding my approval here, just so you can merge, beause otherwise github will not allow you to review (what it thinks is) your own PR

Issue #944, opened 2024-10-09
Replace Github CI `set-output` command by environment variables


almet:
`set-output` command is actually deprecated since 2022.

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

We should use something like this instead:

```yaml
- name: Set output
  run: echo "{name}={value}" >> $GITHUB_OUTPUT
```

PR #945, opened 2024-10-09
CI: Replace `set-output` by environment variables

almet:


apyrgio:
Oh right, I wanted to tackle this as well. Thanks for fixing it, feel free to merge.

_(maybe also add a "Fixes #944" in the commit message)_

apyrgio (approved):

PR #952, opened 2024-10-14
Catch installation errors and display them.

### LINKED ISSUE #193: Handle container install failure

almet:
This catches the errors that can happen during installation, and during some other phases, and display them to the end user in the interface.

Fixes #193

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import platform
-import subprocess
 import tempfile
 import typing
 from multiprocessing.pool import ThreadPool
@@ -20,15 +19,19 @@ else:
         from PySide6.QtWidgets import QTextEdit
     except ImportError:
         from PySide2 import QtCore, QtGui, QtSvg, QtWidgets
-        from PySide2.QtWidgets import QAction, QTextEdit
         from PySide2.QtCore import Qt
+        from PySide2.QtWidgets import QAction, QTextEdit
 
 from .. import errors
 from ..document import SAFE_EXTENSION, Document
-from ..isolation_provider.container import Container, NoContainerTechException
+from ..isolation_provider.container import (
+    Container,
+    NoContainerTechException,
+    NotAvailableContainerTechException,
+)
 from ..isolation_provider.dummy import Dummy
 from ..isolation_provider.qubes import Qubes, is_qubes_native_conversion
-from ..util import get_resource_path, get_subprocess_startupinfo, get_version
+from ..util import format_exception, get_resource_path, get_version
 from .logic import Alert, CollapsibleBox, DangerzoneGui, UpdateDialog
 from .updater import UpdateReport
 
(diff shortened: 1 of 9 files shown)


almet:
On MacOS, the tests are currently failing. I managed to reproduce on our mac mini. For a reason I still need to understand, the following command hangs indefinitely.

```bash
usr/local/bin/docker image list --format "{{.ID}}" dangerzone.rocks/dangerzone
```

And as a result the GUI continues showing that it's "installing" the image.

almet:
Tests are now passing (without doing any changes). I'm not sure if it was due to an error with the runners? :thinking:
I re-triggered them and they seem to be green each time now.

@apyrgio I believe this is ready for review.

dangerzone/gui/main_window.py
```diff
@@ -480,77 +492,78 @@ def check_state(self) -> None:
...
-                runtime_name = self.dangerzone.isolation_provider.get_runtime_name()
+            self.dangerzone.isolation_provider.is_runtime_available()
         except NoContainerTechException as e:
             log.error(str(e))
             state = "not_installed"
-
+        except NotAvailableContainerTechException as e:
+            log.error(str(e))
+            state = "not_running"
+            error = e.error
```
apyrgio, line 502: I would also add the following except clause:

```diff
+        except Exception as e:
+            log.error(str(e))
+            state = "not_running"
+            error = format_exception(e)
```

While toying with the PR (renaming the `docker.exe` file to something else), I managed to get an `OSError` exception here. It's best to catch everything, just to be on the safe side.

dangerzone/isolation_provider/container.py
```diff
@@ -175,19 +190,42 @@ def install() -> bool:
...
+            stderr=subprocess.PIPE,
+            startupinfo=get_subprocess_startupinfo(),
+        ) as p:
+            _, stderr = p.communicate()
+            if p.returncode != 0:
+                raise NotAvailableContainerTechException(runtime_name, stderr.decode())
+            return True
+
+    @staticmethod
+    def is_container_installed(raise_on_error: bool = False) -> bool:
```
apyrgio, line 226: Don't know about you, but I think this method is very confusing. It checks if the container is installed, but also somehow has a side-effect (removing the container image if the image ID does not match). I think we can do better here, but that would be out of scope of this PR.

dangerzone/gui/main_window.py
almet: Good catch! I've updated the code in 8da6d30af68979dc43239aa749071d967fa00441

dangerzone/isolation_provider/container.py
almet: I agree: I was this 🤏🏼 close to propose a change there. Next time we touch here, I guess :+1:

apyrgio (approved):

apyrgio:
Not much else to comment on, I think it's good to merge.

almet:
Tests were intermittently failing on MacOS and Windows, due to the way we tested threads, which was a bit brittle. I changed the approach to flatten threads during the tests (replacing `.start()` with `.run()`) to ensure the logic is tested properly.

If that's okay for you @apyrgio, and that tests are green on all targets, I'll wait until #748 is merged, rebase and we should be good.

PR #953, opened 2024-10-14
Add support for Fedora 41

### LINKED ISSUE #947: Fedora 41 RPM
Fedora 41 just came out, an RPM in the repo would be great.

If there is a place where the spec file lies, I could try to do it

apyrgio:
Add support for Fedora 41. A slight obstacle with this distro is that it has Python 3.13 installed, which is not supported by the PySide6 version shipped via PyPI. The `python3-pyside6` package has support for Python 3.13, so this is not a real problem for our end users. It's only a problem for dev environments, and this PR installs Python 3.12 to sidestep it.

Fixes #947

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,8 @@ jobs:
             version: "39"
           - distro: fedora
             version: "40"
+          - distro: fedora
+            version: "41"
 
     steps:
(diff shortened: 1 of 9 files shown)


almet:
That looks good to me!

almet (approved):

PR #954, opened 2024-10-15
Add support for Ubuntu 24.10

apyrgio:
Ubuntu 24.10 has recently been released, so we need to add support for it. In this PR, we include Ubuntu 24.10 in our CI pipelines, and fix some minor issues pertaining to our dev environments.


almet:
I believe I removed `tests/test_docs/sample-docm.docm` by mistake in the branch you based your work on, and it, but it should still be here.

Build currently failing because of:

```
Error: creating runtime static files directory "/home/user/.local/share/containers/storage/libpod": mkdir /home/user/.local/share/containers/storage: permission denied
```

apyrgio:
Whoops, you're right. Actually, it was me who erroneously removed that file :sweat_smile: . Also, I fixed the reason for the CI failure. I forgot to also delete the stock `ubuntu` user from the end-user environment as well.

almet (approved):
Amazing, it seems to work well :-)

PR #955, opened 2024-10-15
Automate the closing of stale issues

almet:
This automates the closing of stale issues, but marking them as stale after 14 days if inactivity and then closing them 30 days later. This is only for issues marked with the "need info" label.

### DIFF
--- a/.github/workflows/close-issues.yml
+++ b/.github/workflows/close-issues.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "Marking this issue as stale because it has been open for 30 days with no activity. It will be closed in 14 days if there's no activity, or if the `stale` label is not removed. Does anyone want to add something?"
+          close-issue-message: "Closing this issue now. Don't hesitate to reopen if you have anything to add :-)"
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/close-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
```
apyrgio, line 17: Maybe add something like:

"... It will close in 14 days, if there's no activity, or if the `stale` label is not removed. Does anyone want to add something?"

.github/workflows/close-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
```
apyrgio, line 20: Do we need to set this to `-1`? My understanding is that these values are unset by default.

.github/workflows/close-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          any-of-labels: need info
```
apyrgio, line 22: Nit: Should it be `needs info` perhaps?

.github/workflows/close-issues.yml
almet: That is a very good remark! I've updated the string in bb5f6df871dee17201d0bce89ac2b769b6be0ef9

.github/workflows/close-issues.yml
almet: Indeed! Updated.

.github/workflows/close-issues.yml
apyrgio: Cool, cool, better to be explicit.

apyrgio (approved):

apyrgio:
Thanks for the changes, let's merge!

PR #958, opened 2024-10-17
ci: Explicitly use Ubuntu 24.04 for our runner images

### LINKED ISSUE #957: Container image builds fail due to older Poetry version
GitHub has switched the `ubuntu-latest` runner images we're using from Ubuntu 24.04 to Ubuntu 22.04 :
* Runner image on Oct 16th was Ubuntu 24.04: https://github.com/freedomofpress/dangerzone/actions/runs/11368463360/job/31623614028#step:1:8
* Runner image today is Ubuntu 22.04: https://github.com/freedomofpress/dangerzone/actions/runs/11375940576/job/31647550993#step:1:8

This looks like a bug by GitHub, probably happening now because we're getting closer to the October 30th deadline, when all runners who specify `ubuntu-latest` will use Ubuntu 24.04 by default. See https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/.

We can't do much here, other than explicitly specify `ubuntu-24.04` as our runner image, and change it later on to `ubuntu-latest`.

apyrgio:
GitHub actions somehow managed to downgrade our runners from Ubuntu 24.04 to Ubuntu 22.04, even though we use `ubuntu-latest`. Make the Ubuntu 24.04 requirement more explicit, until GitHub migrates fully to this version for the `ubuntu-latest` tag.

Fixes #957

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
               build-dev --sync
 
   build-container-image:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
   # This is already built daily by the "build.yml" file
   # But we also want to include this in the checks that run on each push.
   build-container-image:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 


almet:
The problem seem to already have been reported upstream (https://github.com/actions/runner-images/issues/10796 and https://github.com/actions/runner-images/issues/10767).

On the short term, I think we should merge your fix, but I also think that using `ubuntu-latest` makes more sense in the long term, once the problem is fixed on the github CI.

almet (approved):

almet:
You can actually follow the actual discussion here: https://github.com/actions/runner-images/issues/10636

almet:
> but I also think that using ubuntu-latest makes more sense in the long term, once the problem is fixed on the github CI.

... Reading the thread actually made me change my mind: we'll be better off with using a specific tag and upgrading it when a new release is out.

As part of checking new releases and adding support for them, we could also bump the image used by our runners to reflect this.

PR #959, opened 2024-10-17
CI: Only run CI on branch or pull request events, but not twice.

almet:
The CI currently runs on both `pull-requests` and `branch` events. This is an attempt to only run on one or the other.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,10 @@
 name: Build dev environments
 on:
+  pull_request:
   push:
+    branches:
+      - main
+      - "test/**"
   schedule:
     - cron: "0 0 * * *" # Run every day at 00:00 UTC.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,10 @@
 name: Tests
 on:
-  push:
   pull_request:
-    branches: [main]
+  push:
+    branches:
+      - main
+      - "test/**"
   schedule:
     - cron: "2 0 * * *" # Run every day at 02:00 UTC.
   workflow_dispatch:
@@ -91,7 +93,8 @@ jobs:
 
   windows:
     runs-on: windows-latest
-    needs: download-tessdata
+    needs:
+      - download-tessdata
     env:
       DUMMY_CONVERSION: 1
     steps:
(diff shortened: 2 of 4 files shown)


almet:
Hmm, this doesn't seem to do what I want, as it's still running twice. 🤔 (Keeping it open in case somebody has a better idea, will garbage collect in a few otherwise).

almet:
This should be ready now for review!

.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Build dev environments
 on:
-  push:
+  pull_request:
```
apyrgio, line 3: I have found myself in positions where I wanted to run some CI tests for a branch, without necessarily sending a PR. Or, when I have sent a PR already, and I've made a big variation that I don't want to force-push to a reviewed branch. To accommodate for this case, I think we can have something like this:

```
on:
  push:
    branches:
      - main
      - "test/**"
  pull_request:
```

That is, either run the CI tests on branches named `test/**` (useful only for devs) or `main` (happens whenever we merge a PR), or on pull requests that target any branch.  As long as we don't send PRs with feature branches named `tests/**`, we will not see duplicate runs. What do you think?

I think this way, there's no need for the `should-run` job as well.

.github/workflows/ci.yml
```diff
@@ -24,7 +25,24 @@ concurrency:
...
+  should-run:
+    runs-on: ubuntu-latest
+    outputs:
+      run-workflow: ${{ steps.check.outputs.run-workflow }}
+    steps:
+      - id: check
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "run-workflow=true" >> $GITHUB_OUTPUT
+          elif [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" != "refs/heads/main" ]]; then
```
apyrgio, line 37: My understanding is that there's no push event that would trigger this PR, other than pushing to `main`, since you have already narrowed it down with the `on` trigger.

.github/workflows/build.yml
almet: Thanks, I've updated the pull request accordingly!

apyrgio (approved):

apyrgio:
Thanks, the changes look good. Let's merge!

PR #966, opened 2024-10-22
tests: Add a doc with multimedia elements

### LINKED ISSUE #917: File is too large?
When trying to convert my Pdf it stops and says "A page exceeded the maximum height". Please can somene explain to me if it has to do with the size of the Pdf, the size of the pages, or anything else. I need this for work and it would be wonderful if it worked. Thanks

apyrgio:
Add a doc that contains an MP4 video in it, which has an audio and video stream. This type of document could not be converted with the latest Dangerzone releases, because PyMuPDF threw this error in the container's stdout:

    MuPDF error: unsupported error: cannot create appearance stream for
    Screen annotations

This error message was treated literally by our client code, which parsed the first few bytes in order to find out the page height/width. This resulted to a misleading Dangerzone error, e.g.:

    A page exceeded the maximum height

This issue started occurring since 0.6.0, which added streaming support, and was fixed by commit 3f86e7b4654e0aea458b54248288eb9e2b1e86f8. That fix was not accompanied by a test document that would ensure we would not have this regression from now on, so we add it in this commit.

Refs #877
Closes #917


almet (approved):
Great, let's merge 👍

PR #967, opened 2024-10-22
qubes: Do not close stderr

apyrgio:
Do not close stderr as part of the Qubes termination logic, since we need to read the debug logs. This shouldn't affect typical termination scenarios, since we expect our disposable qube to be either busy reading from stdin, or writing to stdout. If this is not the case, then forcefully killing the `qrexec-client-vm` process should unblock the qube.

With this change, the Qubes CI tests should pass once more on `main`.


almet (approved):
Thanks!

PR #972, opened 2024-10-28
build: Publish the built artifacts.

almet:
- Fedora `.rpm` files
- Windows `.msi`
- macOS `.app`

Are now published as part of the CI pipelines.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,12 @@ jobs:
           sudo apt-get install -y python3-poetry
           python3 ./install/common/build-image.py
 
+      - name: Upload container image
+        uses: actions/upload-artifact@v4
+        with:
+          name: container.tar.gz
+          path: share/container.tar.gz
+
   download-tessdata:
     name: Download and cache Tesseract data
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -213,7 +234,7 @@ jobs:
         if: matrix.distro == 'debian' && matrix.version == 'bookworm'
         uses: actions/upload-artifact@v4
         with:
-          name: dangerzone-${{ matrix.distro }}-${{ matrix.version }}.deb
+          name: dangerzone-${{ matrix.distro }}-all.deb
```
apyrgio, line 237:
```suggestion
          name: dangerzone-${{ matrix.distro }}-all.deb
```

Any particular reason why we want to change the original naming? I'm asking because in this context, the `debian` and `all` tags don't help much (the `all` is a bit misleading as well, since the original file has `amd64` in its name instead).

apyrgio: We can leave it just `dangerzone.deb`, if we want to indicate that it's suitable for every Debian version.

apyrgio (commented):
Nice work! I had only a few nits, but other than that, it looks good to merge.

.github/workflows/ci.yml
almet: This is now fixed in the latest squashed commit of this PR

almet:
I've rebased this branch on top of the latest `main` and addressed your comments. Let me know what you think @apyrgio!

apyrgio (approved):

apyrgio:
Looks good, thanks!

PR #975, opened 2024-10-29
Add a script to help generate release notes from merged pull requests

almet:
I found myself browsing the closed pull requests to generate the release notes, and I figured a small script would help. Here is what I came up with. It's output cannot be taken as-is, but I found this helpful.

### DIFF
--- a/dev_scripts/generate-release-notes.py
+++ b/dev_scripts/generate-release-notes.py
@@ -1,0 +1,254 @@
+#!/usr/bin/env python3
+
+import argparse
+import asyncio
+import re
+import sys
+from datetime import datetime
+from typing import Dict, List, Optional, Tuple
+
+import httpx
+
+REPOSITORY = "https://github.com/freedomofpress/dangerzone/"
+TEMPLATE = "- {title} ([#{number}]({url}))"
+
+
+def parse_version(version: str) -> Tuple[int, int]:
+    """Extract major.minor from version string, ignoring patch"""
+    match = re.match(r"v?(\d+)\.(\d+)", version)
+    if not match:
+        raise ValueError(f"Invalid version format: {version}")
+    return (int(match.group(1)), int(match.group(2)))
+
+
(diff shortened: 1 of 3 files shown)


apyrgio:
That's nice actually, feel free to merge!

apyrgio (approved):

almet:
I've actually made some changes to the script so that it gets the closed issues, and link the issues rather than the PRs. It's now using httpx rather than requests to speedup the whole thing.

dev_scripts/generate-release-notes.py
```diff
@@ -0,0 +1,248 @@
...
+
+        pr_results = await asyncio.gather(*pr_tasks)
+        changes.extend([r for r in pr_results if r is not None])
+
+        return changes
+
+
+async def main_async():
+    parser = argparse.ArgumentParser(description="Generate release notes from GitHub")
+    parser.add_argument("--token", "-t", help="GitHub API token")
```
apyrgio, line 222: One minor comment here. It's best not to grab the access token from the CLI, because it will be stored in your shell's history (and can be grabbed by unprivileged processes in your system, but that's another thing). It's best to read it from a file instead, e.g., like here: https://github.com/freedomofpress/dangerzone/blob/d561878e030ca1eb91853b6c6e656930fceff9b4/dev_scripts/upload-asset.py#L80-L82

apyrgio:
Thanks a lot for the improvements. I can confirm that the scripts returns results quickly.

dev_scripts/generate-release-notes.py
almet: I actually don't get why it's better from a file than from the history, can you explain a bit more?

almet: This: https://security.stackexchange.com/a/190074 explains it well, actually.

almet: Updated in the latest commit. Will merge now :-)

PR #977, opened 2024-10-30
Drop support for Ubuntu Mantic (23.10), which is EOL since 11 Jul 2024.

almet:
Ubuntu mantic is now EOL since July 2024. This PR removes the support for it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,8 +34,6 @@ jobs:
           - distro: ubuntu
             version: "22.04"
           - distro: ubuntu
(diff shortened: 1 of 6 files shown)


apyrgio:
Thanks for the fix. Let's make sure to add a changelog entry for this removal.

almet:
> Thanks for the fix. Let's make sure to add a changelog entry for this removal.

I actually added it directly in #976 yesterday, hence why it wasn't present here. (did this to avoid conflicts)

Merging :-)

PR #980, opened 2024-10-30
tests: Restore previously mocked function

apyrgio:
Restore the `isolation_provider.base.kill_process_group()` function, which was previously mocked, at the end of the
`test_linger_unkillable()` test. This function is initially mocked, in order to simulate a hang process. After the mocking completes, the test needs the original function once more, in order to actually kill the spawned process.

### DIFF
--- a/tests/isolation_provider/base.py
+++ b/tests/isolation_provider/base.py
@@ -164,6 +164,7 @@ class IsolationProviderTermination:
         terminate_proc_mock = mocker.patch.object(
             provider, "terminate_doc_to_pixels_proc", return_value=None
         )
+        kill_pg_orig = base.kill_process_group
         kill_pg_mock = mocker.patch(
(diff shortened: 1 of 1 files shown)


tests/isolation_provider/base.py
```diff
@@ -164,6 +164,7 @@ def test_linger_unkillable(
         terminate_proc_mock = mocker.patch.object(
             provider, "terminate_doc_to_pixels_proc", return_value=None
         )
+        kill_pg_orig = base.kill_process_group
```
almet, line 167: I'm surprised: I thought that the whole reason for using pytest mock was so we don't have to do this ourselves.

almet: Ah: I get it with your initial PR description (maybe commit message): you need it once more after the original patch, but not after the test ends.

That looks good to me then. Note that we could also use it as a context manager (`with`) if we wanted.

almet (approved):

Issue #974, opened 2024-10-29
Executables built with cx_freeze broken after On-host pixels to PDF conversion PR was merged


jkarasti:
### What happened?

On-host pixels to PDF conversion merged in #748 uncovered an incompatibility between pumypdf and cx_freeze which causes an error when running `dangerzone.exe` or `dangerzone-cli.exe`  built with cx_freeze.

### operating system version

Windows 10 Pro 22H2 (OS Build 19045.5073)

### Dangerzone version

Top of main branch

### Additional info

Here's the error log I got:

```
Traceback (most recent call last):
  File "C:\Users\juhok\AppData\Local\pypoetry\Cache\virtualenvs\dangerzone-_8ncJGJf-py3.12\Lib\site-packages\cx_Freeze\initscripts\__startup__.py", line 140, in run
    module_init.run(f"__main__{name}")
  File "C:\Users\juhok\AppData\Local\pypoetry\Cache\virtualenvs\dangerzone-_8ncJGJf-py3.12\Lib\site-packages\cx_Freeze\initscripts\console.py", line 25, in run
    exec(code, main_globals)
  File "install\windows\dangerzone-cli.py", line 1, in <module>
  File "C:\GitHub\dangerzone\dangerzone\__init__.py", line 27, in <module>
    from .cli import cli_main as main
  File "C:\GitHub\dangerzone\dangerzone\cli.py", line 10, in <module>
    from .isolation_provider.container import Container
  File "C:\GitHub\dangerzone\dangerzone\isolation_provider\container.py", line 12, in <module>
    from .base import IsolationProvider, terminate_process_group
  File "C:\GitHub\dangerzone\dangerzone\isolation_provider\base.py", line 12, in <module>
    import fitz
  File "C:\Users\juhok\AppData\Local\pypoetry\Cache\virtualenvs\dangerzone-_8ncJGJf-py3.12\Lib\site-packages\fitz\__init__.py", line 2, in <module>
    from pymupdf import *
  File "C:\Users\juhok\AppData\Local\pypoetry\Cache\virtualenvs\dangerzone-_8ncJGJf-py3.12\Lib\site-packages\pymupdf\__init__.py", line 21706, in <module>
    from . import utils
ImportError: cannot import name 'utils' from partially initialized module 'pymupdf' (most likely due to a circular import) (C \GitHub\dangerzone\build\exe.win-amd64-3.12\lib\pymupdf\__init__.pyc)
```

almet:
[...]

PR #981, opened 2024-10-30
Fix cx_freeze not able to build .exe files (#974)

jkarasti:
Here's a quick workaround for #974 as suggested by @apyrgio  until the fix in upstream lands in a future release of cx_freeze.

Also revert an earlier workaround for an error that no longer triggers that I noticed while debugging this error.


almet (approved):
Thanks for these changes!

PR #989, opened 2024-11-06
Update the README links for 0.8.0

apyrgio:
Also, update our scan action for the released container image, to use the new filename in our assets.


almet (approved):
Looks good!

PR #961, opened 2024-10-17
CI: Update container scanning to account for the arm64 architecture.

almet:
This should be merged after the release for 0.8.0, so that new container images for `i686` and `arm64` are checked on.

PR #990, opened 2024-11-06
Remove version prefix `v` from container filename

apyrgio:


almet (approved):
Thanks, I didn't know about [substring extension](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) in bash. This seem to do the trick.

I was weighting in if we should add the `v` prefix [in the container asset name](#961) but probably we're good staying this way as we never use the `v` prefix except from the git tag.

Issue #968, opened 2024-10-23
CI flakiness with HWP files


apyrgio:
Our CI tests fail from time to time when converting our sample .hwp/.hwpx files. For example, see this CI test run: https://github.com/freedomofpress/dangerzone/actions/runs/11450930480/job/31859301008#step:10:991. This failures are intermittent, since the tests pass if we kick the CI job one more time.

The failed jobs don't report any helpful error message, so we need to dig deeper.

apyrgio:
We can run the HWP conversion locally in a tight loop:

```
$ while poetry run ./dev_scripts/dangerzone-cli <(base64 -d tests/test_docs_external/sample-hwp.hwp.b64) --output-filename /tmp/test-safe.pdf ; do sleep 1; done
```

I've left it running for a few minutes, and I haven't encountered an error yet. The logs I get from the container are:

```
----- DOC TO PIXELS LOG START -----
Installing LibreOffice extension 'h2orestart.oxt'
Converting to PDF using LibreOffice
Converting page 1/6 to pixels
Converting page 2/6 to pixels
Converting page 3/6 to pixels
Converting page 4/6 to pixels
Converting page 5/6 to pixels
Converting page 6/6 to pixels
Converted document to pixels
[COMMAND] unzip -d /usr/lib/libreoffice/share/extensions/h2orestart.oxt/ /libreoffice_ext/h2orestart.oxt
Archive:  /libreoffice_ext/h2orestart.oxt
  inflating: icon/H2Orestart.png
  inflating: description.xml
  inflating: description/desc_en.txt
  inflating: H2Orestart.jar
  inflating: description/desc_ko.txt
  inflating: registry/H2Orestart_types.xcu
  inflating: logger.properties
  inflating: registry/H2Orestart_filters.xcu
  inflating: registry/TypeDetection.xcu
  inflating: META-INF/manifest.xml
[COMMAND] libreoffice --headless --safe-mode --convert-to pdf --outdir /tmp /tmp/input_file
OpenJDK 64-Bit Server VM warning: Can't detect primordial thread stack location - find_vma failed
OpenJDK 64-Bit Server VM warning: Can't detect primordial thread stack location - find_vma failed
OpenJDK 64-Bit Server VM warning: Can't detect primordial thread stack location - find_vma failed
convert /tmp/input_file as a Writer document -> /tmp/input_file.pdf using filter : writer_pdf_Export
----- DOC TO PIXELS LOG END -----
```

[...]

PR #991, opened 2024-11-06
tests: mark the hancom office suite tests for rerun on failures

almet:
It seem that these tests are flaky, and as a result our CI pipeline is failing from time to time. This will rerun it automatically when there is an error.

It's not entirely clear to me if we want to rerun these tests automatically or not before we do further investigation. I personally think that it could help us having a clearer view of the status of the CI, rather than blindly rerunning the CI in case of failures.

See https://github.com/freedomofpress/dangerzone/issues/968 for more information

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -220,73 +220,73 @@ files = [
 
 [[package]]
 name = "coverage"
-version = "7.6.4"
+version = "7.6.7"
 description = "Code coverage measurement for Python"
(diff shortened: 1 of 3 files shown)


apyrgio (approved):

apyrgio:
Thanks Alexis. Let's merge it for now, but revisit at some point the initial issue.

PR #994, opened 2024-11-09
Unpin gVisor, now that upstream is able to support Linux Yama Mode 2

### LINKED ISSUE #298: Replace setuptools bdist_rpm before its deprecation
`bdist_rpm`, our current `.rpm` packaging tool is on [its way to the grave](https://github.com/pypa/setuptools/issues/1988). We need to replace it before deprecation. The deprecation reasoning is that the setuptools devs want to focus only [on supporting building for pypi](https://github.com/Cimbali/pympress/commit/a440f68c1584362d756d0e98c2b633a148d362cd).

almet:
Fixes #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## [Unreleased](https://github.com/freedomofpress/dangerzone/compare/v0.8.0...HEAD)
 
+### Added
+
+- Disable gVisor's DirectFS feature ([#226](https://github.com/freedomofpress/dangerzone/issues/226)).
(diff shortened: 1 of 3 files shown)


almet:
It seems [we're hitting a bug](https://github.com/freedomofpress/dangerzone/actions/runs/11759313428/job/32758682274) where [the python version in debian trixie is shown as `3.12.7+`](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1087011). This should be solved soon.

apyrgio:
Heh, that's a first... PR looks great, and I'm approving it right now. One small ask though, if you can also reinstate in our changelog that the disabling of DirectFS is a contribution by Etienne. That's all, thanks!

apyrgio (approved):

Issue #903, opened 2024-08-29
Consolidate documentation


almet:
Currently, the documentation lives within markdown files on different parts of the repository.

This has the advantage of being simple, but makes it harder to search, and sometimes I have trouble finding the information I'm looking for.

We could have a sphinx tool in place which would compile all this documentation and make it available somewhere, maybe making it available at https://docs.dangerzone.rocks

I see the following advantages:

- The doc will be easier to search
- As part of this effort, we could create an index of our documentation
- We could add some documentation, for instance to help newcomers contribute to the project.

(partial) todolist:

- [ ] Add a contributing documentation
- [ ] Get the list of supported versions out of README.md
- [ ] Have a roadmap available to users

almet:
We've been discussing this today, and we would need to decide what's the best way forward between these two options (I'm trying to summarize and comment on the proposal that have been made during our meeting).

So, two potential ways to handle the docs:

1. Have User facing docs and Contributors facing docs in one place, in a sphinx documentation, or;
2. Have User facing docs on the dangerzone.rocks website and the contributing docs in a sphinx documentation.

## 1. Have all the docs in one central place.

Pros:

- It's possible to do the documentation changes alongside the code changes
- It's possible to version the docs, so that it's possible to point to old version documentation if needed (in case we need to maintain multiple versions of dangerzone, which might or might not happen, e.g. when/if we have the Dangerzone packages done by upstream Debian)
- All

Cons:

- It might confuse the users to have a different interface than the website

## 2. Have the user and contributor docs in two different places

Pros:

- Users can see the user documentation directly inside the website, which becomes their only entrypoint to the project

Cons:

- We will only have one version of the docs (unless there is an option to do that on eleventy, which might be possible)
- Documentation changes will need to be tracked in separate PRs than code changes

There are probably other pros and cons, don't hesitate to update this comment with them. Also, I'm wondering if there is a potential third way where for instance we have sphinx use a theme that is close to what we have for the website, and where the developer facing changes are a bit hidden, for instance.

Also, I would like to mention two things that I think would be worth discussing:

- @eloquence mentioned that the SD docs are writting with `.rst` (restructured text) and our docs are currently done with Markdown. I'm comfortable with either, and Sphinx supports both. Would it be okay if we stay with Markdown, for instance, or should we move to restructuredText? (it's possible to do the conversion with Pandoc, but in my experience it needs manual intervention on the produced text)

[...]

PR #995, opened 2024-11-09
ci: check that the changelog is populated on each pull request

almet:
The goal is to gain time during the release process when building the CHANGELOG, by offloading this to the pull-requests.


.github/workflows/check_pr.yml
```diff
@@ -0,0 +1,35 @@
+name: Check branch conformity
+on:
+  pull_request:
+      types: [opened, synchronize, reopened]
```
apyrgio, line 4: Hm, [that's the default](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) actually, so how come we specify it again?

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

.github/workflows/check_pr.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: ensure CHANGELOG.md is populated
+        run: |
+          CHANGELOG_MODIFIED=$(git diff --name-only ${GITHUB_SHA} | grep -c "CHANGELOG.md")
+          if [ $CHANGELOG_MODIFIED -eq 0 ]; then
+            echo "::error::No CHANGELOG modifications found in this pull request. Please update the CHANGELOG."
```
apyrgio, line 33: Suggestion: Since not every PR needs to update the changelog (but most PRs do), I think we could:

1. Replace `::error::` with `::warning::` (see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message)
2. Point to our `CONTRIBUTING.md` guide, in a section where we explain to contributors how to do so.
    * (not sure how easy it is, let's not do it if it's hard) We could even propose a message, by taking the PR title (is it available?), the contributor name, and the PR number.

.github/workflows/check_pr.yml
almet: Yeap, that was me toying with it to understand why it wasn't triggering builds. I put it back on place.

.github/workflows/check_pr.yml
almet: Thanks for the proposed changes. I've updated the code to use a warning rather than an error, it's a good idea :+1: 

About updating the `CONTRIBUTING.md` guide, I haven't found anything like that. Maybe I'm not looking at the right place :thinking: ?

If we don't already have somewhere to put this information, I propose to just add a warning and enhance / create a contributing guide as part of #903.

apyrgio: My mistake, sorry! Yeap, what you're saying makes sense.

apyrgio (approved):

apyrgio:
Thanks, feel free to merge.

PR #996, opened 2024-11-11
Small fixes for issues we encountered during the 0.8.0 release

apyrgio:
This PR tackles the following minor issues:
1. Make the `dz` qube larger, now that we ask users to download Tesseract data. Initially the `/home` disk size was 2GiB, and with this change, we bump it to 5GiB.
2. Remove a reference to a second container from our README.
3. Update our QA script to implement two more steps from our instructions.


apyrgio:
@deeplow: One quick question about https://github.com/freedomofpress/dangerzone/pull/996/commits/8f8252eff810ef4723005d872a4664aab8f8d281. I haven't found a way to create an app qube **and** define the disk size at the same time, hence the separate command. Do you know of a more canonical method?

dev_scripts/qa.py
```diff
@@ -802,6 +813,31 @@ def _consume_stdin(self):
...
             msvcrt.getch()
 
+    @QABase.task(
+        f"Install the latest version of Python {PYTHON_VERSION_STR}", ref=REF_BUILD
+    )
+    def install_python(self):
+        cur_version = list(sys.version_info[:3])
+
+        logger.info("Getting latest Python release")
+        with urllib.request.urlopen(EOL_PYTHON_URL) as f:
```
almet, line 823: Maybe factorize this as a separate function? I think it will help readability.

dev_scripts/qa.py
```diff
@@ -802,6 +813,31 @@ def _consume_stdin(self):
...
+            resp = f.read()
+            releases = json.loads(resp)
+            for release in releases:
+                if release["cycle"] == PYTHON_VERSION_STR:
+                    latest_version = [int(num) for num in release["latest"].split(".")]
+                    if latest_version > cur_version:
+                        self.prompt(
+                            f"You need to install the latest Python version ({release['latest']})"
```
almet, line 831: Maybe put `release['latest']` in an intermediate variable?

almet (changes requested):
This looks good to me, except some of the comments I added, but they're mainly about readability.

dev_scripts/qa.py
apyrgio: Good call, I think I've simplified this in b63c1f4f5e1eaf23baebeec7243b977dd9dc1667.

dev_scripts/qa.py
apyrgio: Fixed in b63c1f4f5e1eaf23baebeec7243b977dd9dc1667.

dev_scripts/qa.py
```diff
@@ -3,14 +3,21 @@
...
 import selectors
 import subprocess
 import sys
+import urllib.request
+from pathlib import Path
 
 logger = logging.getLogger(__name__)
 
+PYTHON_VERSION_STR = "3.12"
+PYTHON_VERSION = [int(num) for num in PYTHON_VERSION_STR.split(".")]
```
almet, line 18: (nitpick, feel free to ignore if you don't feel it's worth it): 

I think we could have this done as a function rather than a constant, because it's actually converting something.

almet (approved):
Thanks, looks good. I've added a small nitpick, do what you want with it :-)

dev_scripts/qa.py
apyrgio: You're right, I'll factor this logic out.

apyrgio: Actually, `PYTHON_VERSION` was not used anywhere :grimacing:. So, I just removed it and renamed the `PYTHON_VERSION_STR` variable.

PR #1005, opened 2024-11-21
CI: Use grep + cut rather than jq to get the version number

almet:
The scans from yesterday are failing: https://github.com/freedomofpress/dangerzone/actions/runs/11944196277/job/33294696171

I hope this will make them run again :-) Github macOS runners don't come with `jq` pre-installed somehow (or not the right version? Not sure)


apyrgio:
Thanks, looks good.

PR #1012, opened 2024-11-26
Issue templates: rephrase how we ask `docker info` to the users

almet:
People might not now if the issue is related to docker or not, and we've had to ask them for additional information after they opened the issue.

This makes it clearer that this information might be useful.


apyrgio (approved):

apyrgio:
Ok, let's request this info more pressingly. Having a `--debug` flag though, or a `status-check` command is the best way forward I believe.

PR #1013, opened 2024-11-26
CI: Put `build-deb` as its own job

almet:
I wanted to speed up a bit the CI runs by not waiting for all the debian packages to build.

It turns out the way to implement this requires us to duplicate the job definition, because the feature [is not implemented in Github Actions](https://github.com/orgs/community/discussions/42335).

What do you think @apyrgio?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,7 +152,7 @@ jobs:
       - name: Run CLI tests
         run: poetry run make test
 
-  build-deb:
+  build-all-debs:
     needs:
       - build-container-image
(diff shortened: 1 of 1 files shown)


apyrgio:
Sigh, that's a big chunk to copy. Could we turn this into a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) perhaps? If it's too much of a hassle, I'd be in favor of just waiting a bit more. After all, my guess is that waiting for CI actions will be less of a hassle once we implement merge queues.

almet:
Interesting! Ultimately, I think the gain is not worth the effort for now. Closing this, we can reopen if we change our minds.

PR #1015, opened 2024-11-27
CI: check that our Github Actions are able to run on ubuntu-24

almet:
`ubuntu-latest` will point to `ubuntu-24` in the following months rather than `ubuntu-22`. This PR tries to change it beforehand to see if the CI passes, and try to fix the situation.

Related to https://github.com/freedomofpress/infrastructure/issues/5203


almet:
Closing this because for a reason I don't get, the runners aren't able to start (even on retries). Will try opening another PR to see if it's better.

PR #1017, opened 2024-11-28
Publish image attestations

almet:
This pull request introduces a new workflow to automate the release of container images. Each night, the worflow builds a container image and uploads it to the container registry.

> [!NOTE]
> Please note that, while signatures are involved in this workflow, the produced images are not signed by our release key. So, this is not to be confused with `cosign sign` commands, which are also used for independent container updates, but is different that the signatures involved when producing attestations. 

Here is a list of the produced assets:

- [resulting container image](https://github.com/freedomofpress/dangerzone/pkgs/container/dangerzone%2Fdangerzone)
- [transparency log](https://search.sigstore.dev/?logIndex=152228123)
- [attestation](https://github.com/freedomofpress/dangerzone/attestations/3486582)

Verifying the validity of the attestation can be done with the following commands:

_We should be able to also do it using [sigstore-python](https://github.com/sigstore/sigstore-python) and raw HTTP calls, but because the main goal here is to see how it works, I stopped there._

```bash
# First, login to the container registry.
# (We only need this because images are not publicly available yet)
# Enter "USERNAME" instead of your username
# and use your PAT as a password
regctl registry login ghcr.io

# Get the manifest from the latest tag
regctl manifest get --format raw-body ghcr.io/freedomofpress/dangerzone/dangerzone:latest > manifest.json

# The attestation for this manifest hash is available
# at the tag named "sha256-sha256(manifest.json)"
DIGEST="sha256-$(sha256sum manifest.json | awk '{ print $1 }')"
regctl artifact get ghcr.io/freedomofpress/dangerzone/dangerzone:${DIGEST} > bundle.json

# Finally verify that the attestation is the right one
cosign verify-blob-attestation --bundle bundle.json --new-bundle-format --certificate-oidc-issuer="https://token.actions.githubusercontent.com" --certificate-identity-regexp="^https://github.com/freedomofpress/dangerzone/.github/workflows/release-container-image.yml@refs/heads/test/image-publication-cosign" manifest.json
```

Related to [#745](https://github.com/freedomofpress/dangerzone/issues/745)


almet:
Here is [a script I wrote to download the attestations](https://gist.github.com/almet/de10e2b258df5a666c94fbb91be7e315) from a container registry for a specific tag.

almet:
While trying to build our own attestations (without using the GHA), we've been puzzled by how the container registry was populated in different ways.

It turns out that attestations are different from plain signatures:

1. The [Cosign Signature spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md) explains how plain signatures are represented. This is what's being used via `cosign sign`. It's using annotations (not to be confused with attestations), attaching them to the container registry. The signed payload are represented by the "simple signing" format.
2. The [Cosign Bundle Specification](https://github.com/sigstore/cosign/blob/main/specs/BUNDLE_SPEC.md) explains how image attestations are to be represented as OCI artifacts, and attached to the container registry. This is what's being used via `cosign attest`
	- Currently the [Attestation Specification](https://github.com/sigstore/cosign/blob/main/specs/ATTESTATION_SPEC.md) is used, which uses Layers to represent attestations and annotations to specify the verification material.
	- The cosign bundle spec is [currently being implemented](https://github.com/sigstore/cosign/pull/3889) in cosign but this isn't done yet.
---
What's been complex to understand is that cosign, in their documentation, points us to use `cosign sign` in order to sign our container images. This will add plain signatures and is different from an attestation.

(the [attest-build-provenance](https://github.com/actions/attest-build-provenance) this PR has been using uses the Cosign Bundle Specification)

almet:
Replaced with https://github.com/freedomofpress/dangerzone/pull/1080

PR #1031, opened 2024-12-13
More type stubs

jkarasti:
Spotted these while working on #1019. Mypy wants to install `types-colorama` and `types-pygments` for some extra type stubs. Mypy also suggested installing `types-setuptools` at some point, but doesen't anymore. Couldn't figure out why, but types-pygments pulls it in as a dependency anyways so it's now installed too.


apyrgio (approved):

apyrgio:
Looks good, merging it soon.

PR #1034, opened 2024-12-16
grype: Ignore CVE-2024-11053

apyrgio:
Ignore the CVE-2024-11053 vulnerability, since it's a libcurl one, and the Dangerzone container does not make network calls.

Also, clear the previous vulnerabilities, now that we have a new image out.


almet (approved):
LGTM, thanks!

PR #1041, opened 2025-01-06
build: add poetry-plugin-export to the dependencies

### LINKED ISSUE #825: Install poetry-plugin-export, as it will not be installed by default in future poetry versions
When generating the image with `install/common/build-image.py`, a warning is currently generated, telling us that the tool we currently use will not be included in future versions of poetry-export.

> Warning: poetry-plugin-export will not be installed by default in a future version of Poetry.
> In order to avoid a breaking change and make your automation forward-compatible, please install poetry-plugin-export explicitly. See > https://python-poetry.org/docs/plugins/#using-plugins for details on how to install a plugin.
> To disable this warning run 'poetry config warnings.export false'

We want to make sure it's installed, (or to use a different mechanism if it makes more sense)

almet:
Since Poetry 2.0.0, the `export` command has been removed and it's advised to use the "poetry-plugin-export".

This commit adds it to the `env.py` for debian environments and to the `CI`.

Fixes #825

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -17,7 +17,10 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install container build dependencies
-        run: sudo apt install pipx && pipx install poetry
+        run: |
+          sudo apt install pipx
+          pipx install poetry
+          pipx inject poetry poetry-plugin-export
       - name: Build container image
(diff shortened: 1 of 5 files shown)


almet:
The CI is currently failing because it doesn't update the dev environment which is modified in this PR, and so tries to run with the environments that were built during the night. I'll close this PR and open another one with branch named `test-*` so it will be rebuilt.

PR #1043, opened 2025-01-06
build: add poetry-plugin-export to the dependencies

### LINKED ISSUE #825: Install poetry-plugin-export, as it will not be installed by default in future poetry versions

almet:
Since Poetry 2.0.0, the `export` command has been removed and it's advised to use the "poetry-plugin-export" package instead.

This commit adds this dependency to the different places it's needed (debian environments, CI, build instructions etc).

I also took the chance to include `env.py` in the hashed files for the dev containers.

Fixes #825

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -17,7 +17,10 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install container build dependencies
-        run: sudo apt install pipx && pipx install poetry
+        run: |
+          sudo apt install pipx
+          pipx install poetry
+          pipx inject poetry poetry-plugin-export
       - name: Build container image
         run: python3 ./install/common/build-image.py --runtime docker --no-save
       - name: Get image tag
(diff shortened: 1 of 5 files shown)


apyrgio (approved):

apyrgio:
Looks good to me, feel free to merge. Minor note, there's a typo in the first commit of the PR (s/it's needsd/it's needed/).

PR #1051, opened 2025-01-15
CI: Drop Fedora 39 from the CI checks

almet:


almet:
Not sure why the bump of `poetry.lock` makes the tests fail, but it's probably orthogonal to the issue we're trying to fix here, so I will remove this last commit, we can do that in another PR.

apyrgio (approved):

apyrgio:
Cool, thanks.

PR #1053, opened 2025-01-15
CI: Store the signature key using the base64 format

### LINKED ISSUE #1052: Debian trixie repositories are marked as unsigned
The build instructions that we're testing each night [are currently failing](https://github.com/freedomofpress/dangerzone/actions/runs/12778635243/job/35660086627#step:6:18) with the following error:

```log
Warning: GPG error: https://packages.freedom.press/apt-tools-prod trixie Release: Sub-process /usr/bin/sqv returned an error code (1), error message is: Error: Failed to parse keyring "/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg"  Caused by:     0: Reading "/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg": EOF     1: EOF
Error: The repository 'https://packages.freedom.press/apt-tools-prod trixie Release' is not signed.
```

This only seem to happen on trixie. Currently investigating the cause.

almet:
The GPG binary format used until now doesn't seem to please `sqv` which is now used by default on debian trixie.

Fixes #1052

### DIFF
--- a/.github/workflows/check_repos.yml
+++ b/.github/workflows/check_repos.yml
@@ -53,8 +53,13 @@ jobs:
           gpg --keyserver hkps://keys.openpgp.org \
               --no-default-keyring --keyring ./fpf-apt-tools-archive-keyring.gpg \
               --recv-keys "DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281"
+
+          # Export the GPG key in armor mode because sequoia needs it this way
+          # (sqv is used on debian trixie by default to check the keys)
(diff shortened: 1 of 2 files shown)


almet:
Alternative implementation in https://github.com/freedomofpress/dangerzone/commit/1398c4d354822b2cdd67cdea857f73ce6117cf6c, which is using `sq` for trixie to retrieve the keys, but adds another branch of instructions for trixie users.

I'm currently looking if there are ways to configure `apt` to be able to tell `sqv` to read the keys in binary format somehow.

apyrgio (approved):

PR #1054, opened 2025-01-16
feat(doc): Mention `sq` as a way to retrieve the PGP key

almet:
As @legoktm mentioned, we could include this `sq` command in our installation doc.

See #1052


apyrgio:
This PR is probably superseded by #1056.

almet:
> This PR is probably superseded by #1056, right?

Yep it is! Closing.

PR #1056, opened 2025-01-20
doc: update Debian Trixie installation instructions

### LINKED ISSUE #1055: Current installation instructions are failing for debian / ubuntu
The CI job that checks our installation instructions is [currently failing](https://github.com/freedomofpress/dangerzone/actions/runs/12848504366/job/35825916478#step:6:20) for debian and debian-derivatives, with the following error message:

```bash
Warning: GPG error: https://packages.freedom.press/apt-tools-prod oracular Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY F3ACC60F62EA51CB
```

This is a continuation of https://github.com/freedomofpress/dangerzone/issues/1052

almet:
Starting with Debian Trixie, `apt secure` relies on `sqv` to do its verification, which doesn't support the GPG keybox database format.

In [a previous episode](https://github.com/freedomofpress/dangerzone/pull/1053) we switched to the base64 version of the key, but this resulted in all other debian-versions not able to follow our instructions (see #1055 for more details).

The reason is that the standard PGP base64 format makes the verification fail for versions of `apt secure` which relies on `gpg`, as the subkey is not detected there.

This PR takes two different routes for versions which rely on `sq` and versions which rely on `gpg` for the verification.

Fixes #1055, supersedes #1054

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -84,9 +84,20 @@ Dangerzone is available for:
   </tr>
 </table>
 
-Add our repository following these instructions:
+First, retrieve the PGP keys.
 
-Download the GPG key for the repo:
+Starting with Trixie, follow these instructions to download the PGP keys:
+
+```bash
+sudo apt-get update && sudo apt-get install sq -y
+mkdir -p /etc/apt/keyrings/
+sq network keyserver \
+   --server hkps://keys.openpgp.org \
+   search "DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281" \
+   --output /etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg
+```
+
+On other Debian-derivatives:
 
 ```sh
 sudo apt-get update && sudo apt-get install gnupg2 ca-certificates -y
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -84,9 +84,20 @@ Dangerzone is available for:
   </tr>
 </table>
 
-Add our repository following these instructions:
+#### Retrieve the PGP keys
```
apyrgio, line 87: Did you plan to add a section as well for installing Dangerzone? Else, maybe we can use normal text here.

apyrgio:
Thanks for working on this. I manually triggered the failing workflow, and it seems to fail again unfortunately: https://github.com/freedomofpress/dangerzone/actions/runs/12866962625/job/35870604006 (I'd suggest enabling this check to run just for this PR). Any idea why?

INSTALL.md
almet: Thanks, updated in c52da93c90ce4d9ca9bf7eb7a2c8e47e6c77ac17

almet:
> Any idea why?

I believe two things are happening: a) Some ubuntu mirrors are broken b) Might be missing the `ca-certificates` dependency. Added it, let's see how it goes!

almet:
The last changes made it work, see https://github.com/freedomofpress/dangerzone/actions/runs/12868475665

apyrgio (approved):

apyrgio:
Grrreat!

PR #1059, opened 2025-01-23
Use platformdirs to find user configuration files

### LINKED ISSUE #1058: `python3-appdirs` is missing from Trixie: build instructions fail
CI tests testing our installation instructions on various platforms are failing this morning for Debian Trixie.

The cause is that `python3-appdirs`, which is a direct dependency of dangerzone, has disappeared from Trixie (see https://packages.debian.org/search?searchon=names&keywords=python3-appdirs).

The rationale is mentioned [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1060427):

> This is dead upstream and easily replacable with platformdirs.  Rather
than release trixie with appdirs, remaining users should port to
platformdirs instead.

almet:
The previous library we were using for this (`appdirs`) is dead upstream
and not supported anymore in debian testing.

Fixes #1058

---

I haven't bumped the `poetry.lock` file here, and so the CI should be failing because of it, but the reason is that updating it makes the CI fail anyway. We probably should fix it in another PR.


almet:
(rebased on top of #1060 to have the CI green. We can probably merge the two together if it makes it easier to ship)

apyrgio:
Awesome, thanks Alexis!

apyrgio (approved):

PR #1060, opened 2025-01-23
build: pin shiboken6 to specific versions

almet:
At the moment, shiboken6 issued a version only for macOS, and poetry wrongfully considers it as the latest version, leading to a situation where we need to pin it for specific platforms to the latest known version.

That's what we did as part of the `0.8.1` release, but we haven't backported this to the main branch. When doing so, I found that the Fedora packages have been updated to this "semi-broken" version (but only for F40, not F41), so I updated accordingly the `.spec` file to pick any shiboken version for now.

The CI is failing for debian trixie due to #1058.


apyrgio:
Looks good, but something tells me we need to revisit this pin once they fix this issue upstream, right? Maybe create an issue once we merge this PR.

apyrgio (approved):

almet:
Yeah, when a new version will be issued, I believe we should change the way we pin, and remove the hack in the `.spec` file!

Issue #1071, opened 2025-02-05
Create patched 0.8.1 package for Debian Trixie


apyrgio:
Debian Trixie has recently switched from Python 3.12 to Python 3.13.  When doing so, a package named `python3-appdirs` was removed, and now `python3-platformdirs` should be installed instead (see https://github.com/freedomofpress/dangerzone/issues/1058). We have made this change in our `main` branch, and now we must release a `.deb` for our Debian Trixie users.

The release plan is the following:
- [x] Create a Git branch from the [`v0.8.1`](https://github.com/freedomofpress/dangerzone/tree/v0.8.1) tag
   * We have created [`0.8.1-trixie`](https://github.com/freedomofpress/dangerzone/tree/0.8.1-trixie) for this purpose.
- [x] Cherry-pick commits from the `main` branch that are relevant to Debian Trixie:
  - 6ae91b0 - Use `platformdirs` to find user configuration files
  - 40fb657 - Alternatives for debian/control
  - 7409966 - Remove `${Python3:Depends}` as it's not used at the moment.
  - 88a6b37 - Add support for Python 3.13
  - 00e58a8 - build: add poetry-plugin-export to the dependencies
- [x] Bump the Debian package version from `0.8.1-1` to `0.8.1-2`
- [ ] Make sure that the CI passes, at least for Debian Trixie
  * The CI [fails](https://github.com/freedomofpress/dangerzone/actions/runs/13160775742/job/36728663259), but only for the Fedora builds, which kinda makes sense because we haven't cherry-picked these commits.

almet:
The current format of our debian package is a "native" package, but that might well be a mistake, as native packages are used for "software that has n independent exeistence outside of Debian":

> A native source package is one that does not distinguish between Debian packaging releases and upstream releases. A native source package contains a single tar file of source material, and the versioning does not have a Debian-specific component. Native packages are normally (but not exclusively) used for software that has no independent existence outside of Debian, such as software written specifically to be a Debian package.
> — https://www.debian.org/doc/debian-policy/ch-source.html#s-source-packages

Could going back to a non-native package be the way to go here? 🤔

apyrgio:
It seems that we can switch to non-native Debian packages by using the `3.0 (quilt)` format, and building binary-only packages. I have sent a PR for this: https://github.com/freedomofpress/dangerzone/pull/1072

legoktm:
[...]

PR #1072, opened 2025-02-05
debian: Switch to quilt format

apyrgio:
Switch to quilt package format, and use `dpkg-buildpackage -b` to build the new Python package.

Refs #1071


almet:
Given the discussions that happened on #1071 we will instead stay with the native format in the future. Closing this (thanks!)

PR #1076, opened 2025-02-10
grype: Ignore CVE-2025-0665

apyrgio:
Ignore the CVE-2025-0665 vulnerability, since it's a libcurl one, and the Dangerzone container does not make network calls. Also, it seems that Debian Bookworm is not affected.


almet (approved):
Thank you Alex for the addition to our grype ignore list!

PR #1084, opened 2025-02-19
Remove a stale Shiboken6 pin

### LINKED ISSUE #1061: Remove the pin for shiboken6
Shiboken6 issued version `6.8.1.1`  in last december. This version was released only for macOS, as as such makes poetry fail to work on other OSes.

We've pinned this version for our 0.8.1 release and now back-ported the changes on the `main` branch:

https://github.com/freedomofpress/dangerzone/blob/b2f4e2d5235cc20354d030370ff6c62d3875f607/pyproject.toml#L28-L32

Fedora is the only distro we know that issued a release for this 6.8.1.1 version. The `.spec` file for Fedora packaging replaces the information in `pyproject.toml` by any version:

https://github.com/freedomofpress/dangerzone/blob/b2f4e2d5235cc20354d030370ff6c62d3875f607/install/linux/dangerzone.spec#L229-L233

Once a new version of shiboken is out we should remove these: https://pypi.org/project/shiboken6/#history

apyrgio:
Remove the Shiboken6 pin for our Linux and macOS platforms, since a new upstream package has been released, that has wheels for every platform.

Also, remove the `sed` command from our dangerzone.spec, whose purpose was to nullify this pin for our Fedora packages.

Fixes #1061

### DIFF
--- a/install/linux/dangerzone.spec
+++ b/install/linux/dangerzone.spec
@@ -216,12 +216,6 @@ convert the documents within a secure sandbox.
 %prep
 %autosetup -p1 -n dangerzone-%{version}
 
-# Bypass the version pin for Fedora as the 6.8.1.1 package is causing trouble
-# A 6.8.1.1 package was only released with a wheel for macOS, but was picked by
-# Fedora packagers. We cannot use "*" when PyPI is involved as it will fail to download the latest version.
-# For Fedora, we can pick any of the released versions.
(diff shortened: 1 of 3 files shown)


almet (approved):
Great, thanks!  I was first surprised to not see `shiboken6 = *` in the dependencies, but it's in fact not needed as it's a dependency of `Pyside6`.

Issue #663, opened 2024-01-03
Consider switching from gzip to lzma


apyrgio:
Consider changing the compression algorithm from Gzip to LZMA for our container image. Currently, our 1.3GiB container image gets compressed down to ~620 MiB with Gzip. With LZMA, we can further reduce it to ~500MiB:

#### LZMA compression (default level) -> 503.9MiB in 7 minutes

```
lzma --keep --verbose share/container.tar
share/container.tar (1/1)
  100 %     503.9 MiB / 1,273.2 MiB = 0.396   3.0 MiB/s       7:00
```

#### LZMA compression (best level) -> 495.4MiB in 9 minutes

```
lzma --keep --verbose --best share/container.tar
share/container.tar (1/1)
  100 %     495.4 MiB / 1,273.2 MiB = 0.389   2.4 MiB/s       9:00
```

#### LZMA decompression in 24 seconds

```
unlzma --keep --verbose share/container.tar.lzma
share/container.tar.lzma (1/1)
  100 %     503.9 MiB / 1,273.2 MiB = 0.396    52 MiB/s       0:24
```

This ties well with the discussion regarding switching to a container image that can further reduce the download size (see https://github.com/freedomofpress/dangerzone/issues/658#issuecomment-1861239821).

almet:
We do not compress the images anymore, as implemented in https://github.com/freedomofpress/dangerzone/commit/51f432be6b8b5ce4e69a3bd4db16a987a3d5c370, and so we don't need to change the compression algo! Closing.

Issue #1085, opened 2025-02-26
Un-vendor `repro-build` script


apyrgio:
We are about to vendor the `repro-build` script from https://github.com/freedomfopress/repro-build into our codebase. The reason we're doing that is to invoke it and change it easier, at the risk of having these files unsynced.

Once the interface of this script is stabilized, consider fetching it from the repo somehow, either with an HTTP call, or using Git submodules.

PR #1086, opened 2025-02-26
Build and reproduce a multi-platform Dangerzone image in GitHub actions

### LINKED ISSUE #1035: Implement container image attestations
Attestations are ways to attest the **provenance** of the image. Here is an [example of an attestation](https://github.com/freedomofpress/dangerzone/attestations/3486582) (processed by GitHub).

In our case, we want to ensure that these fields are the expected ones:

| Field                   | Value                                                                                         |
| ----------------------- | --------------------------------------------------------------------------------------------- |
| `Source Repository URI` | https://github.com/freedomofpress/dangerzone                                                  |
| `Runner Environment`    | `github-hosted`                                                                               |
| `Build Signer URI`      | `https://github.com/freedomofpress/dangerzone/.github/workflows/release-container-image.yml*` |

> [!NOTE]
> **Cosign Bundle Specification**
>Multiple formats exist to attach the attestations to a container registry. We are following the format and mechanism described by [the Cosign Bundle Spec](https://github.com/sigstore/cosign/blob/main/specs/BUNDLE_SPEC.md)

[Here is an example implementation](https://github.com/freedomofpress/dangerzone/pull/1017) of publishing images with their attestations via Github Actions, and for manual steps to verify the validity of the attestation.

### LINKED ISSUE #1074: Ensure container image reproducibilty across different container runtimes and versions
For independent container updates, we plan to build our container image in a GitHub CI action and attach provenance info to it (#1035). In order to reproduce these builds, we already have a script in place that uses `diffoci` and its "semantic" reproducibility mode (see #1047).

So far, we were able to reproduce images that where built by the exact same container runtime. It seems that different runtimes though, and even different versions of the same runtime, can add annotations to the image that break reproducibility. We need to figure out a way to compare images in a way that small details like these (if they are small of course), don't break our scripts.

apyrgio:
This PR significantly improves the reproducibility properties of our container image, and showcases them by building and reproducing a multi-platform image on GitHub Actions

Fixes #1035
Fixes #1074
Refs #1085


.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
+name: Release multi-arch container image
```
almet, line 1: It would be great to be able to reference this from other workflows. You can find more about how to do that [in the following documentation](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow)

In my opinion, the inputs should be the ones you specify in the `env` section. The need for me is to be able to reuse this workflow building the image on multiple platforms, adding a signature to it in an automated way (probably in the tests themselves).

.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
...
+      - "test/**"
+  schedule:
+    - cron: "0 0 * * *" # Run every day at 00:00 UTC.
+
+env:
+  REGISTRY: ghcr.io/${{ github.repository_owner }}
+  REGISTRY_USER: ${{ github.actor }}
+  REGISTRY_PASSWORD: ${{ github.token }}
+  IMAGE_NAME: dangerzone/dangerzone
+  BUILDKIT_IMAGE: "docker.io/moby/buildkit:v19.0@sha256:14aa1b4dd92ea0a4cd03a54d0c6079046ea98cd0c0ae6176bdd7036ba370cbbe"
```
almet, line 17: Should we actually get this from the same place we define the version of the tools we use? Currently this is [`Dockerfile.env`](https://github.com/freedomofpress/dangerzone/blob/main/Dockerfile.env), but maybe this can change so we are defining everything in the same location.

.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dev. dependencies
+        run: |-
+          sudo apt-get update
+          sudo apt-get install -y git python3-poetry --no-install-recommends
+          poetry install --only package
+
+      - name: Verify that the Dockerfile matches the commited template and params
```
almet, line 31: good idea to do that in here. I believe it should not be under the `lint` job, as it's more than that :-)

.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
...
+  check-reproducibility:
+    needs:
+      - prepare
+      - merge
+    runs-on: ubuntu-24.04${{ matrix.platform.suffix }}
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - suffix: ""
```
almet, line 219: Since the suffix is only used to decide the runner, maybe it would be better to directly specify the runner?

.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
...
+
+  # This step calls the container workflow to generate provenance and push it to
+  # the container registry.
+  provenance:
+    needs:
+      - prepare
+      - merge
+    strategy:
+      matrix:
+        digest:
```
almet, line 193: I don't understand what this is, it says it's a digest, but it looks like a platform 🥸

.github/workflows/release-container-image.yml
```diff
@@ -0,0 +1,237 @@
...
+      digest: ${{ needs.merge.outputs[format('digest_{0}', matrix.digest)] }}
+      image: ${{ needs.prepare.outputs.image }}
+      registry-username: ${{ github.actor }}
+    secrets:
+      registry-password: ${{ secrets.GITHUB_TOKEN }}
+
+  # This step ensures that the image is reproducible
+  check-reproducibility:
+    needs:
+      - prepare
```
almet, line 212: Should `prepare` be a requirement of `merge`?

Dockerfile
```diff
@@ -2,14 +2,14 @@
 # Dockerfile args below. For more info about this file, read
 # docs/developer/reproducibility.md.
 
-ARG DEBIAN_IMAGE_DATE=20250113
+ARG DEBIAN_IMAGE_DATE=20250224
 
-FROM debian:bookworm-${DEBIAN_IMAGE_DATE}-slim as dangerzone-image
+FROM debian:bookworm-${DEBIAN_IMAGE_DATE}-slim AS dangerzone-image
```
almet, line 7: Happy linter is a happy.

Dockerfile
```diff
@@ -165,20 +169,47 @@ RUN mkdir /home/dangerzone/.containers
...
-RUN mkdir /new_root/root /new_root/run /new_root/tmp
-RUN chmod 777 /new_root/tmp
+RUN mkdir -p \
+    /new_root \
+    /new_root/root \
+    /new_root/run \
+    /new_root/tmp \
+    /new_root/home/dangerzone/dangerzone-image/rootfs
+
+# XXX: Remove /etc/resolv.conf, so that the network configuration of the host
```
almet, line 185: This `XXX` seem to be solved (the remove is done a few lines below)

dangerzone/container_utils.py
```diff
@@ -120,30 +127,46 @@ def get_expected_tag() -> str:
...
-    with gzip.open(compressed_container_path) as f:
-        while True:
-            chunk = f.read(chunk_size)
-            if len(chunk) > 0:
-                if p.stdin:
-                    p.stdin.write(chunk)
+    tarball_path = get_resource_path("container.tar")
+    with open(tarball_path) as f:
+        try:
+            res = subprocess.run(
```
almet, line 133: Would need to check on windows and macOS, but there a possibility that you don't even need to untar the file.

In this case, just passing its location to the `{podman,docker} load -i` invocation would be enough (so no need to `open` the tarball path

dangerzone/isolation_provider/qubes.py
```diff
@@ -130,7 +130,7 @@ def is_qubes_native_conversion() -> bool:
         # This disambiguates if it is running a Qubes targetted build or not
         # (Qubes-specific builds don't ship the container image)
 
-        compressed_container_path = get_resource_path("container.tar.gz")
+        compressed_container_path = get_resource_path("container.tar")
```
almet, line 133: I guess we can also rename `compressed_container_path` to `container_tarball`

dev_scripts/reproduce-image.py
```diff
@@ -4,138 +4,80 @@
...
-
-def build_image(tag, use_cache=False):
+def build_image(
+    platform=None,
+    runtime=None,
+    cache=True,
+    date=None,
+):
     """Build the Dangerzone container image with a special tag."""
+    platform_args = [] if not platform else ["--platform", platform]
```
almet, line 34: In this specific case, I think it doesn't make sense to have intermediary variables. But this is very nitpicky, so feel free to address or not. I would feel it's easier to read with everything inline the `run` call (on separate lines).

almet (changes requested):
Hey, thanks @apyrgio from the work on this. Amazing!

I've added comments inline, hope they make sense!

dev_scripts/repro-build
```diff
@@ -0,0 +1,665 @@
+#!/usr/bin/env python3
+
```
almet, line 2: All the other files in this folder have `.py` suffix. Should we stick with it?

dev_scripts/repro-build
```diff
@@ -0,0 +1,665 @@
...
+    info = oci_parse_manifest(tar, path, platform)
+    parsed.append(info)
+    for m in info["manifests"]:
+        oci_parse_manifests_dfs(tar, m["digest"], parsed, m.get("platform"))
+
+
+def oci_parse_tarball(path: Path) -> dict:
+    parsed = []
+    with tarfile.TarFile.open(path) as tar:
+        oci_parse_manifests_dfs(tar, "index.json", parsed)
```
almet, line 267: When running this tool on a local archive, it wasn't able to find `index.json` or `blobs/...`, running a `tar.list()` shown that they are present but need to be accessed with the `./` prefix.

Maybe should should have a way to try both in case a `KeyError` is raised?

almet:
Also, these changes would fix #663 since we're not using gzip anymore

.github/workflows/release-container-image.yml
apyrgio: Hm, currently this is copied verbatim from the `repro-build` script, so the source of truth is there. But we could do as you say, maybe with some comments on what's the purpose of `Dockerfile.env`.

.github/workflows/release-container-image.yml
apyrgio: Oh, it's actually the type of the manifest. `root`, for the top-level manifest list of the image and `amd64` / `arm64` for the platform-specific manifests. So, I can rename it to `manifest` if you agree.

.github/workflows/release-container-image.yml
apyrgio: It is, but we need to define it here as well, in order to get the value with `${{ needs.prepare... }}`.

Dockerfile
apyrgio: Right, the purpose of `XXX` here is to explain what's the need for removing `/etc/resolv.conf` in the following lines, since it will come as a surprise for most people.

dangerzone/container_utils.py
apyrgio: Oh yeah, you're right. I'll just pass the file location with `-i <path>`.

dev_scripts/reproduce-image.py
apyrgio: Hm, the thing here is the handling of `None` values. E.g., what happens if you don't specify a platform? It's not easy to handle this inline in the `subprocess.run` call.

dev_scripts/repro-build
apyrgio: Huh, that's interesting. Sure, I guess I can try it. Out of curiosity, can you show me the structure of the tar archive, as reported by `tar`? Here's an example:

```
tar -tvf ~/Downloads/container.tar
drwxr-xr-x 0/0               0 1970-01-01 02:00 blobs/
drwxr-xr-x 0/0               0 1970-01-01 02:00 blobs/sha256/
-r--r--r-- 0/0             722 1970-01-01 02:00 blobs/sha256/10817f4cd6b0f02392c7eb2d39ef82261ea685b62413b1bfc10c39c3b4a2b45a
-r--r--r-- 0/0             504 1970-01-01 02:00 blobs/sha256/9c10b61876843c3a381f534dcbf3817fbe4b7d6adf12abae8a8b2371d0fbf6dd
-r--r--r-- 0/0       478679035 1970-01-01 02:00 blobs/sha256/cd51d257f86733615c3330c84a518bc43a29365b3897d7ead72aadb9996cfa23
-rw-r--r-- 0/0             524 1970-01-01 02:00 index.json
-rw-r--r-- 0/0             254 1970-01-01 02:00 manifest.json
-r--r--r-- 0/0              30 1970-01-01 02:00 oci-layout
```

.github/workflows/release-container-image.yml
almet: The goal is to be able to run this with a slightly different target, for the upcoming tests on ICUs, where the image will need to be published somewhere so we can test the overall image release and install cycle works, so I would say it's a reusable workflow with `workflow_call`

.github/workflows/release-container-image.yml
almet: Yeah, and I don't like having duplicated strings like that, because they might get out of sync, and it's double work when updating it.

If the `Dockerfile.env` doesn't seem the right place, we could also could also rename it to something else.

.github/workflows/release-container-image.yml
almet: Actually reading again about it, `lint` now seems to be thing it's doing :brain:. Sorry, let's keep it like that.

.github/workflows/release-container-image.yml
almet: `manifest_type` sounds like a good candidate!

.github/workflows/release-container-image.yml
almet: Then maybe `merge` should pass the outputs so we can avoid this?

dev_scripts/repro-build
almet: Here you go:

```
drwx------ alexis/alexis     0 2025-03-03 17:17 ./
drwxr-xr-x alexis/alexis     0 2025-03-03 17:17 ./blobs/
drwxr-xr-x alexis/alexis     0 2025-03-03 17:19 ./blobs/sha256/
-rw------- alexis/alexis   147 2025-03-03 17:18 ./blobs/sha256/0045939fad64c6c29fddd969ca0cecc020edd51301c8f15b5af1be5a62f1c2d3
-rw------- alexis/alexis   234 2025-03-03 17:17 ./blobs/sha256/04e697ba65bfbdae7409ff984f2762e3dfd5055ad591e9739b968db3e6196d4a
-rw-r--r-- alexis/alexis  2377 2025-03-03 17:18 ./blobs/sha256/099bb54a180c8aecdb4686cc474adeda7e8b97c85e73652c6fbbe334f8dd6330
-rw------- alexis/alexis   192 2025-03-03 17:18 ./blobs/sha256/1675277906a2efcf65eb316f5f7a215fe6976729d417865821f509fa7323ef3e
-rw------- alexis/alexis  2479 2025-03-03 17:17 ./blobs/sha256/265c500f045adffd708db2bc63f7f8ca7b6878d2cd3b8a84bb01619e82644846
-rw-r--r-- alexis/alexis  3712 2025-03-03 17:19 ./blobs/sha256/2824272007d49a260a5612fd6e11391a10f80d3a40578a203fea2bdf384e4ae8
-rw------- alexis/alexis 462967279 2025-03-03 17:18 ./blobs/sha256/36b20ca9490d3a3d1b08cf762f6b8c64ddd3c8a41ef79ada323ba5c7537abe29
-rw------- alexis/alexis       251 2025-03-03 17:19 ./blobs/sha256/3ed05bea76eaf30fac570e78e80efd143bed171c3ec0691b6057b7a0034a238e
-rw-r--r-- alexis/alexis      3708 2025-03-03 17:18 ./blobs/sha256/49914d7f3c8ab91408d3a0c8af2a6880532c0055d65374ab7c4801f2bfcafc35
-rw-r--r-- alexis/alexis      1621 2025-03-03 17:19 ./blobs/sha256/4a3f44189387dffe8ac8e1bc1cd3b1b5e3c7a324fcd568f5e7c199bba4e07e84
-rw------- alexis/alexis       235 2025-03-03 17:18 ./blobs/sha256/4befef7a42a3ac8f6e0ddd84e9946427004f125d8838c1556fcc9dadd26b72c0
-rw------- alexis/alexis    513944 2025-03-03 17:18 ./blobs/sha256/4d758dc6f1e2e654da5dd03de34db956f5914c4539f2939ee7408fa7ac6ae964
-rw------- alexis/alexis    587987 2025-03-03 17:17 ./blobs/sha256/52a46e94d0916512d53af06c02ab9b595d76a5a930c0046119c99d6dcad49f08
-rw------- alexis/alexis       191 2025-03-03 17:18 ./blobs/sha256/544d23d53059846e138ede895e4c6df237ebf183814753f72aa0805a4141f52f
-rw------- alexis/alexis    587485 2025-03-03 17:17 ./blobs/sha256/55fcc934571edebb7dc237ec2c359171400adf4f493749b29adc3f9c817a443c
-rw-r--r-- alexis/alexis     10315 2025-03-03 17:19 ./blobs/sha256/56bbf5d45dade2a24e973658a8df66ca754cc26bf11fae77966dae71e642e5ab
-rw------- alexis/alexis       148 2025-03-03 17:17 ./blobs/sha256/5db6895957891ea602069c08b8fa9ee1ead6feb12431a8718c3cc3f0c1683c95
-rw------- alexis/alexis 473547525 2025-03-03 17:19 ./blobs/sha256/5ea5752adb5abb520ef46e82ba3ad97f2e26103b76152f732b8aab6b67a69b0d
-rw------- alexis/alexis     11936 2025-03-03 17:19 ./blobs/sha256/6b9e861e18d74107a413828814825e41e8d7e869979a538fc981b1e8c634b691
-rw------- alexis/alexis    513958 2025-03-03 17:17 ./blobs/sha256/733b69d151f29c9c2cfd677af6ee12480b48a6dcf751b6508ccf8d0c76f6d87e
-rw-r--r-- alexis/alexis      2377 2025-03-03 17:19 ./blobs/sha256/752bb0410f7f30d67355fe97ebc0b8413daf7686eb944549c3e118a87fa2955c
-rw------- alexis/alexis       191 2025-03-03 17:17 ./blobs/sha256/797880d96bbd9aa8c2f51813472f530937c1138887f746bccf9a6d31d5369a4a
-rwxr-xr-x alexis/alexis      1609 2025-03-03 17:19 ./blobs/sha256/7b21dbdebffed855621dfcdeaa52230dc6566997f852ef5d62b0338b46796e01
-rw------- alexis/alexis    587885 2025-03-03 17:18 ./blobs/sha256/846b33c634e989a0a941c7cf25da92124f5fa9ca181b11041d3f8e4f704ae5a3
-rw-r--r-- alexis/alexis       567 2025-03-03 17:19 ./blobs/sha256/8888acdfd202d8df14241d1a230fd5c1ed70356d1eaccf0486348a1c1bb7a629
-rw-r--r-- alexis/alexis       167 2025-03-03 17:18 ./blobs/sha256/99f8b396bb5d24f02d6cc76c21d370c7920616555f7c673ea29dc6f702de3e40
-rw------- alexis/alexis       238 2025-03-03 17:18 ./blobs/sha256/9aa67d6d73707ab71725c1c93325c8b385461b6bc0e6e8d0badec9f5de087042
-rw------- alexis/alexis       236 2025-03-03 17:17 ./blobs/sha256/9c6f603389ae3b65d779c458d2da68676670d0bc3375c8a6701c5156378a9714
-rw------- alexis/alexis       190 2025-03-03 17:17 ./blobs/sha256/aa541d8c7333cf96851375eb6e2546a3f5a6bef1a0370e4cc1a69c70ffa19707
-rw------- alexis/alexis   1805987 2025-03-03 17:17 ./blobs/sha256/b9435c50e03db60bc6e83bb4f1e7d0e851321bca6417efd32c1deca6c1daa22d
-rw------- alexis/alexis     47773 2025-03-03 17:19 ./blobs/sha256/bc1d0dfe8b8e94923c3b0e7a1385e10ffec3f255f14d1a5a5c3fb16cfd0709a3
-rw-r--r-- alexis/alexis       233 2025-03-03 17:19 ./blobs/sha256/cbabac12486b113f06b66320b573c37d28e735b4a893bac8e2fa24a88ee05fbc
-rw-r--r-- alexis/alexis       167 2025-03-03 17:19 ./blobs/sha256/cbe0e81a449386e8b539dc994128005e3d1d86a6a61710f68920cf3dbc9d50f1
-rw-r--r-- alexis/alexis       567 2025-03-03 17:18 ./blobs/sha256/cc9c1c546c07166056a393922df88d0de8d052dad1ec89d288fc382b74088243
-rw------- alexis/alexis    587383 2025-03-03 17:18 ./blobs/sha256/d8d1f312f8f212de4cd908097223dc1828db7922c1e84adeb1b6651655658fde
-rw-r--r-- alexis/alexis       233 2025-03-03 17:19 ./blobs/sha256/e834c38b8462da230536f36a7fc334cfcf20349f6cb23ea7d06e19e7d8793e37
-rw------- alexis/alexis     47774 2025-03-03 17:18 ./blobs/sha256/ee2e22235cc86a6b949876ffa258d847c081f4d9c3ffb7b4382985fde5141183
-rw------- alexis/alexis   1809851 2025-03-03 17:18 ./blobs/sha256/ee43b57efca69b214596a1856fb449d9082ff5763e241f7357b9c9128349a510
-rwxr-xr-x alexis/alexis       984 2025-03-03 17:19 ./index.json
-rwxr-xr-x alexis/alexis        37 2025-03-03 17:17 ./oci-layout
```

.github/workflows/release-container-image.yml
apyrgio: Done in 2e2b6cf30841aee7b887816adab98c6164f26574.

.github/workflows/release-container-image.yml
apyrgio: Done in af8ba7429425ea0027ae5adfb393e071d59a2f55.

.github/workflows/release-container-image.yml
apyrgio: Done in ab0bee568821183be7faa0dae098b3bbded97981

.github/workflows/release-container-image.yml
apyrgio: Sure, we can proxy them in the `merge` job, but this way I fear that we may have to change things in two places in the future :face_with_diagonal_mouth:. If you feel it's better though, I can go ahead with that.

Dockerfile
apyrgio: Good point. Done in 8ca1357a415d11b844d48dffc3ddb791f68176a1.

dangerzone/container_utils.py
apyrgio: Done in c0ff351a2ed5708b6dd548a8c551f594ec0ee9e7

dangerzone/isolation_provider/qubes.py
apyrgio: Done in 5f4f82a

dev_scripts/repro-build
apyrgio: Done in 094d876dbaae6ce51654283bbcde8609e239b887.

dev_scripts/repro-build
apyrgio: Done in bb66e9a.

dev_scripts/reproduce-image.py
apyrgio: I may get it wrong, but do you mean something like this perhaps?

```python
cmd = [
          "podman",                                                                                                                                                                                                                           
          "run",                                                                                                                                                                                                                              
          "-it",                                                                                                                                                                                                                              
          "--rm",                                                                                                                                                                                                                             
          "-v",                                                                                                                                                                                                                               
          "buildkit_cache:/tmp/cache",                                                                                                                                                                                                        
          "-v",                                                                                                                                                                                                                               
          f"{output.parent}:/tmp/image",                                                                                                                                                                                                      
          "-v",                                                                                                                                                                                                                               
          f"{context}:/tmp/work",                                                                                                                                                                                                             
          "--entrypoint",                                                                                                                                                                                                                     
          "buildctl-daemonless.sh",
          ...
          *["--opt", f"platform={platform}"] if platform else *[],
          ...
]
```

I'm using here the star operator (`*`) to unpack the lists into positional arguments to the `cmd = [...]` constructor. This doesn't work though, because Python fails with `SyntaxError: Conditional expression cannot be used here`. 

Are you thinking of something else perhaps?

.github/workflows/release-container-image.yml
almet: Eww yeah!

.github/workflows/release-container-image.yml
almet: Thanks. I feel the added comments have too little value, we probably should remove them though.

.github/workflows/release-container-image.yml
almet: Use your best judgement, I'm fine with both. If we let it like that, a comment would go a long way.

Dockerfile
almet: I'm not sure to get it. The `XXX` is still there. Maybe an error?

dev_scripts/reproduce-image.py
almet: Ah! That's what I was missing! 

The other way around it would be conditional `if`s building the list, but now that I see it, I kind of like it this way. I was missing the list expansion. Marking this as solved.

.github/workflows/release-container-image.yml
apyrgio: Fixed in f1702ab5609ba88481d6c5c2729d6295d06d391b

.github/workflows/release-container-image.yml
apyrgio: Ok cool, removed them in cede19ca90e75f78d0a53259ae0f41a06847fc81

Dockerfile
apyrgio: I changed the comment phrasing to show that removing `resolv.conf` is not a todo item, but something that we intentionally do, for reasons that may not be immediately clear. The `XXX` is there as a signal for our future selves (or a passer-by), to alert them about this unexpected action.

.github/workflows/release-container-image.yml
apyrgio: Take a look at d7c01e755acd2f0f10aea4bd3d5975c7e4b34b99. If it feels weird, I'll just proxy the arguments.

Dockerfile
almet: Sorry, but then I don't understand why we have it show as a todo list. If it's not one, why not just putting a comment?

apyrgio: Could it be that we interpret `XXX` differently? To me, `XXX` means "Alert, we're doing something unexpected here". They are not `FIXME` or `TODO` tags, because it may be something that's just the way it is, albeit surprisingly so.

That's why I'm using it to explain the removal of `resolv.conf`. Do you interpret it the same way, or do you view it as a `TODO` thing?

almet: Interesting, that's the bottom of it: it doesn't have the same meaning for you and me. To me, `XXX` is an alias for `TODO` or `FIXME`.

I've read a bit more on the subject, and [this stackoverflow post](https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments) was interesting.

Could replacing the `XXX` with `NOTE` work for you?

apyrgio: Sure, I can go with that. For me, `NOTE` indicates something more benign in nature, so I don't tend to use it in lines where changing something may impact security. I prefer to grab the reader's attention with `XXX`.

I do understand though that this preference may not be shared, so I'll try to adjust and use `NOTE` from now on. Check out e715bd6.

.github/workflows/release-container-image.yml
almet: Yeah, it feels a bit weird to me, but I'm okay with it if you prefer it this way.

apyrgio: Check out 2b6736f97856293fd8e4a4bea647e3b0bba3778b

almet (approved):
Thanks for all the changes, I think we're good to proceed on this one (of course, once the FIXUPs commits are squashed together).

:raised_hands:

PR #1089, opened 2025-03-05
tests: look for regressions when converting PDFs

### LINKED ISSUE #321: Verify PDF output of CI tests
We currently assume that CI tests work because they complete successfully. However, we don't actually validate their output, and we may miss PDFs that have been converted, but are corrupted (see https://github.com/freedomofpress/dangerzone/issues/315).

In order to improve this situation, since the input is always known, we should also include the output in our tests folder, and compare it with that. The problem here is that we can't perform a byte to byte comparison. We need some sort of visual diffing, and ensure that the output PDF is not significantly different from the input PDF.

almet:
Adds a way to visually diff the generated PDFs with a reference version.

The comparison is done between pixel buffer arrays. Differences are uploaded as a github artifact ([example](https://github.com/freedomofpress/dangerzone/actions/runs/13835710153/artifacts/2745560897)).

Note that some already existing tools exist to do visual diffing, but none of them seemed to have good packaging, so I went with doing diffing myself, which is why this adds a dependency to numpy to run the tests. It's questionable and could be either be done differently or run optionally (only if numpy is present).

The PR currently fails to show how it would be like. There is a new `make regenerate-pdf-reference` target to update the reference version.

<details><summary>Details about other tools for PDF diffing</summary>
<p>

There are multiple tools that allow to do that, but I found that [diff-pdf](https://github.com/vslavik/diff-pdf) good and able to generate an output we can look at without having to run a GUI.

```bash
diff-pdf /tmp/pytest-of-alexis/pytest-current/sample-docx0.pdf ./tests/test_docs/reference/sample-docx.pdf -m --output-diff=diff.pdf
```

Produces this [diff.pdf](https://github.com/user-attachments/files/19091320/diff.pdf) file for the changes between the 0.8.1 release and this commit.

</p>
</details> 

Fixes #321


tests/test_cli.py
```diff
@@ -190,10 +193,66 @@ def test_invalid_lang(self, sample_pdf: str) -> None:
         result = self.run_cli([sample_pdf, "--ocr-lang", "piglatin"])
         result.assert_failure()
 
+    @pytest.mark.reference_generator
```
apyrgio, line 196: Note, we should register this mark, else `pytest` throws this warning:

```
tests/test_cli.py:196
  /home/user/dangerzone/tests/test_cli.py:196: PytestUnknownMarkWarning: Unknown pytest.mark.reference_generator - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.reference_generator
```

See also: https://docs.pytest.org/en/stable/how-to/mark.html#registering-marks

tests/test_cli.py
```diff
@@ -190,10 +193,66 @@ def test_invalid_lang(self, sample_pdf: str) -> None:
...
+    @for_each_doc
+    def test_formats(self, doc: Path, tmp_path_factory: pytest.TempPathFactory) -> None:
+        reference = (doc.parent / "reference" / doc.stem).with_suffix(".pdf")
+        destination = tmp_path_factory.mktemp(doc.stem).with_suffix(".pdf")
+
+        result = self.run_cli([str(doc), "--output-filename", str(destination)])
         result.assert_success()
+        # When needed, regenerate the reference PDFs by uncommenting the following line:
+        # reference.parent.mkdir(parents=True, exist_ok=True)
+        # shutil.copy(destination, reference)
```
apyrgio, line 213: This can be removed, right?

apyrgio (commented):
Woot, amazing work :partying_face:. Left a couple of comments, but other than that, the PR looks great.

tests/test_cli.py
almet: Thanks, check out 68505bd625859ecdef7a06ac1734728f4e58ba79

tests/test_cli.py
almet: Right! Did so in c28b58fe69a99acb2283fd13bbc791d5471e0438

tests/test_cli.py
apyrgio: I still see this warning here: https://github.com/freedomofpress/dangerzone/actions/runs/13990359427/job/39172782169?pr=1089#step:10:940

I think you need to switch `-` with `_` in the mark.

apyrgio (approved):

apyrgio:
Looks great. I've approved it, even though we have a small remaining issue for the mark warning. Once you resolve this, feel free to resolve my comment and merge this PR. Thanks!

PR #1096, opened 2025-03-10
ci: Security scan ARM images

### LINKED ISSUE #1008: Container image security scanning doesn't work for Silicon macOS
We added security scanning for the images shipped for ARM64, but it's [currently failing](https://github.com/freedomofpress/dangerzone/actions/runs/11982280906/attempts/1) due to the fact Docker is not installed in the runners.

Docker is installable (see [the instructions in this issue](https://github.com/actions/runner/issues/1456)) but doesn't come installed by default.

Doing this would also check that the latest Docker Desktop is able to load the container image, which also has some value.

One other option is to use colima, which is [already installed in the mac runners](https://github.com/actions/runner-images/issues/6216).

apyrgio:
Scan ARM images using Anchore's scan action, by utilizing the Ubuntu ARM runners provided by GitHub. While our ARM images are used only in macOS silicon platforms, we can use the Ubuntu ARM runners just for scanning.

CI jobs:
- Scan released app and container: https://github.com/freedomofpress/dangerzone/actions/runs/13769808346
- Scan latest app and container: https://github.com/freedomofpress/dangerzone/actions/runs/13769808387

Closes #1008

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -10,7 +10,12 @@ on:
 
 jobs:
   security-scan-container:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
+    runs-on: ${{ matrix.runs-on }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


almet (approved):
Yay, thanks!

Issue #965, opened 2024-10-22
Deprecate Ubuntu Focal support


apyrgio:
## Problem description

PyMuPDF recently [released](https://github.com/pymupdf/PyMuPDF/releases/tag/1.24.12) version 1.24.12, which drops support for Python 3.8. This affects us because Ubuntu Focal has Python 3.8 by default. Here's a bit of a context on why this is important:

- From 0.8.0, we will start vendoring PyMuPDF in our Debian/Ubuntu packages. The version we will vendor will be the latest one from PyPI, same as we do for Windows/macOS (see #625)
- We cannot rely on the system-provided PyMuPDF packages, because they don't offer OCR support, and we have encountered segfaults while developing with them.
- Ubuntu Focal provides Python 3.9, but we can't use it because it can't import the PySide2 system package. It has some libraries which are compiled for Python 3.8 specifically.

At the same time, Ubuntu Focal will be EOL in ~5 months from now (see https://endoflife.date/ubuntu).

## Suggestion

Since we're nearing the Ubuntu Focal EOL date, it's probably best to cease active development on that platform, and inform our users about it. To this end, we will:

1. Add a warning message when people start the Dangerzone GUI, which will effectively state that:
   * the current release (0.8.0) will be the last release to support Ubuntu Focal and Python 3.8 environments in general.
   * inform users that Dangerzone is available on the newer Ubuntu versions.
2. Unpublish the Dangerzone Debian package from our `focal` APT [repo](https://github.com/freedomofpress/apt-tools-prod).
   * This should take place at least 30 days after the 0.8.0 release, or when we release the next Dangerzone version, whichever is longer.

almet:
This seem to be a good way forward.

One question I asked myself / alternative path: should we continue supporting the `0.8.X` release series for Focal? It would mean updating the image at the same time of the future releases, and bumping the minor there, until Focal support is really EOL, if we want to not drop support before then.

I'm curious about how many downloads we have for Focal. Maybe with these numbers we could make an informed decision?

eloquence:
[...]

PR #1097, opened 2025-03-10
Unpin PyMuPDF and drop Ubuntu Focal support

### LINKED ISSUE #1018: Unpin PyMuPDF and drop Ubuntu Focal support
We have already announced as part of the 0.8.0 release that we will be dropping support for Ubuntu Focal ahead of time (#971). When we do it, we can unpin the PyMuPDF version.

apyrgio:
Drop Ubuntu 20.04 (Focal) support, because it's nearing its end-of-life date. By doing so, we can remove several workarounds and notices we had in place for this version, and most importantly, remove the pin to our vendored PyMuPDF package.

Fixes #1018
Refs #1018
Refs #965


dangerzone/gui/main_window.py
```diff
@@ -55,13 +55,6 @@
 HAMBURGER_MENU_SIZE = 30
 
 
-WARNING_MESSAGE = """\
```
almet, line 58: Rather than removing the code, could it be better to keep it for future announces we would like to make, but for now it doesn't trigger?

dev_scripts/env.py
```diff
@@ -115,15 +97,7 @@
...
-# NOTE: `pipx install poetry` fails on Ubuntu Focal, when installed through APT. By
-# installing the latest version, we sidestep this issue.
-RUN bash -c 'if [[ "$(pipx --version)" < "1" ]]; then \
-                apt-get update \
-                && apt-get remove -y pipx \
-                && apt-get install -y --no-install-recommends python3-pip \
-                && pip install pipx \
-                && rm -rf /var/lib/apt/lists/*; \
-              else true; fi'
+RUN pipx install poetry
```
almet, line 100: :raised_hands:

almet (approved):
Thanks for the changes. Maybe we should keep the mechanism in place to display announces, but other than that looks good.

dangerzone/gui/main_window.py
apyrgio: I was thinking about this as well. Generally, I prefer not to keep unused code for future-proofing, because it's pretty easy for surrounding code to change, and then you have to maintain the unused code as well.

I believe that if we do need something similar in the near future, we should be able to retrieve the code from the Git logs / PR. If such a needs happens much later on, i.e., after our UI revamp, then it will need a rewrite in any case.

Happy to hear your thoughts though, because I do find this code useful.

almet: Yeah. It's not a big deal anyway as it's really not a lot of code, and we currently know it's here if we need to reproduce something like it in the future.

PR #1099, opened 2025-03-13
Document Operating System support

### LINKED ISSUE #986: Clearly define operating system and version coverage
In order for our maintenance effort to be manageable, we should clearly define:

- which operating systems we support, and at what level (e.g., stable vs. experimental)
- which particular versions or distributions of said operating systems (e.g., "the most recent non-stable release and the two most recent LTS")

We should also define our stance for unsupported operating systems/distributions:

- Do we generally advise users to build from source?
- Do we accept PRs to add support or fix compatibility issues?

almet:
The goal is to have rules rather than specific versions, and a table to summarize everything.

Fixes #986 

The rendered table looks like this:
image

Happy to discuss anything about it!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,7 +1,41 @@
+## Operating System support
+
+Dangerzone can run on various Operating Systems (OS), and has automated tests
+for most of them.
+This section explains which OS we support, how long we support each version, and
+how do we test Dangerzone against these.
+
+You can find general support information in this table, and more details in the
+following sections.
+
+(Unless specified, the architecture of the OS is AMD64)
+
+| Distribution  | Supported releases        | Automated tests        | Manual QA |
+| ------------  | ------------------------- | ---------------------- | --------- |
+| Windows       | 2 last releases           | 🗹 (`windows-latest`) ◎ | 🗹         |
+| macOS intel   | 3 last releases           | 🗹 (`macos-13`) ◎       | 🗹         |
+| macOS silicon | 3 last releases           | 🗹 (`macos-latest`) ◎   | 🗹         |
+| Ubuntu        | Follow upstream support ✰ | 🗹                      | 🗹         |
+| Debian        | Current stable, Oldstable and LTS releases | 🗹     | 🗹         |
+| Fedora        | Follow upstream support   | 🗹                      | 🗹         |
+| Qubes OS      | [Beta support](https://github.com/freedomofpress/dangerzone/issues/413) ✢ | 🗷 | Latest Fedora template |
+| Tails         | Only the last release   | 🗷 | Last release only |
+
+Notes:
+
+✰ Support for Ubuntu Focal [was dropped](https://github.com/freedomofpress/dangerzone/issues/1018)
+
+✢ Qubes OS support assumes the use of a Fedora template. The supported releases follow our general support for Fedora.
+
+◎ More information about where that points [in the runner-images repository](https://github.com/actions/runner-images/tree/main)
+
 ## MacOS
 
 - Download [Dangerzone 0.8.1 for Mac (Apple Silicon CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-arm64.dmg)
 - Download [Dangerzone 0.8.1 for Mac (Intel CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-i686.dmg)
+
+> [!TIP]
+> We support the releases of macOS that are still within Apple's servicing timeline. Apple usually provides security updates for the latest 3 releases, but this isn’t consistently applied and security fixes aren’t guaranteed for the non-latest releases. We are also dependent on [Docker Desktop windows support](https://docs.docker.com/desktop/setup/install/mac-install/)
 
 You can also install Dangerzone for Mac using [Homebrew](https://brew.sh/): `brew install --cask dangerzone`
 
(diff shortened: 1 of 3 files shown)


INSTALL.md
```diff
@@ -1,5 +1,35 @@
+## Operating system support
+
+Dangerzone can run on a varieties of Operating systems, and have automated tests for a subset of them.
+This section explains what is the support you should expect from us.
+
+You can find general support information in this table, and more details in the following sections.
+
+(Unless specified, the architecture is AMD64)
```
apyrgio, line 8:
```suggestion
## Operating System support

Dangerzone can run on various Operating Systems, and has automated tests for most of them.
This section explains which Operating Systems we support, how long we support each version, and how do we test new releases against these systems. 

You can find general support information in this table, and more details in the following sections.

(Unless specified, the architecture of the Operating Systems we support is AMD64)
```

INSTALL.md
```diff
@@ -1,5 +1,35 @@
...
+
+| Distribution | Supported versions        | Automated tests        | Manual QA         |
+| ------------ | ------------------------- | ---------------------- | ----------------- |
+| Windows      | 3 last releases           | Last release only      | Last version only |
```
apyrgio, line 12: I'm not sure about the 3 last releases part, given that we currently support Windows 10 and Windows 11. Which is the third one

(also, minor, we refer to "releases", and then the last column refers to a "version")

INSTALL.md
```diff
@@ -1,5 +1,35 @@
...
+| macOS intel  | 3 last releases           | Last release only      | Last release only |
+| macOS ARM    | 3 last releases           | Last release only      | Last release only |
```
apyrgio, line 14: For macOS with Intel CPU, our automated tests take place on macOS 13. The same applies to our manual tests. For macOS ARM (I'd refer to it as "Apple Silicon" though), the manual tests take place on macOS 13 as well.

Since the available devices can change from time to time though, I'd prefer a checkmark here than a specific version.

INSTALL.md
```diff
@@ -1,5 +1,35 @@
...
+| Debian       | Current stable, Oldstable and LTS releases | 🗹     | 🗹                |
+| Fedora       | Follow upstream support   | 🗹                      | 🗹                |
+| Qubes OS     | [Beta support](https://github.com/freedomofpress/dangerzone/issues/413) ✢ | 🗷 | Latest Fedora template |
+| Tails        | Only the latest version   | 🗷 | Last release only |
+
+Notes:
+
+✰ Support for Ubuntu Focal [was dropped](https://github.com/freedomofpress/dangerzone/issues/1018)
+
+✢ Qubes OS support is done via the use of other OSes (e.g. Fedora or Debian) for the VMs, so the versions we support are tied to them.
```
apyrgio, line 25:
```suggestion
✢ Qubes OS support assumes the use of a Fedora template. The supported versions follow our general support for Fedora.
```

INSTALL.md
```diff
@@ -1,5 +1,35 @@
...
+
 ## MacOS
 
+We support the versions of macOS that are still within Apple's servicing timeline. Apple usually provides security updates for the latest 3 releases, but this isn’t consistently applied and security fixes aren’t guaranteed for the non-latest releases.
+
+We have automated tests for the latest version of macOS only.
+
 - Download [Dangerzone 0.8.1 for Mac (Apple Silicon CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-arm64.dmg)
 - Download [Dangerzone 0.8.1 for Mac (Intel CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-i686.dmg)
```
apyrgio, line 34: What bothers me slightly is that the installation instructions and the support instructions are meshed together. What if we moved the support instructions under a tooltip?

```suggestion
- Download [Dangerzone 0.8.1 for Mac (Apple Silicon CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-arm64.dmg)
- Download [Dangerzone 0.8.1 for Mac (Intel CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.8.1/Dangerzone-0.8.1-i686.dmg)

> [!TIP]
>
> We support the versions of macOS that are still within Apple's servicing timeline. Apple usually provides security updates for the latest 3 releases, but this isn’t consistently applied and security fixes aren’t guaranteed for the non-latest releases.
>
> We have automated tests for the latest version of macOS only.

apyrgio: Also, I think we should point to the relevant Docker Desktop section as well: https://docs.docker.com/desktop/setup/install/mac-install/

INSTALL.md
almet: Thanks, I've updated this in c0a4f388b36669d0fdbe3726b2f2cc47e7cdca28

INSTALL.md
almet: I've updated to "2 last versions" then.

About the version VS releases. I decided to stick to "version" everywhere, as I believe it's clearer this way. See in c0a4f388b36669d0fdbe3726b2f2cc47e7cdca28

INSTALL.md
almet: I've added some more information about how we run our tests, and a link to the github runners, see c0a4f388b36669d0fdbe3726b2f2cc47e7cdca28

INSTALL.md
almet: Thanks! Updated in c0a4f388b36669d0fdbe3726b2f2cc47e7cdca28

INSTALL.md
almet: Thanks, updated in 87ec9b73c1f198e278276b60edff11712ecf9d91

INSTALL.md
almet: Hmm, and in the end I went to opposite direction as what we're supporting is more releases than versions, if I read this correctly. See b04135897b7cc1999339254970894f2fb3cae6b

apyrgio:
I think we're close to a final version of this document. Before merging it, I'd like to give the opportunity to @DeltaEpsilon19498 to comment on this PR, and let us know if their questions were covered. We can't unfortunately support everything, but we at least want to accurately document what we currently support. Let us know if there is something we may be missing.

INSTALL.md
apyrgio: The thing is that our test devices are on an earlier version, so it's not accurate :-/. So, either we should be more accurate (and keep this section in mind when updating our devices), or just add a tick there, to show that we manually test Dangerzone on macOS.

almet: Ah, gotcha!

For macOS, I'm always testing on the latest macOS release (on my `m1` machine), but this is not true for intel mac and for windows.

I've updated everything with a checkbox, and squashed the commits.

apyrgio: Makes sense, thanks!

apyrgio (approved):

PR #1102, opened 2025-03-17
Improve our release instructions regarding versioned links

### LINKED ISSUE #1100: Installation docs in README point to old version
[https://github.com/freedomofpress/dangerzone?tab=readme-ov-file t](https://github.com/freedomofpress/dangerzone/blob/main/README.md?plain=1#L17-L23) for macOS and Windows points (via an intermediate page) to https://github.com/freedomofpress/dangerzone/blob/v0.8.1/README.md#macos

Unfortunately, it looks like the tagged version of the `README.md` hasn't been updated to point to v0.8.1; it still links to v0.7.1.

apyrgio:
Update our `RELEASE.md` so that we don't forget to bump the download links in `INSTALL.md` prior to tagging a release. This way, we won't have a versioned `INSTALL.md` page pointing to an older download link.

Note that this means that the latest version of the `INSTALL.md` page will point to a broken link, in the short period of time between the pre-release and the actual release. That's not an issue in our case, because we don't point to the latest version of our `INSTALL.md` from our `README.md`. We use versioned links instead, and thus we minimize the chance that a user may encounter a broken link.

Fixes #1100

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,6 +17,7 @@ Here is a list of tasks that should be done before issuing the release:
 - [ ] Bump the Debian version by adding a new changelog entry in `debian/changelog`
 - [ ] [Bump the minimum Docker Desktop versions](https://github.com/freedomofpress/dangerzone/blob/main/RELEASE.md#bump-the-minimum-docker-desktop-version) in `isolation_provider/container.py`
 - [ ] Bump the dates and versions in the `Dockerfile`
+- [ ] Update the download links in our `INSTALL.md` page to point to the new version (the download links will be populated after the release)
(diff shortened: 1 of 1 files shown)


almet:
Thanks.

In the "release" section, I think we should mention that we could check that the docs are up to date, especially the release links.

Other than that, looks good!

apyrgio:
I've added this line in our Release section:

> - [ ] Update version and links to our installation instructions (`INSTALL.md`) in `README.md`

I specified `README.md` because we don't point to a release link from a different place (and `INSTALL.md` is already covered by our pre-release instructions). Want me to add something extra there?

almet (approved):
Thanks!

PR #1103, opened 2025-03-17
Avoid passing wrong options `-B` to subprocesses

### LINKED ISSUE #873: Dangerzone help message printed when we open the Dangerzone app via the CLI in macOS
While testing the 0.7.0 release on macOS, we were sometimes opening the Dangerzone application via the CLI. When we were doing so and were converting a document, we were surprised to see that the help message of Dangerzone was printed to the terminal, and a `-B` argument was provided.

This does not affect the conversion, but it may be a sign of a larger problem. Also this is something that most of our users will never encounter, so it's a low priority issue.

almet:
This is a common pitfall of pyinstaller, when using multiprocessing.

In our case, the spawned processes is passed the `-B` option, thinking it's python (but it's dangerzone), so it fails. 

In case you are wondering, `-B` is an option you can pass python to avoid generating `.pyc` files at runtime:

> -B     Don't write .pyc files on import. See also PYTHONDONTWRITEBYTECODE.

From the pyinstaller docs:

> In the frozen application, sys.executable points to your application
> executable. So when the multiprocessing module in your main process
> attempts to spawn a subprocess (a worker or the resource tracker), it
> runs another instance of your program, with the following arguments for
> resource tracker:
>
> my_program -B -S -I -c "from multiprocessing.resource_tracker import main;main(5)"

https://pyinstaller.org/en/stable/common-issues-and-pitfalls.html#multi-processing

Fixes #873

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -3,6 +3,7 @@ import os
 import platform
 import tempfile
 import typing
+from multiprocessing import freeze_support
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import List, Optional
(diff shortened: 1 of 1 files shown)


apyrgio (approved):

apyrgio:
Great dig, feel free to merge!

PR #1108, opened 2025-03-20
CI: Enforce updating the CHANGELOG in the CI

almet:
Currently, this is only returning warnings, but we seem to just skip them. As it's possible to merge PRs when the CI is red, issuing an error would help us to think about populating this file.

### DIFF
--- a/.github/workflows/check_pr.yml
+++ b/.github/workflows/check_pr.yml
@@ -31,5 +31,6 @@ jobs:
         shell: bash
         run: |
           if git diff --exit-code "origin/${BASE_REF}" -- CHANGELOG.md; then
-              echo "::warning::No CHANGELOG.md modifications were found in this pull request."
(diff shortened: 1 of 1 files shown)


apyrgio (approved):

apyrgio:
Good idea, thanks!

almet:
Hmm, this is not doing what I want, it's marking a failure, but the CI is marked as green, still:

image

almet:
Okay, this looks better. Could you review that again @apyrgio? Thanks!

apyrgio:
Yeap, it seems to work. I wonder though, given that the PR now fails, should we add a changelog item to it?

_(I really like this change, thanks a lot for taking the initiative to do that!)_

almet:
> should we add a changelog item to it?

For this specific PR I don't think we should! Will merge!

PR #1110, opened 2025-03-24
Allow to specify the container_runtime in the settings

almet:
This allows to use Podman Desktop on macOS (or windows) as an alternative container runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 - Document Operating System support [#986](https://github.com/freedomofpress/dangerzone/issues/986)
 - Tests: Look for regressions when converting PDFs [#321](https://github.com/freedomofpress/dangerzone/issues/321)
 
+## Added
+
+- (experimental): It is now possible to specify a custom container runtime in
+  the settings, by using the `container_runtime` key. It should contain the path
+  to the container runtime you want to use. Please note that this doesn't mean
+  we support more container runtimes than Podman and Docker for the time being,
+  but enables you to chose which one you want to use, independently of your
+  platform. ([#925](https://github.com/freedomofpress/dangerzone/issues/925))
+
 ## [0.8.1](https://github.com/freedomofpress/dangerzone/compare/v0.8.1...0.8.0)
 
 - Update the container image
--- a/dangerzone/container_utils.py
+++ b/dangerzone/container_utils.py
@@ -3,7 +3,8 @@ import os
 import platform
 import shutil
 import subprocess
-from typing import List, Tuple
+from pathlib import Path
+from typing import List, Optional, Tuple
 
 from . import errors
 from .settings import Settings
@@ -14,29 +15,26 @@ CONTAINER_NAME = "dangerzone.rocks/dangerzone"
 log = logging.getLogger(__name__)
 
 
-def get_runtime_name() -> str:
-    settings = Settings()
-    try:
-        runtime_name = settings.get("container_runtime")
-    except KeyError:
-        return "podman" if platform.system() == "Linux" else "docker"
-    return runtime_name
-
-
-def get_runtime() -> str:
-    container_tech = get_runtime_name()
-    runtime = shutil.which(container_tech)
-    if runtime is None:
-        # Fallback to the container runtime path from the settings
+class Runtime(object):
+    def __init__(self) -> None:
         settings = Settings()
-        runtime_path = settings.get("container_runtime_path")
-        if os.path.exists(runtime_path):
-            return runtime_path
-        raise errors.NoContainerTechException(container_tech)
-    return runtime
+
+        if settings.custom_runtime_specified():
+            self.path = Path(settings.get("container_runtime"))
+            self.name = self.path.stem
+        else:
+            self.name = self.get_default_runtime_name()
+            binary_path = shutil.which(self.name)
+            if binary_path is None or not os.path.exists(binary_path):
+                raise errors.NoContainerTechException(self.name)
+            self.path = Path(binary_path)
+
+    @staticmethod
+    def get_default_runtime_name() -> str:
+        return "podman" if platform.system() == "Linux" else "docker"
 
 
-def get_runtime_version() -> Tuple[int, int]:
+def get_runtime_version(runtime: Optional[Runtime] = None) -> Tuple[int, int]:
     """Get the major/minor parts of the Docker/Podman version.
 
     Some of the operations we perform in this module rely on some Podman features
@@ -45,15 +43,15 @@ def get_runtime_version() -> Tuple[int, int]:
     just knowing the major and minor version, since writing/installing a full-blown
     semver parser is an overkill.
     """
-    # Get the Docker/Podman version, using a Go template.
-    runtime_name = get_runtime_name()
+    runtime = runtime or Runtime()
 
-    if runtime_name == "podman":
+    # Get the Docker/Podman version, using a Go template.
+    if runtime.name == "podman":
         query = "{{.Client.Version}}"
     else:
         query = "{{.Server.Version}}"
 
-    cmd = [get_runtime(), "version", "-f", query]
+    cmd = [str(runtime.path), "version", "-f", query]
     try:
         version = subprocess.run(
             cmd,
@@ -72,7 +70,7 @@ def get_runtime_version() -> Tuple[int, int]:
         return (int(major), int(minor))
     except Exception as e:
         msg = (
-            f"Could not parse the version of the {runtime_name.capitalize()} tool"
+            f"Could not parse the version of the {runtime.name.capitalize()} tool"
             f" (found: '{version}') due to the following error: {e}"
         )
         raise RuntimeError(msg)
@@ -85,10 +83,11 @@ def list_image_tags() -> List[str]:
     images. This can be useful when we want to find which are the local image tags,
     and which image ID does the "latest" tag point to.
     """
+    runtime = Runtime()
     return (
         subprocess.check_output(
             [
-                get_runtime(),
+                str(runtime.path),
                 "image",
                 "list",
                 "--format",
@@ -105,19 +104,21 @@ def list_image_tags() -> List[str]:
 
 def add_image_tag(image_id: str, new_tag: str) -> None:
     """Add a tag to the Dangerzone image."""
+    runtime = Runtime()
     log.debug(f"Adding tag '{new_tag}' to image '{image_id}'")
     subprocess.check_output(
-        [get_runtime(), "tag", image_id, new_tag],
+        [str(runtime.path), "tag", image_id, new_tag],
         startupinfo=get_subprocess_startupinfo(),
     )
 
 
 def delete_image_tag(tag: str) -> None:
     """Delete a Dangerzone image tag."""
+    runtime = Runtime()
     log.warning(f"Deleting old container image: {tag}")
     try:
         subprocess.check_output(
-            [get_runtime(), "rmi", "--force", tag],
+            [str(runtime.name), "rmi", "--force", tag],
             startupinfo=get_subprocess_startupinfo(),
         )
     except Exception as e:
@@ -134,11 +135,12 @@ def get_expected_tag() -> str:
 
 
 def load_image_tarball() -> None:
+    runtime = Runtime()
     log.info("Installing Dangerzone container image...")
     tarball_path = get_resource_path("container.tar")
     try:
         res = subprocess.run(
-            [get_runtime(), "load", "-i", str(tarball_path)],
+            [str(runtime.path), "load", "-i", str(tarball_path)],
             startupinfo=get_subprocess_startupinfo(),
             capture_output=True,
             check=True,
@@ -163,7 +165,7 @@ def load_image_tarball() -> None:
     # `share/image-id.txt` and delete the incorrect tag.
     #
     # [1] https://github.com/containers/podman/issues/16490
-    if get_runtime_name() == "podman" and get_runtime_version() == (3, 4):
+    if runtime.name == "podman" and get_runtime_version(runtime) == (3, 4):
         expected_tag = get_expected_tag()
         bad_tag = f"localhost/{expected_tag}:latest"
         good_tag = f"{CONTAINER_NAME}:{expected_tag}"
--- a/dangerzone/container_utils.py
+++ b/dangerzone/container_utils.py
@@ -21,6 +21,8 @@ class Runtime(object):
 
         if settings.custom_runtime_specified():
             self.path = Path(settings.get("container_runtime"))
+            if not self.path.exists():
+                raise errors.UnsupportedContainerRuntime(self.path)
             self.name = self.path.stem
         else:
             self.name = self.get_default_runtime_name()
@@ -28,6 +30,9 @@ class Runtime(object):
             if binary_path is None or not os.path.exists(binary_path):
                 raise errors.NoContainerTechException(self.name)
             self.path = Path(binary_path)
+
+        if self.name not in ("podman", "docker"):
+            raise errors.UnsupportedContainerRuntime(self.name)
 
     @staticmethod
     def get_default_runtime_name() -> str:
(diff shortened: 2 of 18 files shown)


dangerzone/gui/main_window.py
```diff
@@ -1222,7 +1233,7 @@ def start_conversion(self) -> None:
         if not self.thread_pool_initized:
             max_jobs = self.dangerzone.isolation_provider.get_max_parallel_conversions()
             # Call freeze_support() to avoid passing unknown options to the subprocess.
-            # See https://github.com/freedomofpress/dangerzone/issues/873
+            # See https://github.com/freedomofpress/Adangerzone/issues/873
```
almet, line 1236: FIXME

dangerzone/container_utils.py
```diff
@@ -1,27 +1,40 @@
...
-        # Windows, Darwin, and unknown use docker for now, dangerzone-vm eventually
-        runtime_name = "docker"
-    return runtime_name
+class Runtime(object):
+    def __init__(self) -> None:
+        settings = Settings()
+
+        if settings.custom_runtime_specified():
+            self.path = Path(settings.get("container_runtime"))
+            self.name = self.path.stem
```
apyrgio, line 24: My initial reaction was "why use our settings code for this, if we can just rely on envvars?". Environment variables seem to work nice for GUI apps on Windows, but not on macOS, which I think is why you went down the settings route, right?

If that's the case, I still think it would be helpful to add an `if` statement before the settings one, which checks if a custom runtime has been specified as an envvar (e.g., `DANGERZONE_CONTAINER_RUNTIME`). I think it will help when we want to debug things ourselves, or when we want to help a user to switch to a different container runtime. Else, it's probably hard for non-devs to use this functionality.

What are your thoughts on this?

dangerzone/isolation_provider/container.py
```diff
@@ -123,27 +125,30 @@ def should_wait_install() -> bool:
...
 
     def check_docker_desktop_version(self) -> Tuple[bool, str]:
         # On windows and darwin, check that the minimum version is met
         version = ""
-        if platform.system() != "Linux":
+        runtime = Runtime()
+        runtime_is_docker = runtime.name == "docker"
+        platform_is_not_linux = platform.system() != "Linux"
+
+        if runtime_is_docker and platform_is_not_linux:
```
apyrgio, line 151: I think `if runtime.name == "docker" and platform.system() != "Linux":` is already clear enough, so I wonder why the extra variables were necessary :thinking:

tests/test_container_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    settings = Settings()
+    settings.set(
+        "container_runtime", "/opt/somewhere/new-kid-on-the-block", autosave=True
+    )
+
+    assert Runtime().name == "new-kid-on-the-block"
+
+
+def test_get_runtime_name_linux(mocker: MockerFixture, tmp_path: Path) -> None:
```
apyrgio, line 20: Why not make this a test that runs only on Linux platforms, and thus remove the need for mocking? Similarly for the non-Linux test below.

tests/test_container_utils.py
apyrgio, line 1: A test that I feel is missing is what if a runtime does not have the name "docker" or "podman", and what if it doesn't exist. We can live without it, but I wanted to know if you feel the same.

CHANGELOG.md
```diff
@@ -16,6 +16,15 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
...
 - Tests: Look for regressions when converting PDFs [#321](https://github.com/freedomofpress/dangerzone/issues/321)
 
+## Added
+
+- It is now possible to specify a custom container runtime in the settings, by
+  using the `container_runtime` key. It should contain the path to the container
+  runtime you want to use. Please note that this doesn't mean we support more
+  container runtimes than Podman and Docker for the time being, but enables you to
+  chose which one you want to use, independently of your platform.
+  ([#925](https://github.com/freedomofpress/dangerzone/issues/925))
```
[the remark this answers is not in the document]
almet: Instead, I would like to put this inside a "settings" section in the documentation, once #1113 is done. I can add it to the FAQ for the time being, or... actually I think it's good if it's undocumented for now, as we don't want to really rely on it.

Would you be okay with no docs for now, and then add documentation? If not, I can add a section in the FAQ.

dangerzone/container_utils.py
almet: I actually haven't thought about using envvars at first, to be honest!

It seems like a food option for the CLI, but I don't think it is for the GUI. Using environment variables would require the users to do some changes to how they run the application, which seem less trivial than editing the JSON file by hand (which is also a bit involved, but less, to my eyes).

CHANGELOG.md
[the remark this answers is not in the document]
apyrgio: I'm cool with that, we can treat it as an escape hatch and add docs later on. They will probably not be tagged after all. I'd still recommend adding somewhere that this feature is experimental, or something to that effect.

tests/test_container_utils.py
almet: Because it makes it possible to catch the errors on my dev machine :-)

dangerzone/container_utils.py
apyrgio: Note that for Windows, you don't need to change the desktop entry or anything. It's super simple to add envvars globally. But for Linux and (more importantly) macOS, there's no elegant way to set up global environment variables. So yeah, we can't rely on just that.

I guess the question becomes: do we want to add an envvar for the users who will invoke Dangerzone from the CLI (either our GUI or CLI version) before 0.9.0 is out, or not? If we don't, how should we instruct users to switch container runtimes, in case of a suspected Docker regression?

dangerzone/isolation_provider/container.py
apyrgio: Nah nah, it's fine. I was just wondering if they were stale or something. Resolving this comment...

CHANGELOG.md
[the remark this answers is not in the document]
apyrgio: I dig it, thanks.

tests/test_container_utils.py
apyrgio: Heh, I assumed as much. This makes the test cases a bit more synthetic, but I do get that we'll catch errors more easily, so it's a pass from me. Resolving this thread...

dangerzone/container_utils.py
almet: I added the ability to pass a flag to set the container runtime, with `dangerzone-cli --set-container-runtime podman` for instance.

You can see that in 73d7a466906bb87e8d5e90dac7d3d83c9fadf67e

tests/test_container_utils.py
almet: At first I thought it would be a good thing to allow people to use a different container runtime, but I start to think this might get us in awkward situations, because the docker/podman flags we add when running the container are expected to be either one of podman or docker.

Are you okay if I error out in these cases?

apyrgio: For now, I think that's best, yes.

almet: Okay. I added some checks and tests about that in  2a29cf7c27b53df43437e8cb6e88b31ad2675fad

dangerzone/container_utils.py
apyrgio: Looks good, but I'd also add this:

```diff
diff --git a/dangerzone/cli.py b/dangerzone/cli.py
index c5ba9d8d..9c25cd8a 100644
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -38,7 +38,7 @@ def print_header(s: str) -> None:
 )
 @click.argument(
     "filenames",
-    required=True,
+    required=False,
     nargs=-1,
     type=click.UNPROCESSED,
     callback=args.validate_input_filenames,
@@ -59,7 +59,7 @@ def print_header(s: str) -> None:
 def cli_main(
     output_filename: Optional[str],
     ocr_lang: Optional[str],
-    filenames: List[str],
+    filenames: Optional[List[str]],
     archive: bool,
     dummy_conversion: bool,
     debug: bool,
@@ -71,6 +71,9 @@ def cli_main(
         settings = Settings()
         settings.set("container_runtime", set_container_runtime, autosave=True)
         click.echo(f"Set the settings container_runtime to {set_container_runtime}")
+        return 0
+    elif not filenames:
+        raise click.UsageError("Missing argument 'FILENAMES...'.")
 
     if getattr(sys, "dangerzone_dev", False) and dummy_conversion:
         dangerzone = DangerzoneCore(Dummy())
```

tests/test_container_utils.py
apyrgio: Thanks for that. When running the code, I now see the following error:

```python
  File "/home/user/dangerzone/./dev_scripts/dangerzone", line 13, in <module>
    dangerzone.main()                                                                                                  
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)                                                                                  
           ^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                  
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1082, in main                                                                                                     
    rv = self.invoke(ctx)                                                                                              
         ^^^^^^^^^^^^^^^^                                                                                                                                                                                                                     
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1443, in invoke                                                                                                   
    return ctx.invoke(self.callback, **ctx.params)                                                                                                                                                                                            
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                            
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 788, in invoke                                                                                                    
    return __callback(*args, **kwargs)                                                                                                                                                                                                        
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                        
  File "/home/user/dangerzone/dangerzone/errors.py", line 110, in wrapper                                                                                                                                                                     
    return func(*args, **kwargs)                                                                                                                                                                                                              
           ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                              
  File "/home/user/dangerzone/dangerzone/gui/__init__.py", line 156, in gui_main                                                                                                                                                              
    window = MainWindow(dangerzone)                                                                                                                                                                                                           
             ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                           
  File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 193, in __init__                                                                                                                                                           
    self.waiting_widget: WaitingWidget = WaitingWidgetContainer(self.dangerzone)                                                                                                                                                              
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                              
  File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 528, in __init__                                                                                                                                                           
    self.check_state()                                                                                                                                                                                                                        
  File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 551, in check_state                                                                                                                                                        
    self.state_change(state, error)                                                                                    
  File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 615, in state_change                                                                                                                                                       
    self.show_error(message, error)                      
                    ^^^^^^^                                                                                            
UnboundLocalError: cannot access local variable 'message' where it is not associated with a value
```

which I think can be fixed with:

```diff
diff --git a/dangerzone/gui/main_window.py b/dangerzone/gui/main_window.py
index d162fcf8..f107ef32 100644
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -602,17 +602,16 @@ class WaitingWidgetContainer(WaitingWidget):
                 )
             elif platform.system() == "Linux":
                 # "not_running" here means that the `podman image ls` command failed.
-                message = (
+                self.show_error(
                     "<strong>Dangerzone requires Podman</strong><br><br>"
                     "Podman is installed but cannot run properly. See errors below"
                 )
             else:
-                message = (
+                self.show_error(
                     "<strong>Dangerzone requires Docker Desktop</strong><br><br>"
                     "Docker is installed but isn't running.<br><br>"
                     "Open Docker and make sure it's running in the background."
                 )
-            self.show_error(message, error)
         else:
             self.show_message(
                 "Installing the Dangerzone container image.<br><br>"
```

But even with this fix, I see the following with a bad container name/path:

```python
  File "/home/user/dangerzone/./dev_scripts/dangerzone", line 13, in <module>                                                                                                                                                                 
    dangerzone.main()       
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1161, in __call__                                                                                                 
    return self.main(*args, **kwargs)                                                                                                                                                                                                         
           ^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                         
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1082, in main                                                                                                     
    rv = self.invoke(ctx)                                                                                                                                                                                                                     
         ^^^^^^^^^^^^^^^^                                                                                                                                                                                                                     
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 1443, in invoke                                                                                                   
    return ctx.invoke(self.callback, **ctx.params)                                                                                                                                                                                            
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                            
  File "/home/user/.cache/pypoetry/virtualenvs/dangerzone-hQU0mwlP-py3.11/lib/python3.11/site-packages/click/core.py", line 788, in invoke                                                                                                    
    return __callback(*args, **kwargs)                                                                                                                                                                                                        
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                        
  File "/home/user/dangerzone/dangerzone/errors.py", line 110, in wrapper                                                                                                                                                                     
    return func(*args, **kwargs)                                                                                                                                                                                                              
           ^^^^^^^^^^^^^^^^^^^^^                     
  File "/home/user/dangerzone/dangerzone/gui/__init__.py", line 156, in gui_main
    window = MainWindow(dangerzone)                      
             ^^^^^^^^^^^^^^^^^^^^^^                                                                                    
  File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 225, in __init__
    self.dangerzone.isolation_provider.check_docker_desktop_version()
  File "/home/user/dangerzone/dangerzone/isolation_provider/container.py", line 147, in check_docker_desktop_version
    runtime = Runtime()       
              ^^^^^^^^^                    
  File "/home/user/dangerzone/dangerzone/container_utils.py", line 25, in __init__
    raise errors.UnsupportedContainerRuntime(self.path)                                                                
dangerzone.errors.UnsupportedContainerRuntime: /usr/asdsadfdsaf/podman  
```

which I think should be handled and shown in the UI.

almet: Thanks, I fixed that in e7cdc15d4f7f4a49dad001d56bb4ab546ee336a2, and a nice error is shown now.

almet: I also changed the way setting this `container_runtime` value is done, by actually getting the real path and putting this in the settings, in 7399b316bbf34831c2d777733ca76c2fadd38ecd.

If setting to an unknown value, then an error will be raised:

```
dangerzone.errors.NoContainerTechException: /usr/yay/podman is not installed
```

dangerzone/container_utils.py
almet: Applied your suggestion as part of https://github.com/freedomofpress/dangerzone/commit/7399b316bbf34831c2d777733ca76c2fadd38ecd, thanks!

tests/test_container_utils.py
apyrgio: Looks great!

apyrgio:
Just did a recheck locally, and I don't have anything else to comment on. Feel free to merge!

PR #1113, opened 2025-03-25
Use Sphinx to revamp the documentation

almet:
This is an attempt to use sphinx for our documentation, with the idea of pushing it to docs.dangerzone.rocks, potentially managed by readthedocs.


almet:
I'm closing this PR for now. It can still serve as a reference point for future sphinx implementation, but the documentation content is out of date. Updating this PR will be more work than starting afresh, so I think we'll go for this second option.

PR #1115, opened 2025-03-27
Provide a way to opt-out from CHANGELOG check

almet:


apyrgio (approved):

apyrgio:
I've updated the PR by using a GHA job that Alexis mentioned the other day (https://github.com/tarides/changelog-check-action). I've also pinned it to a specific Git commit, to avoid any nasty surprises down the road.

Alexis, feel free to re-sign the commit and merge it.

PR #1116, opened 2025-03-30
Use an image digest to enforce container image determinism

sudoforge:
66600f32dc7c4f45638dbcd7a450bb5707f9f9cd introduced various improvements
to the determinism of the container image in this repository. This
change builds on to this effort by introducing support for a container
image digest, without removing the human-friendly tag (date) reference.


Dockerfile
```diff
@@ -185,8 +186,8 @@ RUN mkdir -p \
 # Copy the /etc and /var directories under the new root directory. Also,
 # copy /etc/, /opt, and /usr to the Dangerzone image rootfs.
 #
-# NOTE: We also have to remove the resolv.conf file, in order to not leak any DNS
-# servers added there during image build time.
+# NOTE: We also have to remove the resolv.conf file, in order to not leak any
+# DNS servers added there during image build time.
```
sudoforge, line 190: autoformatted by my editor. i'll revert this.

Dockerfile
```diff
@@ -3,8 +3,9 @@
 # docs/developer/reproducibility.md.
 
 ARG DEBIAN_IMAGE_DATE=20250224
+ARG DEBIAN_IMAGE_DIGEST=sha256:12c396bd585df7ec21d5679bb6a83d4878bc4415ce926c9e5ea6426d23c60bdc
 
-FROM debian:bookworm-${DEBIAN_IMAGE_DATE}-slim AS dangerzone-image
+FROM debian:bookworm-${DEBIAN_IMAGE_DATE}-slim@${DEBIAN_IMAGE_DIGEST} AS dangerzone-image
```
sudoforge, line 8: We could also add the registry hostname here, as in `docker.io/library/debian:bookworm-${DEBIAN_IMAGE_DATE}-slim@${DEBIAN_IMAGE_DIGEST}`, if we wanted to remove the choice from the user (building this image with `podman`, for example, prompts the user to choose between `docker.io/library/...` or `quay.io/...`).

Users can set a preference for this in `registries.conf` (`man 5 containers-registries.conf`), but it may be preferable for this repository to specify an upstream (users can always remap the registry to their preferred one anyway).

`docker` (used in this repository's pipelines) will default to `docker.io/library` and prefix the image URI with this, so the workflow files and any docker users will not encounter the prompt.

Dockerfile.env
```diff
@@ -1,5 +1,7 @@
 # Can be bumped to the latest date in https://hub.docker.com/_/debian/tags?name=bookworm-
 DEBIAN_IMAGE_DATE=20250224
+# Should be the INDEX DIGEST for the tag with the selected build date
+DEBIAN_IMAGE_DIGEST=sha256:12c396bd585df7ec21d5679bb6a83d4878bc4415ce926c9e5ea6426d23c60bdc
```
sudoforge, line 4: i don't personally see any value in keeping `DEBIAN_IMAGE_DATE` around; no OCI build tool will use the tag in any way (it could literally be `RANDOMTEXT` and would not impact the build at all, because only the digest is used).

the purpose it _does_ serve is as a human-friendly reference to the layer, but this is easily accomplished by something like a comment:

```suggestion
# Should be the INDEX DIGEST from an image tagged `bookworm-<DATE>-slim`
# You can find images at https://hub.docker.com/_/debian/tags?name=bookworm-
# Clicking on the image will take you to the detail page. At the top of this page is the index digest. Copy this value and replace it in the variable below.
#
# DATE: 20250224
DEBIAN_IMAGE_DIGEST=sha256:12c396bd585df7ec21d5679bb6a83d4878bc4415ce926c9e5ea6426d23c60bdc
```

docs/developer/reproducibility.md
```diff
@@ -28,6 +28,7 @@ This means that rebuilding the image without updating our Dockerfile will
 Here are the necessary variables that make up our image in the `Dockerfile.env`
 file:
 * `DEBIAN_IMAGE_DATE`: The date that the Debian container image was released
+* `DEBIAN_IMAGE_DIGEST`: The date that the Debian container image was released
```
apyrgio, line 31: I think you meant to write something like this, right?

```suggestion
* `DEBIAN_IMAGE_DIGEST`: The SHA-256 checksum of the Debian container image
```

Dockerfile.env
apyrgio: Yeap, I agree. If we use image digests, then the tags are not consulted by the container runtime. Your suggestion looks great.

apyrgio:
Thanks a lot for this very thoughtful addition @sudoforge. Let me know if you want me to make the changes, or if you can do them. In any case, it looks good to go.

docs/developer/reproducibility.md
sudoforge: heh, yep! good ol' copypasta. this is resolved in the latest push.

sudoforge:
@apyrgio i struggle with submitting changes that break my internal rule of "1 change, 1 commit". i'm going to pull the addition of the fully qualified hostname out of this change. i agree that it's useful, so i'll submit another pr after this to add that, so that it can be reviewed/tracked independently.

apyrgio:
Sure thing :slightly_smiling_face: . Merging this PR right now, thanks again for the contribution!

PR #1126, opened 2025-04-07
Fixes from the 0.9.0 QA

apyrgio:
This PR tracks some fixes for some small bugs/issues we encountered during our 0.9.0 QA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,22 +54,21 @@ Dockerfile: Dockerfile.env Dockerfile.in
 poetry-install:
 	poetry install
 
-
 .PHONY: build-clean
 build-clean:
-	doit clean
+	poetry run doit clean
 
 .PHONY: build-macos-intel
 build-macos-intel: build-clean poetry-install
-	doit -n 8
+	poetry run doit -n 8
 
 .PHONY: build-macos-arm
 build-macos-arm: build-clean poetry-install
-	doit -n 8 macos_build_dmg
+	poetry run doit -n 8 macos_build_dmg
 
 .PHONY: build-linux
 build-linux: build-clean poetry-install
-	doit -n 8 fedora_rpm debian_deb
+	poetry run doit -n 8 fedora_rpm debian_deb
 
 .PHONY: regenerate-reference-pdfs
 regenerate-reference-pdfs:
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ fix: ## apply all the suggestions from ruff
 	ruff format
 
 .PHONY: test
-test:
+test: ## Run the tests
 	# Make each GUI test run as a separate process, to avoid segfaults due to
 	# shared state.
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -50,21 +50,25 @@ test-large: test-large-init  ## Run large test set
 Dockerfile: Dockerfile.env Dockerfile.in
 	poetry run jinja2 Dockerfile.in Dockerfile.env > Dockerfile
 
+.PHONY: poetry-install
```
almet, line 53: Just a note here (so auto-closing the comment as it's not something to solve): I believe this should be what we do in this makefile: run `make target` and have the makefile call `poetry` for us, rather than doing `poetry run make target`.

dangerzone/gui/main_window.py
```diff
@@ -229,6 +228,8 @@ def __init__(self, dangerzone: DangerzoneGui) -> None:
                     self.handle_docker_desktop_version_check(is_version_valid, version)
             except errors.UnsupportedContainerRuntime as e:
                 pass  # It's catched later in the flow.
+            except errors.NoContainerTechException as e:
+                pass  # It's catched later in the flow.
```
apyrgio, line 232: I have an `IMPROVE_ME` note in this commit, and I was wondering the following. Wouldn't it be better if we added the Docker Desktop check in this part of the code?

https://github.com/freedomofpress/dangerzone/blob/b5c09e51d8fbc3f74e5af78193f51b055ab7ebc4/dangerzone/gui/main_window.py#L620-L627

I'm suggesting this part because at this point we know that the container runtime is installed, and is up and running. This way, we won't have to swallow these exceptions. What do you think?

apyrgio:
I've updated the PR to tidy up some fixups, fix some typos here and there, and point to the the cx_Freeze issue (#1128). Also, I took a look at your changes, and they look good to me, thanks!

As for the changelog, I don't see something particular that can be added there, so I'd propose adding the `no changelog` label.

almet (approved):

almet (approved):

dangerzone/gui/main_window.py
apyrgio: I've opened a separate issue for this (https://github.com/freedomofpress/dangerzone/issues/1132), and I'll refer to it from this commit in the code. In the meantime, we'll merge this PR to not block the 0.9.0 release.

dangerzone/cli.py
```diff
@@ -52,7 +52,11 @@ def print_header(s: str) -> None:
 @click.option(
     "--set-container-runtime",
     required=False,
-    help="The path to the container runtime you want to set in the settings",
+    help=(
+        "The name or full path of the container runtime you want Dangerzone to use."
```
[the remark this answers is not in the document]
apyrgio: Hm, I'm not too hang up on a particular style, but I think it's best to let it pass for now. I'd be interested to check how linters break long lines, and where they put the extra space.

Issue #1127, opened 2025-04-07
Podman flags make conversions fail on Windows platforms


apyrgio:
While testing Podman Desktop support on Windows platforms, we found out that the following flags that Dangerzone passes may fail document conversions:

#### `--userns nomap`

This flag works only if Podman has been set in rootless mode. Here's how you can do this on Windows: 

https://github.com/containers/podman/blob/main/docs/tutorials/podman-for-windows.md#rootful--rootless

From personal experience, it seems that users who set up the Podman Desktop machine via CLI end up with a **rootless** Podman installation, by default. However, users who set up Podman Desktop machinge via the GUI end up with a **rootful** Podman installation, by default.

In the latter case, conversions fail with message `Unknown error code '125'`. The underlying Podman error though is `Error: nomap is only supported in rootless mode`.

CLI output:

```
[INFO ] > 'C:\Program Files\RedHat\Podman\podman.EXE' run --log-driver none --security-opt no-new-privileges --userns nomap --cap-drop all --cap-add SYS_CHROOT --security-opt label=type:container_engine_t --network=none -u dangerzone --rm -i --name dangerzone-doc-to-pixels-Ox6ydU dangerzone.rocks/dangerzone:20250331-0.9.0-rc1-13-g067c741-a2ec /usr/bin/python3 -m dangerzone.conversion.doc_to_pixels
                                                                                
[INFO ] Conversion output (doc to pixels)
----- DOC TO PIXELS LOG START -----
Error: nomap is only supported in rootless mode
----- DOC TO PIXELS LOG END -----
[ERROR] [doc Ox6ydU] 0% Unknown error code '125'
[DEBUG] Marking doc Ox6ydU as 'failed'
Failed to convert document(s)
C:\Users\dz\dangerzone\tests\test_docs\sample-pdf.pdf
```

#### `--security-opt seccomp=C:/Program Files/Dangerzone/share/seccomp.gvisor.json`

Dangerzone passes a platform-agnostic flag to the underlying container runtime that makes it use a particular seccomp policy. For Podman Desktop in Windows though, passing this flag makes conversions fail with message `Unknown error code '125'`. The underlying Podman error though is `Error: opening seccomp profile failed: open C:\Users\dz\dangerzone\share\seccomp.gvisor.json: no such file or directory`. Interestingly, this file exists locally, so this looks like an issue we need to open upstream.

CLI output:

```
[...]

PR #1129, opened 2025-04-07
container: Add workarounds for Podman Desktop support on Windows

apyrgio:
In case we run on Windows and use Podman Desktop (for which we currently offer experimental support), we must not pass some Podman flags in order to avoid conversion errors.

Refs #1127
Based on #1126

### DIFF
--- a/dangerzone/isolation_provider/container.py
+++ b/dangerzone/isolation_provider/container.py
@@ -56,7 +56,14 @@ class Container(IsolationProvider):
             security_args = ["--log-driver", "none"]
             security_args += ["--security-opt", "no-new-privileges"]
             if container_utils.get_runtime_version() >= (4, 1):
-                security_args += ["--userns", "nomap"]
+                # We perform a platform check to avoid the following Podman Desktop
+                # error on Windows:
+                #
+                #   Error: nomap is only supported in rootless mode
+                #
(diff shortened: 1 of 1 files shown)


almet:
(I changed the base of this PR to `qa-0.9.0-fixes`  to better see what this adds)

Thanks for proposing these changes. With them, we can see how it would offer different security guarantees than on other platforms.

Here are the questions that come to mind (sorry it's a bit defensive):

- Should we allow / encourage the use of non rootless podman ? If so, what are the added risks when doing so? Because we're on windows, it means that the Linux kernel is running in a VM, I'm not clear as to what
- Same question about the `seccomp` policy: it mean we would be defaulting to the one shipped with Docker Desktop, is it different from the one we currently ship, and how? Is it acceptable to rely on this seccomp policy?

If any of the answers to these questions are true, we might not want to allow running with Podman Desktop on Windows for now, even in an experimental way.

almet:
After a discussion together, I'm happy with these changes. Currently, windows user use Docker Desktop, which doesn't have a rootless mode anyway, so we'll be on par with them here, and we don't require people to have Podman rootless on their machine.

In the future, we could detect that they're running podman rootfull and ask them if they're okay for us to configure their system for them (but that's for later)

PR #1135, opened 2025-04-09
Docs: Update the Podman Desktop docs for macOS

almet:
In order to access our custom seccomp policy, we require it to be mounted on the podman machine.

An alternative way to solve this would be to not pass a specific seccomp policy for Podman on macOS, as we did for windows:

https://github.com/freedomofpress/dangerzone/blob/c99c424f87bf35085fa39b7b36d5f94dc047c201/dangerzone/isolation_provider/container.py#L84-L86

Since that would require us to rebuild the artifacts, and that support is optional, I believe the extra steps documented here might work, but happy to do it the other way if anybody prefers.


docs/podman-desktop.md
```diff
@@ -12,7 +12,16 @@ You will be required to open a terminal and follow these steps:
...
+You will need to configure podman to access the shared Dangerzone resources:
+
+```bash
+podman machine stop
+podman machine rm
+podman machine init -v /Applications/Dangerzone.app:/Applications/Dangerzone.app
+podman machine set --rootful=false  
+podman machine start
+```
+Then, set the container runtime to podman, use this command:
```
apyrgio, line 24:
```suggestion
Then, set the container runtime to podman, using this command:
```

apyrgio:
The workaround looks good to me. Ultimately though, if we had a bit more time, I would suggest the following:
1. In case of Podman Desktop on Windows/macOS, copy the seccomp profile in the cache dir of Dangerzone (under `/Users` in macOS), and spawn a container by pointing to the copied seccomp profile. I think this way we can bypass this limitation.
2. Remove `init -v` from our instructions and make them common for Windows and macOS users. After all, we prefer rootless installations in any case.

We can do this after 0.9.0 though.

docs/podman-desktop.md
almet: Thanks, updated and rebased.

PR #1138, opened 2025-04-09
Include Ubuntu Plucky and Fedora 42 in our nightly repo checks

apyrgio:


almet (approved):
Weird this is put in there in reverse order but yeah, that looks good to me. Feel free to merge!

PR #1144, opened 2025-04-28
Ignore CVE-2025-43859 / GHSA-vqfr-h8mv-ghfj

apyrgio:
Ignore an h11 vulnerability that is present in the Dangerzone application released from the `v0.9.0` tag. This vulnerability
reportedly affects web servers behind reverse proxies, which is not Dangerzone's case.

### DIFF
--- a/.github/workflows/scan_released.yml
+++ b/.github/workflows/scan_released.yml
@@ -67,8 +67,12 @@ jobs:
           fetch-depth: 0
       - name: Checkout the latest released tag
         run: |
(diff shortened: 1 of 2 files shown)


almet (approved):
Hey thanks. I've also reviewed the CVE and it looks harmless for the reasons you mentioned (we're not a server, we're not using a reverse proxy).

Thanks for making it possible to use the grype list from the main branch when doing the security scans.

PR #1149, opened 2025-05-07
Introduce inventory tool to manage GitHub assets

### LINKED ISSUE #1146: Vendor Dangerzone assets in a uniform way
Dangerzone currently vendors the following assets:
* The Tesseract data (see `./install/common/download-tessdata.py`)
* The H2ORestart plugin, as part of our Dockerfile
* The PyMuPDF wheel (for Debian distros, see `./install/linux/debian-vendor-pymupdf.py`)

We will soon start to have a few more vendored assets as part of the independent container updates effort (#1080) and the embedding of Podman (#1145).  In order to uniformly keep track of them, download and update them, we need the following:

First, we need an inventory that tracks the following:

- Asset description
- Asset remote URL
- Asset name (platform-dependent)
- Asset checksum
- Asset version
- Asset local path

> [!NOTE]
> Most of our assets are actually provided by GitHub releases of other projects, so the logic should be easy to generalize.

Ideally, this inventory should be stored in our repo in a format like TOML/INI/JSON or similar, like our `poetry.lock` file.

Then, we need a script that will do the following:
* Download an asset or all assets for a platform based on their URL and the local path
* Check if there are new versions for the downloaded assets
* Get the new version of an asset, calculate its checksum, and update the inventory in place

Finally, once we have the above, we should add:

- A check in our release instructions to verify that the assets are in their latest version, else update them.
- A third-party notice about the assets we include in our Windows/macOS and their licenses
- A step in our build/release instructions to download all the assets. This should replace the step that currently downloads the Tesseract data.

The prerequisites are simple enough for us to write a helper script that does the above, but ideally we should find a project that does it for us.

apyrgio:
Introduce a tool for managing GitHub assets, called `inventory.py`. This is a Python script designed to manage assets from GitHub releases. It expects a configuration file (in TOML format) that contains a list of assets and some parameters. Using this config file, it can query GitHub for release information, compute checksums for assets, update a lock file (JSON format) and sync assets as described in the lock file.

In this PR, we use the inventory script to replace the `download-tessdata.py` script. We've been using the latter for a while to download Tesseract data, but we now have a more formalized way to handle this asset.

Fixes #1146


dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+            "checksum": checksum,
+            "executable": executable,
+            "destination": destination_str,
+            "extract": extract,
+        }
+
+    return asset_lock_data
+
+
+def sync_asset(asset_name, target_plat, asset):
```
almet, line 559: Proposal to use a different naming, like this:
```suggestion
def sync_asset(name, platform, asset_dict):
```

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+    if target_plat not in asset:
+        if "all" in asset:
+            target_plat = "all"
+        else:
+            raise InvException(
+                f"No entry for platform '{target_plat}' or 'platform.all'"
+            )
+
+    info = asset[target_plat]
+    download_url = info["download_url"]
```
almet, line 571: Nit / design: should we use a class / dataclass for the assets? It might be easier to read / avoid these calls in different locations.

docs/developer/inventory.md
```diff
@@ -0,0 +1,164 @@
...
+| `destination`         | yes | The local path where the asset should be saved after download. If the asset is a file, the destination will be its filename. Else, it will be the directory where the contents will be extracted in | Any valid file path string (e.g., `"downloads/asset.exe"`).                                                                                                                                                           |
+| `extract`             | no  | Instructions for file extraction from the downloaded asset.                                                                                                    | `false` (default) for no extraction; a list of glob strings to extract matching files; or a table with keys (see below). |
+| `extract.globs`       | no  | a list of glob strings to match specific files from an archive | Any valid glob such as `*.exe`, `bin/**/asset` ([options](https://docs.python.org/3/library/fnmatch.html)). Will extract all files in the archive if omitted.  |
+| `extract.flatten`     | no  | copy the files to the destination root | `true` or `false` (default) |
```
almet, line 69: Nitpicky, but tables don't really scale well for settings, in my experience. What about using a list instead? (I also find it easier to review, for little downsides (it looks a bit better in a table, but I don't think that's worth the "cost")

docs/developer/inventory.md
```diff
@@ -0,0 +1,164 @@
...
+Example:
+
+```
+./dev_scripts/inventory.py lock
+Processing 'asset1'
+Processing 'asset2'
+Lock file 'inventory.lock' updated.
+```
+
+#### sync
```
almet, line 98: I'm wondering if this `sync` should be actually called `download` instead. I know this is coming from the analogy with `poetry`, but it just seems better to my eyes to have this named differently, as `sync` doesn't really convey it's meaning to non poetry-aware readers. Curious about what you think?

docs/developer/inventory.md
```diff
@@ -0,0 +1,164 @@
...
+- `-p, --platform`:
+  Specify the platform for which the assets should be processed. Examples
+  include: `windows/amd64`, `linux/amd64`, `darwin/amd64`, `darwin/arm64`.
+  If not provided, the current platform is auto-detected.
+
+- `-v, --verbose`:
+  Enable verbose logging. Use `-v` for INFO level or `-vv` (or more) for DEBUG
+  level messaging.
+
+- `-C, --directory`:
```
almet, line 162: I'm curious: why is this `-C` and not `-d`? Why in caps?

almet (commented):
Hey, thanks for this addition @apyrgio, it's amazing!

I've reviewed the code and added comments. Let me know what you think. If that's helpful, we can also have a video call about it. Let me know when I can have another look :-)

The dependencies to our code are so little that I think this could actually be packaged as a separate project instead of something that's inside Dangerzone itself.

What are your thoughts about that? I do think it could be useful to some other people :-)

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
apyrgio: Great idea! This name is not taken in PyPI as well :wink:

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+from pathlib import Path
+
+import requests
+import semver
+import toml
+from colorama import Back, Fore, Style
+from platformdirs import user_cache_dir
+
+# CONSTANTS
+CONFIG_FILE = Path("inventory.toml")
```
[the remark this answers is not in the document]
apyrgio: Good idea, but my intention is to point to the actual file. That is, the expected path is `$CWD/inventory.toml`. You can influence where this file is by changing the current working directory, using the `-C` / `--directory` argumenst.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+
+def read_config():
+    """
+    Read the config for the inventory tool, either from its own configuration file
+    (inventory.toml) or a [tool.inventory] section in pyproject.toml.
+    """
+    if CONFIG_FILE.exists():
+        # First, attempt to read the configuration from inventory.toml.
+        try:
+            return toml.load(CONFIG_FILE.open("r"))
```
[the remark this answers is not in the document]
apyrgio: Hm, I could use `toml.loads(CONFIG_FILE.read_text())` as well, to avoid this ambiguity. Thanks for pointing it out.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+        # Then, attempt to read the configuration from pyproject.toml.
+        try:
+            config = toml.load(PYPROJECT_FILE.open("r"))
+        except Exception as e:
+            msg = f"Could not load configuration file '{PYRPROJECT_FILE}': {e}"
+            raise InvException(msg) from e
+
+        # If the pyproject.toml file does not have a [tool.inventory] section, return an
+        # error.
+        config = config.get("tool", {}).get("inventory")
```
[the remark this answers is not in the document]
apyrgio: There goes my one-liner :stuck_out_tongue: . Ok, you're right, it's better this way.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+    with open(LOCK_FILE, "w") as fp:
+        json.dump(lock_data, fp, indent=2)
+
+
+def check_lock_stale(lock):
+    config = read_config()
+    config_hash = hashlib.sha256(json.dumps(config).encode()).hexdigest()
+    if config_hash != lock["config_checksum"]:
+        raise InvException(
+            "You have made changes to the inventory since you last updated the lock"
```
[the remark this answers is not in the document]
apyrgio: Nice, I agree.

docs/developer/inventory.md
apyrgio: Blatantly copy-pasted from Poetry, who have copy-pasted it from some other Unix tools :grimacing: If you feel too strong about it, I can change it though.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+                    if any(fnmatch.fnmatch(m.name, glob) for glob in globs)
+                ]
+                if not members:
+                    raise InvException("Globs did not match any files in the archive")
+                tar.extractall(path=destination, members=members, filter="data")
+        except Exception as e:
+            raise InvException(f"Error extracting '{archive_path}': {e}") from e
+    elif ft == "zip":
+        try:
+            with zipfile.ZipFile(archive_path, "r") as zip_ref:
```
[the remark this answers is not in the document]
almet: Ah yeah, there is that as well, which could also be done with another `functools.partial`. Use your best judgement here; I can see why you would like to have the code separated as well. I generally don't like it as it looks like a code smell to me, but we could well be in a case where it doesn't matter ;-)

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+    plus the asset name.
+    """
+    # Calculate a unique hash for this URL name, so that it doesn't class with different
+    # versions of the same asset.
+    url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
+
+    # Get the name of the asset from the URL, which is typically the last part of the
+    # URL. However, if the asset is the GitHub-generated zipball/tarball, use a
+    # different naming scheme.
+    parsed = url.split("/")
```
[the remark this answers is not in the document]
apyrgio: Sure, I can add a sanity check.

dev_scripts/inventory.py
apyrgio: I agree. I've actually never tried them before, but I'll give them a shot.

docs/developer/inventory.md
apyrgio: I originally called it `install`, would that look better to you? `sync` is probably wrong after all, since we cannot remove assets for other platforms, as is.

docs/developer/inventory.md
apyrgio: Yeah, yeah, let's go with `-d`.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+            return toml.load(CONFIG_FILE.open("r"))
+        except Exception as e:
+            msg = f"Could not load configuration file '{CONFIG_FILE}': {e}"
+            raise InvException(msg) from e
+    elif PYPROJECT_FILE.exists():
+        # Then, attempt to read the configuration from pyproject.toml.
+        try:
+            config = toml.load(PYPROJECT_FILE.open("r"))
+        except Exception as e:
+            msg = f"Could not load configuration file '{PYRPROJECT_FILE}': {e}"
```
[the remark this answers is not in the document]
apyrgio: Fixed in e63bc8a1, in which I've bundled several other small fixes.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+        h.update(chunk)
+    return h.hexdigest()
+
+
+def cache_file_path(url):
+    """
+    Generate a safe cache file path for a given URL, using the SHA-256 hash of the path,
+    plus the asset name.
+    """
+    # Calculate a unique hash for this URL name, so that it doesn't class with different
```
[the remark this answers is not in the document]
apyrgio: Fixed in e63bc8a1, in which I've bundled several other small fixes.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+        raise InvException(
+            f"Unexpected response when fetching releases for repo '{repo}': HTTP"
+            f" {response.status_code}"
+        )
+    releases = response.json()
+
+    matching = []
+    for release in releases:
+        tag = release.get("tag_name", "")
+        # Strip any prefix 'v' if necessary
```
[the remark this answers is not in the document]
apyrgio: Fixed in e63bc8a1, in which I've bundled several other small fixes.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+    * 'globs': A list of patterns that will be used to match members in the archive.
+      If a member does not match a pattern, it will not be extracted.
+    * 'flatten': A boolean value. If true, after extraction, move all files to the
+       destination root.
+    * 'filetype': The type of the archive, which indicates how it will get extracted.
+
+    For tarfiles, use filter="data" when extracting to mitigate malicious tar entries.
+    """
+    logger.info(f"Extracting '{archive_path}' to '{destination}'...")
+    ft = options["filetype"]
```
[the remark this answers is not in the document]
apyrgio: Fixed in e63bc8a1, in which I've bundled several other small fixes.

dev_scripts/inventory.py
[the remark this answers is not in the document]
apyrgio: Fixed in 62f5d7ebb87e59b41e775e11d854d62317efde66

dev_scripts/inventory.py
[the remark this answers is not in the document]
apyrgio: Added in df76e762447420827194eeb0285017a31853c5ff.

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+    # Return a string like 'windows/amd64' or 'linux/amd64' or 'darwin/amd64'
+    sys_platform = sys.platform
+    if sys_platform.startswith("win"):
+        os_name = "windows"
+    elif sys_platform.startswith("linux"):
+        os_name = "linux"
+    elif sys_platform.startswith("darwin"):
+        os_name = "darwin"
+    else:
+        os_name = sys_platform
```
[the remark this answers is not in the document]
apyrgio: Fixed in 26b3a0ec68a704bcf86cf28ab260a38a6f74f141

docs/developer/inventory.md
almet: I don't see where the `directory` option is used at all. Maybe you could point me to it? :thinking:

dev_scripts/inventory.py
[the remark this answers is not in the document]
apyrgio: I have heavily refactored this function in d7915fada11cfa8a333cd7a0ab99bd4a6dbf10f4. I have ironed out a few more details as well, besides factoring out some parts of the code.

As a result of some tests I did, I found out that working with zipballs is much faster than working with tarballs. For the Tesseract data, it takes 6.4 seconds in my machine to extract files from a tarball, whereas it takes 2 seconds to extract from a zipball. Therefore, I switched from to a zipball for our assets in f222bca29b304cac4325838cd62c50dcc26073a9.

dev_scripts/inventory.py
apyrgio: Done in f9580525f8734e28871c2e816dea6d43c64857d7.

almet (approved):
Hey, here is another round of review for this, as I would like to use it to bundle cosign in the ICU pull request.

I've used it without issues on my machine, so :partying_face: !

I have two remarks, but they're minor and could be fixed later:

The CLI `--help` could be enhanced to have something more compelling than what we have right now, which is this:

```
usage: inventory.py [-h] {lock,sync,list} ...

Inventory

positional arguments:
  {lock,sync,list}
    lock            Update lock file from config
    sync            Sync assets as per lock file
    list            List assets for a platform

options:
  -h, --help        show this help message and exit
```

More specifically, I think something that just presents the tool might be useful.

The second thing is the fact that syncing the assets can take some time, and it might be useful to see a progressbar of some sort. It seem that [`tqdm`](https://pypi.org/project/tqdm/) might be a good dependency to add here.

Otherwise, it all looks good!

docs/developer/inventory.md
almet: Yep, install seem a better naming :+1:

dev_scripts/inventory.py
apyrgio: Ok, I took a swing at 96ecb3c34047bf9d5a9e1e93b7b615702ecf616f

docs/developer/inventory.md
apyrgio: Fixed in 64afdb1aa116eece6e1e4048bf31467029d1acdc.

docs/developer/inventory.md
apyrgio: Damn, it totally slipped! Check out b0ca4c4b6b8d4ed5ffb2227d7220b198ade01b60 for a fix.

docs/developer/inventory.md
apyrgio: Changed in dd453f1f

dev_scripts/inventory.py
```diff
@@ -0,0 +1,793 @@
...
+            config = toml.load(PYPROJECT_FILE.open("r"))
+        except Exception as e:
+            msg = f"Could not load configuration file '{PYRPROJECT_FILE}': {e}"
+            raise InvException(msg) from e
+
+        # If the pyproject.toml file does not have a [tool.inventory] section, return an
+        # error.
+        config = config.get("tool", {}).get("inventory")
+        if not config:
+            raise InvException(
```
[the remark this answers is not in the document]
apyrgio: Fixed in 5bbdd48dde2404048365f4684561da121ff9a804.

apyrgio:
Hey Alexis. Thanks a lot for the review comments, they were pretty spot on! I've tried to address the main bulk of them, and I'd like to comment here on the ones I haven't addressed:

> The dependencies to our code are so little that I think this could actually be packaged as a separate project instead of something that's inside Dangerzone itself.

Definitely, I think we can have a `freedomofpress/assets` repo, or something like that, and maybe even push to PyPI. Actually, that's what I plan to do once this is ready to merge. Open a new GitHub repo and copy the code there, with its own `pyproject.toml` and license (probably MIT). If you don't mind, I'd like to keep the `inventory.py` name just for the review, and I'll change it to `assets` once I copy it to its own repo.

> The CLI --help could be enhanced to have something more compelling than what we have right now,
> [...]
> The second thing is the fact that syncing the assets can take some time, and it might be useful to see a progressbar of some sort. It seem that [tqdm](https://pypi.org/project/tqdm/) might be a good dependency to add here.

Yes and yes. For the first part, I wanted to somehow reuse what I've written in the docs, but I didn't want to copy-paste right away. Do you have a better way to tackle this?

For the second part, I agree. It's one of the things that were cut from the PR because I didn't want to over-engineer it (didn't know if it would be useful for projects other than DZ). That's something I can add later on, if you don't mind.

almet:
Okay, great! No probs to add things later. It's already great :raised_hands: 

> Do you have a better way to tackle this?

I'm not sure it will be useful here, but that the second time in the day that I think about it, so…  In other projects, we've been using `cog` for this, to avoid recopying things. You can see an example of that here ([code](https://framagit.org/framasoft/framaspace/argos/-/blob/main/docs/cli.md?ref_type=heads&plain=1), and [resulting doc](https://argos-monitoring.framasoft.org/cli.html)), but of course that works when you have something to handle the docs (here we use sphinx)

dev_scripts/inventory.py
almet: Nice ;-)

almet:
Done in #1153

PR #1153, opened 2025-05-20
Use `assets` tool to manage GitHub assets

### LINKED ISSUE #1146: Vendor Dangerzone assets in a uniform way

apyrgio:
_(previous version of this PR: #1149)_

In this PR we:
1. Add a dependency to the [`assets`](https://github.com/freedomofpress/assets) tool in our `pyproject.toml`.
2. Add an [asset entry](https://github.com/freedomofpress/assets?tab=readme-ov-file#configuration) in our `pyproject.toml`, as well as an `assets.lock` lock file, for the Tesseract data that we package.
3. Replace the `install/common/download-tessdata.py` script with a call to `poetry run assets install` throughout our codebase.

Fixes #1146


BUILD.md
```diff
@@ -75,10 +75,10 @@ Build the latest container:
 python3 ./install/common/build-image.py
 ```
 
-Download the OCR language data:
+Download the necessary assets:
```
almet, line 78:
```suggestion
Dangerzone depends on some assets that should be downloaded in order to run (think binaries and others resources). This can be done with the following command:
```

almet (approved):
We've discussed that PR locally (yay!) and we're all good here! Let's merge it :tada:

PR #1161, opened 2025-05-26
Docs: fix Debian and Ubuntu installation instructions

### LINKED ISSUE #1158: Ubuntu 24.04 LTS sq failing
### What happened?

I am currently trying to install dangerzone on a fresh Ubuntu 24.04 LTS install and the commands for sq neither find "search" (but fetch seems to be the correct thing)
nor dearmor or sq packet subcommand. 

### Linux distribution

Ubuntu 24.04 LTS

### Dangerzone version

None

### Podman info

```shell
None
```

### Document conversion logs

```shell
None
```

### Additional info

None

almet:
When using gpg to retrieve the keys in `sudo` mode, one needs to run `dirmngr` in the background first. Also change the order in which the different distributions appear and make the separation between the two clearer.

Fixes #1158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## [Unreleased](https://github.com/freedomofpress/dangerzone/compare/v0.9.0...HEAD)
 
+## Fixed
+
+- Fix Debian and Ubuntu installation instructions ([#1161](https://github.com/freedomofpress/dangerzone/pull/1161))
+
 ## Changed
 
 - Update installation instructions (and CI checks) for Debian derivatives ([#1141](https://github.com/freedomofpress/dangerzone/pull/1141))
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -110,10 +110,22 @@ Dangerzone is available for:
   </tr>
 </table>
 
-First, retrieve the PGP keys. The instructions differ depending on the specific
-distribution you are using:
+First, retrieve the PGP keys.
 
-For Debian Trixie and Ubuntu Plucky (25.04), follow these instructions to
+**The instructions differ depending on the specific distribution you are using**
+
+For all distributions **except Debian Trixie and Ubuntu Plucky (25.04)**:
+
+```sh
+sudo apt-get update && sudo apt-get install gnupg2 ca-certificates -y
+sudo mkdir -p /etc/apt/keyrings/
+sudo dirmngr&
+sudo gpg --keyserver hkps://keys.openpgp.org \
+    --no-default-keyring --keyring /etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg \
+    --recv-keys "DE28 AB24 1FA4 8260 FAC9 B8BA A7C9 B385 2260 4281"
+```
+
+For **Debian Trixie** and **Ubuntu Plucky (25.04)**, follow these instructions to
 download the PGP keys:
 
 ```bash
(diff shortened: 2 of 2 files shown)


apyrgio:
Hm, the `dirmngr &` fix looks interesting, but I have a slight distaste to leaving things running in the background. Coupled with the fact that it's easy for our users to get confused about when to use the one instruction block for Trixie and Plucky, and when the other, I wanted to simplify things.

The original problem was that GnuPG creates a keyring in keybox format, which is not parseable in latest versions of `apt-secure` [^1].  That's easily solved though, if you prepend the `-keyring` argument with `gnupg-ring:` [^2]. The rest of the issues, which is basically sidestepping `dirmngr` can also be resolved with a random `--homedir`, as mentioned in the GPG docs [^3].

So, this gives us this block, which works everywhere:

```bash
sudo apt-get install -y wget gpg ca-certificates
sudo mkdir -p /etc/apt/keyrings
sudo gpg --keyserver hkps://keys.openpgp.org \
    --no-default-keyring --no-permission-warning --homedir $(mktemp -d) \
    --keyring gnupg-ring:/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg \
    --recv-keys DE28AB241FA48260FAC9B8BAA7C9B38522604281
sudo chmod +r /etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg
. /etc/os-release
echo "deb [signed-by=/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg] \
    https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main" \
    | sudo tee /etc/apt/sources.list.d/fpf-apt-tools.list
```

I have tested it with the following Dockerfile, which we can run in our CI as well, so that we can properly test our installation instructions:

```Dockerfile
FROM debian:trixie-slim

# Create a user with sudo powers.
# NOTE: Use UID 1001 because Ubuntu images already contain UID 1000.
RUN apt-get update && apt-get install -y sudo adduser

RUN addgroup --gid 1001 user
RUN addgroup wheel
RUN adduser --uid 1001 --ingroup user --shell /bin/true \
    --disabled-password --home /home/user user
RUN adduser user wheel

RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers

USER user
WORKDIR /home/user

# Now run the installation instructions
RUN sudo apt-get install -y wget gpg ca-certificates
RUN sudo mkdir -p /etc/apt/keyrings
RUN sudo gpg --keyserver hkps://keys.openpgp.org \
    --no-default-keyring --no-permission-warning --homedir $(mktemp -d) \
    --keyring gnupg-ring:/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg \
    --recv-keys DE28AB241FA48260FAC9B8BAA7C9B38522604281
RUN sudo chmod +r /etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg
RUN . /etc/os-release && echo "deb [signed-by=/etc/apt/keyrings/fpf-apt-tools-archive-keyring.gpg] \
    https://packages.freedom.press/apt-tools-prod ${VERSION_CODENAME?} main" \
    | sudo tee /etc/apt/sources.list.d/fpf-apt-tools.list

RUN sudo apt-get update
RUN sudo apt-get install -y dangerzone
```

Although `sq` is the way to go for newer Debian versions - and much better than the esoteric GnuPG options - I'm afraid incorporating it in our instructions may confuse users, so I would recommend going with a single set of instructions.

If you agree, I can update the PR with this suggestion and test it in various platforms, since you have lots of stuff on your hands with ICU.

CCing @legoktm, who suggested adding the `sq` command, to see if he's ok with this change.


[^1]: https://github.com/freedomofpress/dangerzone/issues/1052#issuecomment-2594065503
[^2]: https://stackoverflow.com/a/72278246
[^3]: https://www.gnupg.org/documentation/manuals/gnupg/Ephemeral-home-directories.html

almet:
Closing in favor of  #1163

PR #1176, opened 2025-06-09
Update dnf command to add repo

inknos:
Since Fedora 41 dnf is replaced by dnf5 which uses an updated syntax and includes config-manager as a core command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## Changed
 
+- (Docs) Update installation instructions for Fedora. `dnf config-manager` is not a plugin ([#1176](https://github.com/freedomofpress/dangerzone/pull/1176))
 - (Docs) Update installation instructions (and CI checks) for Debian derivatives ([#1141](https://github.com/freedomofpress/dangerzone/pull/1141),
   [#1163](https://github.com/freedomofpress/dangerzone/pull/1163))
 - (Docs) Point to relevant install sections in the OS-support table ([#1160](https://github.com/freedomofpress/dangerzone/pull/1160))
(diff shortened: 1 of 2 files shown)


almet:
Hi, and thanks for the proposed changes.

As we support Fedora 40, 41 and 42, it seems probably best to wait until we drop Fedora 40 and then move to the new updated instructions you provide here. Right now it seems a bit too early, as these instructions won't work on F40.

Of course, unless the current state of things is broken, in which case don't hesitate to tell us more about the specifics :-)

Thanks!

inknos:
> Note
> 40 is EOL since 2025.05.13. [Fedora 40 Schedule](https://fedorapeople.org/groups/schedule/f-40/f-40-key-tasks.html) and [Release EOL](https://docs.fedoraproject.org/en-US/releases/eol/)

Hey,

If you support F40, then it makes sense to keep it for a bit :) . Just noted that by following the installation instructions, you will have dnf5 and dnf-3 in the system, which could be unwanted behavior. Not a huge deal, and the packages are made to not conflict on the same system.

Possibly related question: did you consider releasing dangerzone as an official package in Fedora?

almet:
Thanks I haven't noted the EOL date is past us :-)
Then things are different, you're right!

So, I do think we should remove support for F40 altogether, which means merging your changes and removing mentions of it in our docs.

We can start by merging your changes, and I've opened [a follow-up issue](https://github.com/freedomofpress/dangerzone/issues/1178) to track end of support for F40 in general.

almet (approved):

almet:
One missing thing before merging is to update the CHANGELOG file and add an entry about your changes (we added this to ease the release workflow). I can also do it myself if you don't want to bother.

(The security-scan errors seem to be due to a critical vulnerability in an unreleased image so we're safe to ignore them in the scope of this PR)

inknos:
> One missing thing before merging is to update the CHANGELOG file and add an entry about your changes (we added this to ease the release workflow). I can also do it myself if you don't want to bother.

all right, @almet , I updated the changelog. let me know :)

almet:
Thanks @inknos!

Issue #1178, opened 2025-06-10
Drop Fedora 40 support


almet:
Security support for Fedora 40 ended 4 weeks ago (13 May 2025), so we could remove it in the next release.

This mainly means:

- Updating the documentation ;
- Updating the CI checks ;
- Potentially removing the F40 packages from our [yum-tools-prod repository](https://github.com/freedomofpress/yum-tools-prod), potentially grouping with the move to the R2 infrastructure there.

PR #1183, opened 2025-06-12
Fix Fedora install instructions

deeplow:
Looks like this was [fixed in CI](https://github.com/freedomofpress/dangerzone/commit/eaef95b7748191878cf5c71fad9008b0eb8e7e90) but missed in the official instructions.

(signed with Github's web key. Feel free to force-push to re-sign)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -167,7 +167,7 @@ sudo apt install -y dangerzone
 Type the following commands in a terminal:
 
 ```
-sudo dnf config-manager add-repo --from-repofile=https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo
+sudo dnf config-manager addrepo --from-repofile=https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo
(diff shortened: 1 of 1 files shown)


almet:
This seems related to the recent changes in https://github.com/freedomofpress/dangerzone/pull/1176.

The situation might not be clear because `main` doesn't state clearly that F40 is out of support. This whole situation will be solved once we resolve #1178.

 Out of curiosity, which Fedora version are you using?

deeplow:
> The situation might not be clear because main doesn't state clearly that F40 is out of support. This whole situation will be solved once we resolve https://github.com/freedomofpress/dangerzone/issues/1178.

:+1: but actually I realized the reason the command wasn't working was because it's `addrepo` instead of `add-repo`. I have changed the PR to reflect that (may need squashing).

> Out of curiosity, which Fedora version are you using?

41 (on Qubes)

almet (approved):

almet:
Thanks, just squashed and merged!

PR #1195, opened 2025-07-02
Ignore CVE-2025-49794 and CVE-2025-49796

apyrgio:
Ignore these two libxml2 CVEs, because they are actually minor, and there is no fix for them yet.


almet (approved):
Thanks!

PR #1199, opened 2025-07-07
Bump version to 0.9.1

almet:


apyrgio:
I don't see an updated `poetry.lock`. Do you have one handy, or do you want me to do it after this PR is merged?

Also, what about the latest WiX release? Should we use the 6.0.1 one?

Dockerfile.env
apyrgio, line 1: Note that we also need to render the `Dockerfile` based on `Dockerfile.env`. I assume that our CI checks will catch this though...

almet:
> I don't see an updated poetry.lock. Do you have one handy, or do you want me to do it after this PR is merged?

Actually this is already up to date because [we did it last week](https://github.com/freedomofpress/dangerzone/commit/18a5e8056d8b58e89cab35a8d9d9836af1765d62) (poetry lock doesn't return anything more)

> Also, what about the latest WiX release? Should we use the 6.0.1 one?

Ah, the docs mention that we should keep with the 5.X series, so I stopped there, as I assumed that going to 6.X would be a major change and not needed now. Probably interesting to have a look at it as a subsequent task though.

Dockerfile.env
almet: Ah! Yeah sorry I forgot (I knew I would!). Just updated the PR and grouped the commit that change `Dockerfile`.

almet:
The failures seem related to the bumping of the sandbox versions, currently trying to find what's the culprit on my local machine.

PR #1207, opened 2025-07-09
Bump PyMuPDF version to 1.25.3

almet:


pyproject.toml
```diff
@@ -20,7 +20,7 @@ python = ">=3.9,<3.14"
 click = "*"
 platformdirs = "*"
 PySide6 = "^6.7.1"
-PyMuPDF = "^1.23.3" # The version in Fedora 39
+PyMuPDF = "^1.25.3" # The version in Fedora 41
```
apyrgio, line 23: I wonder, what does this bump translate to. I think `poetry.lock` will use the latest PyMuPDF version in any case, right?

almet: Yes this is right; It sets the lower version and then will not jump to breaking major versions. But... in the case of Pymupdf it doesn't actually mean anything since they don't follow semver.

I would go for * here (and actually for a lot other deps, I have changes locally that reflect that), but wanted to check with you first, as I'm not sure to understand the reason this is not free range here for PyMuPDF.

apyrgio: I recall that we use a lower version here to help rpm/deb know what's our lower bound. Else, we would be using `"*"`, I agree.

apyrgio (approved):

PR #1216, opened 2025-07-24
Use archived debian bullseye backports for testing

### LINKED ISSUE #1213: Dev scripts are broken for Debian bullseye
bullseye-backport [was recently discontinued](https://lists.debian.org/debian-devel-announce/2025/06/msg00003.html), [as advertised by the FAQ about LTS](https://lts-team.pages.debian.net/wiki/FAQ.html#who-fixes-security-issues-with-packages-in-backports):

> As a result, backports for a given release are closed after 3 years so as to not overload (volunteer) backport maintainers with the responsibility of maintaining their backports for 5 years.

The `env.py` script is currently relying on backports for  `libqt6gui6` and `pipx`, which are used only when running the tests (e.g. they are not run-dependencies).

We could actually install `pipx` with `pip install pipx`, but we would have more issues with `libqt6gui6`.

That asks us the following questions (each of which has a separate issue):

- **How should we fix the current situation?**. Using [debian-archives](https://archive.debian.org/) will get us back in a working situation, but the archives there won't be updated.
- **Do we still stand behind this LTS support?** Our support matrix says we support  "Current stable, Oldstable and LTS releases" for Debian. When we did this call, we weren't aware that the backport where actually following another route, meaning that the tooling we need for testing might get harder to get together.
- There are discrepancies between the versions of Qt we're testing VS the versions that are actually installed (e.g. for bullseye we're testing with Qt6 whereas the installed version is Qt5)

almet:
Debian bullseye stopped providing backports for this LTS release, breaking our testing scripts. This makes use of the archived backports, which is not an ideal situation but has the benefit of making us having tests for this platform.

Other solutions are currently being discussed, so this could be seen as a temporary fix.

Fixes #1213

### DIFF
--- a/dev_scripts/env.py
+++ b/dev_scripts/env.py
@@ -78,6 +78,16 @@ COPY apt-tools-prod.sources /etc/apt/sources.list.d/
 COPY apt-tools-prod.pref /etc/apt/preferences.d/
 """
 
+# We are relying on Debian bullseye archived backports for testing only.
+# They are not required to build or run Dangerzone but are useful to run
+# the tests because they provide Qt6.
+# See https://github.com/freedomofpress/dangerzone/issues/1213
+DOCKERFILE_USE_BULLSEYE_BACKPORTS = r"""
+RUN apt-get update && apt-get install -y ca-certificates
+RUN echo "deb https://archive.debian.org/debian/ bullseye-backports main" \
+    >> /etc/apt/sources.list
+"""
+
 # FIXME: Do we really need the python3-venv packages?
 DOCKERFILE_BUILD_DEV_DEBIAN_DEPS = r"""
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


dev_scripts/env.py
```diff
@@ -78,6 +78,16 @@
 COPY apt-tools-prod.pref /etc/apt/preferences.d/
 """
 
+# We are relying on Debian bullseye archived backports for testing only.
+# They are not required to build or run Dangerzone but are useful to runs
```
apyrgio:
```suggestion
# They are not required to build or run Dangerzone but are useful to run
```

apyrgio:
Great, that's exactly how I wanted to deal with this situation, down to the "reinstate bullseye instead of bullseye-backports container image". Other than a nit and a need for a changelog entry (most likely dev-related) feel free to merge once the tests pass (well, except for our security scans :roll_eyes: )

PR #1217, opened 2025-07-29
Ignore CVE-2025-6965 in our grype scans

almet:
libsqlite is required by python, but we are not using nor exposing the parts that make use of it, and so are not affected.


apyrgio (approved):

apyrgio:
Thanks for this fix Alexis, and I still owe the security-related doc.

PR #1219, opened 2025-08-04
Prioritize installing PySide6 dependencies

### LINKED ISSUE #1218: QWidget: Must construct a QApplication before a QWidget
## Background

Now that Debian Bullseye will become `oldoldstable` (#1214), and Ubuntu Focal is no longer supported, the number of distros that use PySide6 instead of PySide2 are considerably more (#211).

Still, our Debian Dangerzone packages require PySide2 as a package dependency:

https://github.com/freedomofpress/dangerzone/blob/a2035a2a369b116c715c34eac461b1d52e37ef59/debian/control#L12

Our code however can work either with PySide6 or PySide2 (we actually install PySide6 on Windows/macOS environments).

## Problem

There seems to be a [Dangerzone bug](https://gitlab.tails.boum.org/tails/tails/-/issues/21095) in the upcoming Tails 7.0-RC1 version, which will be based on Debian Trixie, and where PySide6 is present due to some other GUI apps:

```
$ dangerzone
[DEBUG] Inferred system color scheme as OSColorMode.LIGHT
QWidget: Must construct a QApplication before a QWidget
Aborted
```

I confirm that I can reproduce this bug in a local Tails 7 installation, and that when I force Dangerzone to use the PySide6 packages, this issue is resolved. Since there's nothing Tails-specific in this bug, it could very well be that this may start affecting a growing number of our Linux users, outside Tails.

apyrgio:
Since we prioritize PySide6 libraries at import time, we should prioritize installing PySide6 dependencies, in systems where these exist.

Fixes #1218


almet:
Fedora repositories are currently timeout-ing, and it obviously is unrelated to the changes here, so I think we can safely ignore that and consider the CI green:

```
registry.fedoraproject.org: Get \"[https://registry.fedoraproject.org/v2/\](https://registry.fedoraproject.org/v2//)": dial tcp 38.145.32.21:443: i/o timeout"
```

PR #1247, opened 2025-09-17
Feat: Capture commands output

### LINKED ISSUE #1236: Capture commands and their output
We should have a unified way to capture the commands that Dangerzone spawns, as well as their output. This is important mostly for debugging and error reporting reasons, but can even help with progress reports, if we don't have any other way to create a progress bar (see #1164).

almet:
Monkeypatch `subprocess.run` at runtime in order to capture the stdout and stderr.

I've tested succesfully that the commands for `cosign` and `podman machine` have been captured properly and shown in the new shiny log window.

Fixes #1236


dangerzone/capture_output.py
```diff
@@ -0,0 +1,54 @@
...
+def patched_subprocess_run(  # type: ignore[no-untyped-def]
+    args,
+    *,
+    stdout=None,
+    stderr=None,
+    capture_output=False,
+    **other_popen_kwargs,
+) -> subprocess.CompletedProcess:
+    # capture_output and stderr/stdout can't be specified at the same time
+    if not capture_output:
```
apyrgio: In this case, if `capture_output=False` and `stdout=None` and `stderr=None`, the intent of the user is to print the output to the standard streams, potentially even live. However, I think the end result here is that the stdout/stderr will be logged only once the process finishes.

For instance, when `podman pull` runs:
* GUI users will see the output only once the process is over. Unfortunately, this does not make it a feasible progress bar alternative.
* CLI users will similarly not see any output, whereas they were able to see it in the past.

dangerzone/capture_output.py
```diff
@@ -0,0 +1,54 @@
...
+        capture_output = True
+
+    if stdout == subprocess.PIPE or stderr == subprocess.PIPE:
+        capture_output = False
+
+    process = original_subprocess_run(
+        args,
```
apyrgio: It would be really nice if we could capture the full command that will be run, and its exit code once it completes.

dangerzone/capture_output.py
```diff
@@ -0,0 +1,54 @@
...
+        **other_popen_kwargs,
+    )
+
+    def _copy_pipe(source: Any) -> None:
+        if isinstance(source, io.IOBase):
+            # tell() and then seek() the stream to restore it
+            # to the same place it was before copying it.
+            position = source.tell()
+            log.debug(source.read())
+            source.seek(position)
```
apyrgio: I wonder if this code block is used. `IOBase.{tell,seek}()` works only for seekable files, else they raise an `OSError` (https://docs.python.org/3/library/io.html#io.IOBase.seekable). I think that in Dangerzone, we rarely use seekable files in `subprocess.run`.

almet:
Here is the new version of this code, which now captures the output as it goes using threads, which can be useful to track processes that take some time to complete.

dangerzone/capture_output.py
```diff
@@ -0,0 +1,107 @@
...
+            super().__init__(*args, **kwargs)
+            return
+
+        # Read the stdout and stderr as streams (text=True and bufsize=1)
+        kwargs["stdout"] = subprocess.PIPE
+        kwargs["stderr"] = subprocess.PIPE
+        kwargs["text"] = True
+        kwargs["bufsize"] = 1
+
+        # Patch open() to fake a tty during the Popen init
```
apyrgio: This is no longer the case, right?

dangerzone/container_utils.py
```diff
@@ -384,19 +385,13 @@ def expected_image_name() -> str:
...
 
-def container_pull(
-    image: str, manifest_digest: str, callback: Optional[Callable] = None
-) -> None:
+def container_pull(image: str, manifest_digest: str) -> None:
     """Pull a container image from a registry."""
     podman = init_podman_command()
     process = podman.run(
-        ["pull", f"{image}@sha256:{manifest_digest}"], wait=False, text=True, bufsize=1
+        ["pull", f"{image}@sha256:{manifest_digest}"], wait=False, capture_output=False
```
apyrgio: We can also set `wait=True` here, since we're undoing the previous code.

dangerzone/capture_output.py
```diff
@@ -0,0 +1,107 @@
...
+    ):
+        """Patch the subprocess.Popen to generate log entries:
+
+        - If stdout and stderr are defined, don't alter the behavior ;
+        - Otherwise, create PIPEs for stdout and stderr, and then start threads in
+          the background to consume stdout and stdin, and pass it to `log.debug()`
+        """
+        stdout = kwargs.get("stdout")
+        stderr = kwargs.get("stderr")
+        log.debug(shlex.join(args[0]))
```
apyrgio: Can we maybe add something more descriptive here?

```
Running <command>
```

dangerzone/capture_output.py
```diff
@@ -0,0 +1,107 @@
...
+        thread_out = threading.Thread(target=_consume_pipe, args=(self.stdout,))
+        thread_err = threading.Thread(target=_consume_pipe, args=(self.stderr,))
+
+        original_process_poll = self.poll
+
+        # Patch the .poll() method to wait for the threads to finish
+        def patched_poll(*args, **kwargs):  # type: ignore[no-untyped-def]
+            original_process_poll(*args, **kwargs)
+            for t in (thread_out, thread_err):
+                t.join()
```
apyrgio: I think there's something off here. First of all, we don't return the exit code. Second, we should check if the command has exited before joining the threads, else we'll block in `poll()` until the command is over (which is `wait()`'s job).

apyrgio: Also, I suggest logging the exit code of this function here, since it may prove helpful.

dangerzone/container_utils.py
almet: Actually, if putting `wait=True` the implementation of the podman command isn't returning the process, hence why I let it.

dangerzone/capture_output.py
almet: Actually I prefer just having this way, because it's just... easier to copy and paste. What's the rationale for having it prefixed with `Running` in your case? If the goal is to see that's a command, my experience is that the current command is enough (to me, at least). Or is it for a different purpose?

dangerzone/capture_output.py
almet: Correct! Removed in 6458161d125f874dcad0ffcf49b1d9e4a3d35f66

dangerzone/capture_output.py
almet: Updated in 1722b213501e34de023202364e288934ada26991

dangerzone/capture_output.py
almet: Updated in 1722b213501e34de023202364e288934ada26991

almet: (See the new diff for updates)

dangerzone/container_utils.py
apyrgio: Oh god, you're right. I wonder if we need something equivalent for `podman load` then, or `podman machine start`.

almet: It really depends if the caller needs to read the `process` to get some info or not, I guess?

dangerzone/capture_output.py
apyrgio: Mostly to find more easily command invocations in long lines of logs (or even grep for them). But that's ok, it's minor after all.

dangerzone/capture_output.py
apyrgio: Looks good!

dangerzone/capture_output.py
almet: (i did update it, fwiw, thanks)

dangerzone/container_utils.py
apyrgio: In those two cases, the caller does not need any info. But you know what, I think I overlooked something. You're saying that with `wait=True` the podman command does not return the process, which is true. Why is that a problem though? The rest of the commands:

```
    process.wait()
    if process.returncode != 0:
 ```
 
 can be covered by a `try: ... except CalledProcessError`. We can keep the wait as is, if you want to, but I wanted to be sure I'm not missing something.

File: dangerzone/capture_output.py [hunk 8]
```diff
@@ -0,0 +1,109 @@
...
+        self.stderr = None
+
+
+def patched_subprocess_run(  # type: ignore[no-untyped-def]
+    *args, **kwargs
+) -> subprocess.CompletedProcess:
+    """
+    Patch subprocess.run to log stdout and stderr and the command that was run.
+    """
+    process = original_subprocess_run(
```
almet: We need to also capture the errors if they happen, like this:

```traceback
Error calling Python override of QThread::run(): Traceback (most recent call last):
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/podman/command/cli_runner.py", line 261, in run_raw
    ret = subprocess.run(
          ^^^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/capture_output.py", line 89, in patched_subprocess_run
    process = original_subprocess_run(
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 573, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/opt/podman/bin/podman', 'image', 'list', '--format', '{{ .Digest }}', 'ghcr.io/almet/dangerzone/dangerzone-testing']' returned non-zero exit status 125.

The above exception was the direct cause of the following exception:

  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/startup.py", line 232, in run
    if task.should_skip():
       ^^^^^^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/startup.py", line 166, in should_skip
    return installer.get_installation_strategy() == InstallationStrategy.DO_NOTHING
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/updater/installer.py", line 105, in get_installation_strategy
    podman_images = runtime.list_image_digests()
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/container_utils.py", line 244, in list_image_digests
    podman.run(
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/podman/command/command.py", line 80, in run
    return self.runner.run(
           ^^^^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/podman/command/cli_runner.py", line 217, in run
    return self.run_raw(
           ^^^^^^^^^^^^^
  File "/Users/alexis/dev/fpf/dangerzone/dangerzone/podman/command/cli_runner.py", line 267, in run_raw
    raise errors.CommandError(e) from e
dangerzone.podman.errors.exceptions.CommandError: The Podman process failed with the following error: Command '['/opt/podman/bin/podman', 'image', 'list', '--format', '{{ .Digest }}', 'ghcr.io/almet/dangerzone/dangerzone-testing']' returned non-zero exit status 125.
Stderr: b'Cannot connect to Podman. Please verify your connection to the Linux system using `podman system connection list`, or try `podman machine init` and `podman machine start` to manage a new Linux VM\nError: unable to connect to Podman socket: failed to connect: dial tcp 127.0.0.1:64148: connect: connection refused\n'
qt.qpa.drawing: Layer-backing is always enabled.  QT_MAC_WANTS_LAYER/_q_mac_wantsLayer has no effect.
```

apyrgio: If I get it right, our `subprocess.run` wrapper does not capture the stdout/stderr if the original `subprocess.run` throws an exception, right? I think that's important, else we risk not seeing error messages in case a command fails.

It looks to me that this calls for:

```
try:
    process = original_subprocess_run(...)
except CalledProcessError as e:
    # Log stdout/stderr>
    raise

# Log stdout, stderr from process.{stdout,stderr}
return process
```

almet: Yep, exactly the issue.

almet: Should be addressed in e3cc0957298eb6d372d0023686f6b03acd4def55 (where I discovered the existence of [logger.exception](https://docs.python.org/3/library/logging.html#logging.exception))

dangerzone/container_utils.py
almet: See 75c490fd87ddafcf1d55154b01e87c7eefc2f90e about this.

apyrgio: Right on, thanks!

dangerzone/capture_output.py
apyrgio: Hm, we miss the stdout, but that's ultimately ok. Thanks for this :slightly_smiling_face:

PR #1225, opened 2025-08-05
Embed Podman and drop Docker support

apyrgio:
This PR introduces a very core change to Dangerzone, that is required for the Independent Container Updates feature (https://github.com/freedomofpress/dangerzone/issues/1006). It adds first class support for Podman, by embedding it into our Windows / macOS packages, and using it internally to create VMs in these platforms where we can run containers. As a result, we drop Docker support, since it's no longer needed, and because it's not compatible with the format of the image we use in the Independent Containers Update feature.

This PR also introduces some visual changes. First, it allows users to select documents without waiting for the container installation to finish, or the Podman machine to start. We have measured that after the first time that Dangerzone runs, where initialization takes a while, the rest of the times require a few seconds to get Dangerzone starting, so we figued it's fine to let the user select documents, and do some work in the background. Second, we now have a widget that shows all the logs of our background tasks, which will help when debugging startup issues.

Finally, the integration of Podman allows us to do end-to-end CI tests on Windows and macOS platforms, which was previously highly difficult. This means that each PR, and more critically all of our Podman machine logic, will be tested on the two main platforms we are supporting, and we will catch errors sooner.

Here is a list of open issues that this PR affects:
* Closes #1170
* Closes #1172
* Refs containers/podman-py#558
* Closes #1168
* Fixes #118
* Fixes #1045
* Fixes #1145
* Fixes #1171
* Fixes #1184
* Closes #253
* Closes #470
* Closes #1132
* Fixes #1009
* Closes #668

PR #1251, opened 2025-09-24
Generate a new container image in the CI

almet:
This is done to enable the end to end tests to check the upgrade procedure.


almet:
After a discussion with @apyrgio, we've decided to merge this PR after we've merged the bigger #1225 PR rather than before.

almet:
After a discussion with @apyrgio, we've decided to merge this PR after we've merged the bigger https://github.com/freedomofpress/dangerzone/pull/1225 PR rather than before.

.github/workflows/build-push-image.yml
```diff
@@ -26,6 +26,10 @@ on:
         required: false
         type: string
         default: "dangerzone-tests"
+      update_container:
```
apyrgio: Maybe `trivially_update_container`, to show that this is for CI purposes only?

.github/workflows/build-push-image.yml
```diff
@@ -63,6 +67,11 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
+      
+      - name: Update the contents of the inner container if asked-for
+        if: ${{ inputs.update_container }}
+        run: |
+          echo "# A comment to generate a new container hash" >> dangerzone/conversion/doc_to_pixels.py
```
apyrgio: I would add a random value here, else the hash will be consistently the same. I believe we may have issues in subsequent runs if the hash of the updated images is the same, but I'm not sure.

.github/workflows/ci.yml
```diff
@@ -531,3 +549,86 @@ jobs:
           path: tests/test_docs/diffs/*.jpeg
         # Always run this step to publish test results, even on failures
         if: ${{ always() }}
+
+  end-to-end-test:
+    name: "Sandbox upgrade E2E test"
+    runs-on: ubuntu-latest
+    needs:
+      - build-new-container-image
+      - build-container-image
```
apyrgio, line 571: What happens if `build-new-container-image` ends quicker than `buld-container-image`?

.github/workflows/ci.yml
```diff
@@ -531,3 +549,86 @@ jobs:
...
+
+      - name: Update the settings so they check for updates
+        run: |
+          jq '.updater_check_all = true' ~/.config/dangerzone/settings.json > temp.json && mv temp.json ~/.config/dangerzone/settings.json
+
+      - name: Check the digest of the loaded podman image
+        run: |
+          podman images $(cat share/image-name.txt | head -1) --format "{{.Digest}}"
+
+      - uses: imjasonh/setup-crane@v0.1
```
apyrgio, line 627: I'm about to create a `mazette.toml` under `dev_scripts/` to handle tooling like this, instead of relying to third-party GitHub actions. No need to do something now, just an FYI.

apyrgio:
Couple of small comments, but other than that, I like this, and I think we'll sleep better once it's merged.

.github/workflows/build-push-image.yml
almet: I'm not sure it's really problematic, but I can add a random value just in case, yes!

.github/workflows/ci.yml
almet: As they depend on each other, I think we're covered!

.github/workflows/ci.yml
almet: Perfect, I was thinking the same!

.github/workflows/ci.yml
apyrgio: Whoops, I hadn't seen this, sorry!

.github/workflows/build-push-image.yml
almet: Thanks, updated and in https://github.com/freedomofpress/dangerzone/pull/1251/commits/bc4fc1e10e66b5c38739081b534d09ce223e5c8c

.github/workflows/build-push-image.yml
almet: Just did so in https://github.com/freedomofpress/dangerzone/pull/1251/commits/bc4fc1e10e66b5c38739081b534d09ce223e5c8c

almet:
Just rebased and applied your suggestions @apyrgio.

r?

apyrgio:
Looks good to me, feel free to merge once the CI tests pass.

Issue #1145, opened 2025-04-29
Embed Podman as standalone binary


apyrgio:
## Background

Dangerzone primarily relies on Docker Desktop in Windows and macOS, which is where the majority of our users are. Unfortunately, Docker Desktop has some limitations that we are not particularly pleased with:
* The Docker Desktop application which runs in the host is closed-source. Don't be confused with the Docker **client** that runs in the Linux VM, which is open-source.
* Docker Desktop has a license that makes it cumbersome to use in a corporate setting with lots of employees.
* Dangerzone cannot orchestrate starting Docker Desktop, and thus relied on the user doing so.

It's reasons like that there's a long standing issue to remove Docker Desktop support (https://github.com/freedomofpress/dangerzone/issues/118). Podman Desktop, on the other hand, does not have these limitations. It's fully open source, and the `podman machine` interface in particular was super interesting for us. That's why as of Dangerzone 0.9.0, we have [added](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) experimental Podman Desktop support.

Can we do better? Yes! We have successfully experimented with standalone Podman binaries in Windows and macOS, and we believe it's possible to embed them in our MSI/DMG installers, and thus not require from users to install a separate application, or open Podman Desktop before a conversion.

This issue will outline how we can implement this tigher Podman integration, by embedding it in our installers.

## Design

The main idea is the following. Dangerzone should expect that it ships with a standalone Podman binary (_GH issue pending_), and will use this binary to create a Linux VM on first use (`podman machine init`), start it when we open the application (`podman machine start`), and stop it before the application quits (`podman machine stop`).

If a user wants more control over the container runtime that Dangerzone will use, they can call `dangerzone-cli --set-container-runtime <runtime>`, and set their own. This is the opposite of how it's used in Dangerzone 0.9.0, where users typically call it to make Dangerzone use Podman instead of Docker.

apyrgio:
## Proposed flow

### First Run

The user opens Dangerzone after its installation. Dangerzone will immediately show a prompt that informs the user it needs to fetch resources from the internet in order to run, and periodically update them. If the user does not agree, then we will effectively treat their installation as air-gapped. If they agree, we will assume that they want to auto-update. In the future, we should add an in-between state, where Dangerzone can perform network calls, but they will be notified before a new update gets installed.

Assuming that the user agrees on downloading resources over the Internet, Dangerzone should:
[...]

PR #1254, opened 2025-09-29
Update issue-templates

almet:
- Use `podman-machine` commands rather than raw Docker commands
- Add the `--debug` flag to the conversion command

Note: This is to be merged after #1145 hits the `main` branch.

I'm also thinking about adding an utility `dangerzone-machine doctor` command to retrieve the logs in one command, avoiding the need for the users to copy-paste multiple commands. I'm not doing this here to avoid stacking too much things on top of the `embed-podman` effort for now, though.


apyrgio:
Looks great, thanks! One small (but probably important) thing I'd suggest is to remove the ticks around `'json'`, since they are redundant and we've experienced issues in Windows before.

almet:
Updated this, and now targeting the `test/main-icu` branch, let me know if I can merge there!

PR #1255, opened 2025-09-29
Add a deprecation warning for debian bullseye

almet:
Also, this clarifies that we only support the last two stable releases for Debian.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


apyrgio:
Looks good to me. Once you rebase on top of `test/main-icu`, I'll check it out in a test environment as well.

almet:
Cool. Just rebased atop of `test/main-icu`. The warning displays after the container installation and update phase. I think it's correct, but we might want to display it on both screens. WDYT?

apyrgio:
You mean because the conversion widget does not show immediately, right? Well, I'm experimenting with always showing that widget, based on what we discussed, so I think we'll be fine.

almet:
> You mean because the conversion widget does not show immediately, right? 

Exactly this. Okay perfect then!

PR #1257, opened 2025-10-02
Various UI/UX improvements

### LINKED ISSUE #1245: 0.10.0 UX review
Review the overall conversion flow for UX issues.

apyrgio:
In this PR, we have tried to improve the UI/UX of Dangerzone, based on some discussions our team had around that matter. The main changes in this PR are:

* Convert the status bar to a Qt `QStatusBar`, with a solid background color, and remove the info icon. Also add it in our design system.
* Add "Copy to clipboard" button in log window
* Drop the waiting widget from showing on startup, and always show the conversion widget instead.
* Catch startup errors in a better way, and link to a wiki page.
* Better indicate that documents are enqueued until conversion finishes.

Closes #1245

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -226,28 +226,28 @@ class StatusBar(QtWidgets.QStatusBar):
         self.set_status_working("Shutting down")
 
     def handle_task_machine_init(self) -> None:
-        self.set_status_working("Initializing Dangerzone VM")
+        self.set_status_working("Initializing Dangerzone sandbox (creating VM)")
 
     def handle_task_machine_start(self) -> None:
-        self.set_status_working("Starting Dangerzone VM")
+        self.set_status_working("Initializing Dangerzone sandbox (starting VM)")
 
     def handle_task_machine_stop_others(self) -> None:
-        self.set_status_working("Stopping other Podman VMs")
+        self.set_status_working("Initializing Dangerzone sandbox (stopping other VMs)")
 
     def handle_task_machine_stop(self) -> None:
-        self.set_status_working("Stopping Dangerzone VM")
+        self.set_status_working("Stopping Dangerzone sandbox (shutting down VM)")
 
     def handle_task_update_check(self) -> None:
         self.set_status_working("Checking for updates")
 
     def handle_task_container_install_local(self) -> None:
-        self.set_status_working("Installing local Dangerzone sandbox")
+        self.set_status_working("Installing Dangerzone sandbox (from local archive)")
 
     def handle_task_container_install_remote(self) -> None:
-        self.set_status_working("Downloading and installing Dangerzone sandbox")
+        self.set_status_working("Downloading Dangerzone sandbox (from trusted remote)")
 
     def handle_task_container_stop(self) -> None:
-        self.set_status_working("Stopping Dangerzone sandbox")
+        self.set_status_working("Stopping Dangerzone sandbox (clearing jobs)")
 
     def handle_startup_error(self) -> None:
         self.set_status_error("Startup failed")
--- a/dangerzone/gui/log_window.py
+++ b/dangerzone/gui/log_window.py
@@ -84,48 +84,52 @@ class LogWindow(QtWidgets.QDialog):
 
     def handle_task_machine_init(self) -> None:
         self.label.setText(
-            "Initializing the Dangerzone VM<br>This might take a few minutes…"
+            "Initializing Dangerzone sandbox (creating VM)<br>This might take a few minutes…"
         )
 
     def handle_task_machine_start(self) -> None:
-        self.label.setText("Starting the Dangerzone VM…")
+        self.label.setText("Initializing Dangerzone sandbox (starting VM)…")
 
     def handle_task_machine_stop_others(self) -> None:
-        self.label.setText("Stopping other Podman VMs…")
+        self.label.setText("Initializing Dangerzone sandbox (stopping other VMs)…")
 
     def handle_task_update_check(self) -> None:
         self.label.setText("Checking for updates…")
 
     def handle_task_container_install_local(self) -> None:
         self.label.setText(
-            "Installing local Dangerzone sandbox<br>This might take a few minutes…"
+            "Installing Dangerzone sandbox (from local archive)<br>"
+            "This might take a few minutes…"
         )
 
     def handle_task_container_install_remote(self) -> None:
         self.label.setText(
-            "Downloading and installing Dangerzone sandbox<br>This might take a few minutes…"
+            "Downloading Dangerzone sandbox (from trusted remote)<br>"
+            "This might take a few minutes…"
         )
 
     def handle_task_container_stop(self) -> None:
-        self.label.setText("Stopping the Dangerzone sandbox…")
+        self.label.setText("Stopping the Dangerzone sandbox (clearing jobs)…")
 
     def handle_task_machine_init_failed(self, error: str) -> None:
-        self.label.setText("Initializing the Dangerzone VM… failed")
+        self.label.setText("Initializing Dangerzone sandbox (creating VM)… failed")
 
     def handle_task_machine_start_failed(self, error: str) -> None:
-        self.label.setText("Starting the Dangerzone VM… failed")
+        self.label.setText("Initializing Dangerzone sandbox (starting VM)… failed")
 
     def handle_task_machine_stop_others_failed(self, error: str) -> None:
-        self.label.setText("Stopping other Podman VMs… failed")
+        self.label.setText(
+            "Initializing Dangerzone sandbox (stopping other VMs)… failed"
+        )
 
     def handle_task_machine_stop(self) -> None:
-        self.label.setText("Stopping Dangerzone VM…")
+        self.label.setText("Stopping Dangerzone sandbox (shutting down VM)…")
 
     def handle_task_update_check_failed(self, error: str) -> None:
         self.label.setText("Checking for updates… failed")
 
     def handle_task_container_install_failed(self, error: str) -> None:
-        self.label.setText("Installing the Dangerzone sandbox… failed")
+        self.label.setText("Installing Dangerzone sandbox… failed")
 
     def handle_startup_success(self) -> None:
         # We want to hide the text if there's nothing of importance to say to the user,
--- a/tests/gui/test_main_window.py
+++ b/tests/gui/test_main_window.py
@@ -670,7 +670,7 @@ def test_installation_failure_exception(
     assert mock_installer.call_count == 1
     assert window.status_bar.property("style") == "status-error"
     assert window.status_bar.message.text() == "Startup failed"
-    assert window.log_window.label.text() == "Installing the Dangerzone sandbox… failed"
+    assert window.log_window.label.text() == "Installing Dangerzone sandbox… failed"
     assert "Error during install" in window.log_window.traceback_widget.toPlainText()
 
 


dangerzone/gui/main_window.py
```diff
@@ -694,24 +662,33 @@ def closeEvent(self, e: QtGui.QCloseEvent) -> None:
...
-            "Welcome! I'm afraid you gonna have to wait a bit.<br>"
-            "Check the bottom-right corner for a progress report"
-        )
+    def handle_error(self, msg: str) -> None:
+        self.header.setText("<p>Dangerzone encountered an error during startup:<p>")
+        self.header.setStyleSheet("QLabel { font-size: 20px; }")
+        self.traceback_widget.process_output(msg)
+        self.footer.setText("""\
+            <p>
+              <a href="https://github.com/freedomofpress/dangerzone/wiki/Reporting-an-issue">
```
almet, line 712: Nice to have some doc! Instead of a wiki page, I think it would be better to point to a page in our website instead (and potentially update the link to open an issue to this intermediate page)

dangerzone/gui/updater.py
```diff
@@ -26,14 +26,15 @@
 
 <p>This is recommended in most settings. For alternative methods to keep
 Dangerzone up-to-date and secure (e.g., in a networkless environment),
-see our documentation.</p>
+see <a href=https://github.com/freedomofpress/dangerzone/wiki/Updates>our
```
almet, line 29: I do think this might not be up to date, and as we're not really updating the wiki, linking to the related .md files, until we decide on a better fate for the doc (cf #903)

share/spinner.svg
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" style="margin: auto; background: rgb(255, 255, 255); display: block; shape-rendering: auto;" width="200px" height="200px" viewBox="0 0 100 100" preserveAspectRatio="xMidYMid">
```
almet, line 2: Ah, thanks, I did the same thing 👍

tests/gui/test_status_bar.py
```diff
@@ -26,31 +26,27 @@ def status_bar(qtbot: QtBot, mocker: MockerFixture) -> StatusBar:
...
+    assert status_bar.property("style") == "status-success"
 
 
 def test_set_status_working(status_bar: StatusBar) -> None:
     status_bar.set_status_working("Something is happening")
     assert status_bar.message.text() == "Something is happening"
     assert not status_bar.spinner.isHidden()
-    assert not status_bar.info_icon.isHidden()
-    assert status_bar.message.property("style") == "status-attention"
+    assert status_bar.property("style") == "status-attention"
```
almet: I think we should name this `-info` instead

dangerzone/gui/main_window.py
apyrgio: Makes sense. For the time being, I've done the same thing as in our wiki page for docs, but we can switch to something else, if time permits.

dangerzone/gui/updater.py
apyrgio: I mostly reinstated the old link, until we figure out how to proceed. It's indeed out of date for ICU though, you're right.

almet: Ah, just found that we were actually already pointing to this doc in our dev docs, so I guess this is the canonical source for now. Let's do it this way then!

dangerzone/gui/main_window.py
```diff
@@ -694,24 +662,33 @@ def closeEvent(self, e: QtGui.QCloseEvent) -> None:
...
+                 Report issue ↗️
```
almet, line 713: ↗️ should not be inside the link. And I believe we could just get rid of it entirely, or use it consistently for all our links.

apyrgio: Damn, you're right, Wikipedia has it outside the link. The reason I added it is because it felt weird to me having this underlined "Report issue" message there. I wanted to showcase that it's an external link, which is why I added this character. Do you have a better indicator here (and I agree about the consistency part).

almet:
Thanks, here are some tweaks to the UI which I think make it look a bit better in terms of color  (even though I'm not a fan of these unfortunately).

```diff
diff --git a/assets/design-system.html b/assets/design-system.html
index 5d52d241..e196442d 100644
--- a/assets/design-system.html
+++ b/assets/design-system.html
@@ -126,10 +126,11 @@
         .status-bar {
             padding: 8px 15px;
             font-size: 14px;
-            display: flex;
             align-items: center;
+            text-align: right;
             gap: 8px;
-            border-radius: 4px;
+            border-top-left-radius: 3px;
+            border-top-right-radius: 3px;
             margin: 10px 0;
         }
 
 
 class MainWindow(QtWidgets.QMainWindow):
diff --git a/share/dangerzone.css b/share/dangerzone.css
index 022c3c79..84a7dcba 100644
--- a/share/dangerzone.css
+++ b/share/dangerzone.css
@@ -80,31 +80,44 @@ MainWindow[OSColorMode="dark"] QLabel[style="warning"] {
 /*
  * Light mode assumes background-color: #ECECEC
  */
-MainWindow[OSColorMode="light"] QStatusBar[style="status-error"] {
-    background-color: #FFB3B3;
+ 
+MainWindow[OSColorMode="light"] QStatusBar[style="status-success"], MainWindow[OSColorMode="light"] QStatusBar[style="status-success"] QLabel {
+    background-color: #D4EDDA;
+    color: #155724;
 }
 
-MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"] {
-    background-color: #FFDDC1;
+MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"], MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"] QLabel {
+    background-color: #D1ECF1;
+    color: #0C5460;
+    border-top: 1px solid #BEE5EB;
 }
-
-MainWindow[OSColorMode="light"] QStatusBar[style="status-success"] {
-    border-top: 0px;
+    
+MainWindow[OSColorMode="light"] QStatusBar[style="status-error"], MainWindow[OSColorMode="light"] QStatusBar[style="status-error"] QLabel {
+    background-color: #F8D7DA;
+    color: #721C24;
+    border-top: 1px solid #F5C6CB;
 }
 
 /*
  * Dark mode assumes background-color: #323232
  */
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"] {
-    background-color: #FF9494;
+
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"] QLabel {
+    background-color: #1A4D2E;
+    color: #86EE72;
+    border-top: 1px solid #2D6B3F;
 }
 
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"] {
-    background-color: #FFD18F;
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"] QLabel {
+    background-color: #1A3A52;
+    color: #9DD9F0;
+    border-top: 1px solid #2B5472;
 }
 
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"] {
-    border-top: 0px;
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"] QLabel {
+    background-color: #4D1A1A;
+    color: #FF9494;
+    border-top: 1px solid #722D2D;
 }
 
 MainWindow QStatusBar {

```

dangerzone/gui/main_window.py
almet: Should we make this a button instead?

almet:
When playing a bit about the UI I figured the following:

- When an error arises in the conversion, no error is shown in the status bar, and I find it misleading to have a green status bar when we're dealing with an issue.
- By default, with my color changes, you'll see that "success" is using a green background with nothing to tell, which is... weird. So maybe we should have a neutral status for it as well, or we could add a "Dangerzone is ready" text, like this:

```diff
diff --git a/dangerzone/gui/main_window.py b/dangerzone/gui/main_window.py
index a39c1913..f44e0638 100644
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -235,7 +235,7 @@ class StatusBar(QtWidgets.QStatusBar):
         self.set_status_error("Startup failed")
 
     def handle_startup_success(self) -> None:
-        self.set_status_ok("")
+        self.set_status_ok("Dangerzone is ready")
```

dangerzone/gui/main_window.py
apyrgio: I've thought about this as well. We currently use buttons for actions in the UI, but here we redirect the user to another site. So it's more of a link to me, but if you feel it's confusing, I can change it.

apyrgio:
> Thanks, here are some tweaks to the UI which I think make it look a bit better in terms of color (even though I'm not a fan of these unfortunately).

Thanks for the tweaks. My understanding is that the main changes are the removal of the black border, and the switch of the orange color to blue, when things are in progress.

I'm fine with those changes, and I can add them in our design system as well. My only issue is the green status bar which, to address your last two points, I think we can safely hide it on steady state. There's nothing to convey to the user after all, and we punt the need for showing conversion errors there. Are you ok if I remove just that?

dangerzone/gui/main_window.py
almet: After some thinking, I believe I now understand why this could be useful to some users. The goal is to convey some meaning that this will open in a browser and not in the app, hence doing a network request.

I'm not sure this is clear with the unicode char, but probably it's better than without anything!

I've created an issue about this to use it for all our links in #1262

almet:
> Thanks for the tweaks. My understanding is that the main changes are the removal of the black border, and the switch of the orange color to blue, when things are in progress.

The removal of the black border was mainly to avoid it showing up when there were nothing to report.

> I can add them in our design system as well.

Hmm, my last diff was actually missing the meat of it, I exported it wrongly. Here is it, which contains the changes to the design system as well:

```diff
diff --git a/assets/design-system.html b/assets/design-system.html
index 614b0ccc..e196442d 100644
--- a/assets/design-system.html
+++ b/assets/design-system.html
@@ -3,7 +3,7 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Dangerzone Color Showcase</title>
+    <title>Dangerzone Design System</title>
     <style>
         body {
             font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
@@ -93,26 +93,9 @@
             max-width: 100%;
         }
 
-        .status-bar {
-            text-align: right;
-            padding: 3px 5px;
-            height: 15px;
-            line-height: 15px;
-            margin-bottom: 5px;
-            font-weight: bold;
-            border-radius: 3px;
-            font-size: 12px;
-        }
-
         /* Light Mode */
         .light-mode { background-color: #ececec; color: black; }
         .light-mode .state-section { border-color: #c8c8c8; }
-        .light-mode .status-error { color: #9D0000; font-weight: bold; }
-        .light-mode .status-attention { color: #7A3800; font-weight: bold; }
-        .light-mode .status-success { color: #005000; font-weight: bold; }
-        .light-mode .status-attention-bg { background-color: #FFDDC1; color: black; border: 1px solid black; }
-        .light-mode .status-error-bg { background-color: #FFB3B3; color: black; border: 1px solid black; }
-        .light-mode .status-success-bg { background-color: #C8E6C9; color: black; border: 1px solid black; }
         .light-mode .warning { background-color: #FFF3CD; color: #856404; border: 1px solid #FFEEBA; border-radius: 4px; padding: 10px; margin: 10px; }
         .light-mode .safe-extension-filename { background: white; color: grey; border-color: #c8c8c8; }
         .light-mode .line-edit { border-color: #c8c8c8; background-color: white; color: black; }
@@ -125,12 +108,6 @@
         /* Dark Mode */
         .dark-mode { background-color: #323232; color: white; }
         .dark-mode .state-section { border-color: #555; }
-        .dark-mode .status-error { color: #FF9494; font-weight: bold; }
-        .dark-mode .status-attention { color: #FFD18F; font-weight: bold; }
-        .dark-mode .status-success { color: #86EE72; font-weight: bold; }
-        .dark-mode .status-attention-bg { background-color: #FFD18F; color: black; }
-        .dark-mode .status-error-bg { background-color: #FF9494; color: black; }
-        .dark-mode .status-success-bg { background-color: #86EE72; color: black; }
         .dark-mode .warning { background-color: #332D00; color: #FFD970; border: 1px solid #665A00; border-radius: 4px; padding: 10px; margin: 10px; }
         .dark-mode .safe-extension-filename { background: black; color: grey; border-color: #c8c8c8; }
         .dark-mode .line-edit { border-color: #c8c8c8; background-color: #222; color: white; }
@@ -144,6 +121,54 @@
         .safe-extension-filename { border-width: 1px; border-right-width: 0px; border-style: solid; padding: 5px 0px 5px 5px; }
         .docs-selection { font-size: 18px; }
         .version { font-size: 20px; padding-bottom: 5px; }
+
+        /* Bottom status bar */
+        .status-bar {
+            padding: 8px 15px;
+            font-size: 14px;
+            align-items: center;
+            text-align: right;
+            gap: 8px;
+            border-top-left-radius: 3px;
+            border-top-right-radius: 3px;
+            margin: 10px 0;
+        }
+
+ 
+
+        /* Light mode status bar colors */
+        .light-mode .status-bar.status-bar-success {
+            background-color: #D4EDDA;
+            color: #155724;
+            border-top: 1px solid #C3E6CB;
+        }
+        .light-mode .status-bar.status-bar-info {
+            background-color: #D1ECF1;
+            color: #0C5460;
+            border-top: 1px solid #BEE5EB;
+        }
+        .light-mode .status-bar.status-bar-error {
+            background-color: #F8D7DA;
+            color: #721C24;
+            border-top: 1px solid #F5C6CB;
+        }
+
+        /* Dark mode status bar colors */
+        .dark-mode .status-bar.status-bar-success {
+            background-color: #1A4D2E;
+            color: #86EE72;
+            border-top: 1px solid #2D6B3F;
+        }
+        .dark-mode .status-bar.status-bar-info {
+            background-color: #1A3A52;
+            color: #9DD9F0;
+            border-top: 1px solid #2B5472;
+        }
+        .dark-mode .status-bar.status-bar-error {
+            background-color: #4D1A1A;
+            color: #FF9494;
+            border-top: 1px solid #722D2D;
+        }
     </style>
 </head>
 <body>
@@ -155,9 +180,18 @@
 
             <div class="state-section">
                 <h3>Status Bar</h3>
-                <div class="status-bar status-attention-bg">Working: Checking for updates...</div>
-                <div class="status-bar status-error-bg">Error: Startup failed</div>
-                <div class="status-bar status-success-bg">Success: Ready</div>
+                <div class="status-bar status-bar-success">
+                    <img src="../share/status_safe.png" width="16" height="16" />
+                    <span>All documents converted successfully</span>
+                </div>
+                <div class="status-bar status-bar-info">
+                    <img src="../share/spinner.svg" width="16" height="16" />
+                    <span>Converting documents...</span>
+                </div>
+                <div class="status-bar status-bar-error">
+                    <img src="../share/status_failed.png" width="16" height="16" />
+                    <span>Error: Conversion failed for 1 document</span>
+                </div>
             </div>
 
             <div class="state-section">
@@ -176,6 +210,7 @@
                 </div>
             </div>
 
+
             <div class="state-section">
                 <h3>Document Selection</h3>
                 <div class="doc-selection-drop-frame">
@@ -228,9 +263,18 @@
 
             <div class="state-section">
                 <h3>Status Bar</h3>
-                <div class="status-bar status-attention-bg">Working: Checking for updates...</div>
-                <div class="status-bar status-error-bg">Error: Startup failed</div>
-                <div class="status-bar status-success-bg">Success: Ready</div>
+                <div class="status-bar status-bar-success">
+                    <img src="../share/status_safe.png" width="16" height="16" />
+                    <span>All documents converted successfully</span>
+                </div>
+                <div class="status-bar status-bar-info">
+                    <img src="../share/spinner-dark.svg" width="16" height="16" />
+                    <span>Converting documents...</span>
+                </div>
+                <div class="status-bar status-bar-error">
+                    <img src="../share/status_failed.png" width="16" height="16" />
+                    <span>Error: Conversion failed for 1 document</span>
+                </div>
             </div>
 
             <div class="state-section">
diff --git a/share/dangerzone.css b/share/dangerzone.css
index 022c3c79..84a7dcba 100644
--- a/share/dangerzone.css
+++ b/share/dangerzone.css
@@ -80,31 +80,44 @@ MainWindow[OSColorMode="dark"] QLabel[style="warning"] {
 /*
  * Light mode assumes background-color: #ECECEC
  */
-MainWindow[OSColorMode="light"] QStatusBar[style="status-error"] {
-    background-color: #FFB3B3;
+ 
+MainWindow[OSColorMode="light"] QStatusBar[style="status-success"], MainWindow[OSColorMode="light"] QStatusBar[style="status-success"] QLabel {
+    background-color: #D4EDDA;
+    color: #155724;
 }
 
-MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"] {
-    background-color: #FFDDC1;
+MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"], MainWindow[OSColorMode="light"] QStatusBar[style="status-attention"] QLabel {
+    background-color: #D1ECF1;
+    color: #0C5460;
+    border-top: 1px solid #BEE5EB;
 }
-
-MainWindow[OSColorMode="light"] QStatusBar[style="status-success"] {
-    border-top: 0px;
+    
+MainWindow[OSColorMode="light"] QStatusBar[style="status-error"], MainWindow[OSColorMode="light"] QStatusBar[style="status-error"] QLabel {
+    background-color: #F8D7DA;
+    color: #721C24;
+    border-top: 1px solid #F5C6CB;
 }
 
 /*
  * Dark mode assumes background-color: #323232
  */
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"] {
-    background-color: #FF9494;
+
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"] QLabel {
+    background-color: #1A4D2E;
+    color: #86EE72;
+    border-top: 1px solid #2D6B3F;
 }
 
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"] {
-    background-color: #FFD18F;
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-attention"] QLabel {
+    background-color: #1A3A52;
+    color: #9DD9F0;
+    border-top: 1px solid #2B5472;
 }
 
-MainWindow[OSColorMode="dark"] QStatusBar[style="status-success"] {
-    border-top: 0px;
+MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"], MainWindow[OSColorMode="dark"] QStatusBar[style="status-error"] QLabel {
+    background-color: #4D1A1A;
+    color: #FF9494;
+    border-top: 1px solid #722D2D;
 }
 
 MainWindow QStatusBar {

```

> My only issue is the green status bar which, to address your last two points, I think we can safely hide it on steady state. There's nothing to convey to the user after all, and we punt the need for showing conversion errors there. Are you ok if I remove just that?

Hiding it makes sense, to avoid having two different (and potentially conflicting) information shown to the user. I was initially thinking of having the status bar report if any error occurred, but that's not currently how it functions.

As such, we can remove it to avoid inconsistencies as the one I mentioned in a previous post. My understanding is that we'll never have a "success" state shown to the user, and that green will become "transparent". 

Correct?

apyrgio:
Ah, much better now. Thanks for the changes in the design system as well.

> My understanding is that we'll never have a "success" state shown to the user, and that green will become "transparent".

Yes, and I'll add a fix for that. I'll also try to add the proper text color in the status bar, because what I'm seeing in the UI doesn't match the color in the design system (I'm seeing black instead of the color you chose).

dangerzone/gui/main_window.py
apyrgio: > The goal is to convey some meaning that this will open in a browser and not in the app, hence doing a network request.

Yep!

> I'm not sure this is clear with the unicode char, but probably it's better than without anything!

I like the CSS trick most sites use, but I haven't tried it out here for... obvious reasons.

Thanks for the issue, I assume we can resolve this conversation then and keep the issue open?

tests/gui/test_status_bar.py
apyrgio: Done in 631fa1a76ed829525dbd48ebf41c0a16fcf04377

almet:
Here are a couple of comments when trying this locally:

- In the design system, you added my changes, but I believe that since we never show the green status, maybe it should go away from there, and that the error shouldn't reflect an error with the conversion, but related to something else. Also, I'm not sure if the error is currently using the cross icon that's in there :thinking: ?
- When a conversion error occurs, it might be useful to open the logs window when double clicking on it, rather than having to go in the menu. At least that's what I intuitively tried without success.
- In the log window, it mentions "Dangerzone is ready", even in a failure mode, and I find it odd (I'm thinking: "well, it's ready, but it failed"). I'm not sure of the added value of the "DZ is ready" text there, and/or we could switch to something that says there is an error when it so happen.

apyrgio:
> In the design system, you added my changes, but I believe that since we never show the green status, maybe it should go away from there,

You have a point, I don't like this discrepancy as well.

> and that the error shouldn't reflect an error with the conversion, but related to something else. 

I can restore the previous text, which was about startup issues.

> Also, I'm not sure if the error is currently using the cross icon that's in there 🤔 ?

Correct, I'll remove it.

> When a conversion error occurs, it might be useful to open the logs window when double clicking on it, rather than having to go in the menu. At least that's what I intuitively tried without success.

Never thought about it. Ok, I can add it.

> In the log window, it mentions "Dangerzone is ready", even in a failure mode, and I find it odd (I'm thinking: "well, it's ready, but it failed"). I'm not sure of the added value of the "DZ is ready" text there, and/or we could switch to something that says there is an error when it so happen.

We should handle it similarly to the status bar then, i.e., clear the message, if there's nothing to show to the user.

almet:
Some more comments (currently trying on OSX):

- When downloading an image, it's actually saying "installing dangerzone sandbox", but I believe we could make it say "downloading new sandbox version" or something that differs from actually "installing".
- When waiting, I've also tried to double click on the text, and was expecting the logs window to open, but it wasn't the case
- I also found some stuff not related to this code, and will open specific issues about it.

apyrgio:
> When downloading an image, it's actually saying "installing dangerzone sandbox", but I believe we could make it say "downloading new sandbox version" or something that differs from actually "installing".

Good point, took a stab at it in 0f5cd2cc.

> When waiting, I've also tried to double click on the text, and was expecting the logs window to open, but it wasn't the case

Oh, so you wanted to see the logs regardless of the document's conversion status, not just for errors, right? That's easy, but I'm worried that the log window will show when the user does errand clicks in the UI (bound to happen). Not only that, but they will not see any conversion logs for that document, so they will get minimal information :-/

almet:
> Oh, so you wanted to see the logs regardless of the document's conversion status, not just for errors, right?

Yep, because it was a bit long, I wanted to know what was going on and clicked on the text.

> That's easy, but I'm worried that the log window will show when the user does errand clicks in the UI (bound to happen). Not only that, but they will not see any conversion logs for that document, so they will get minimal information :-/

Just to make sure we're talking about the same thing, I'm referring to the text in the status bar, while waiting for the sandbox to be downloaded (and not to the text of the conversion window)

In this case, seeing whats going on under the hood while waiting and wondering if Dangerzone is blocked feels reassuring to me. J

almet:
While using this branch and doing some tests, I remembered a past discussion about how we want to refer to the sandbox in a kind of a unique way. As such we might want to update our language to be more consistent:

I'm thinking about these different steps:

- Initializing Dangerzone sandbox (creating VM)
- Initializing Dangerzone sandbox (starting VM)
- Installing Dangerzone sandbox (local archive)
- Downloading Dangerzone sandbox (from trusted remote)
[...]

The main idea here is to always use "the sandbox", instead of mentioning "the VM".

What do you think?

apyrgio:
> Just to make sure we're talking about the same thing, I'm referring to the text in the status bar, while waiting for the sandbox to be downloaded (and not to the text of the conversion window)

Ah, ok. So you want to click on the text in the status bar and see the logs, basically as if the info icon was there. I can add it.

apyrgio:
>     Initializing Dangerzone sandbox (creating VM)
>     Initializing Dangerzone sandbox (starting VM)

Do you perhaps mean "Starting" instead of "Initializing" here?

>     Installing Dangerzone sandbox (local archive)
>     Downloading Dangerzone sandbox (from trusted remote)
>     [...]
> 
> The main idea here is to always use "the sandbox", instead of mentioning "the VM".

Hm, I think we're muddying the terms here, to the point that they don't reflect what's going on, unless they look at the code. We can maybe do away with:

* Initializing Dangerzone machine
* Starting Dangerzone machine
* Installing Dangerzone sandbox (local)
* Downloading Dangerzone sandbox (remote)

That is, exchange "VM" with "machine". Still kind of alludes to VMs, without the acronym. What do you think?

almet:
> > ```
> > Initializing Dangerzone sandbox (creating VM)
> > Initializing Dangerzone sandbox (starting VM)
> > ```
> 
> Do you perhaps mean "Starting" instead of "Initializing" here?

Nope, I was thinking that the VM management is part of the initialization of the sandbox.

> Hm, I think we're muddying the terms here, to the point that they don't reflect what's going on, unless they look at the code. We can maybe do away with:
> 
>     * Initializing Dangerzone machine
>     * Starting Dangerzone machine
>     * Installing Dangerzone sandbox (local)
>     * Downloading Dangerzone sandbox (remote) 

As I mentioned, I believe we should reduce the complexity and always refer to "the sandbox" in place of machine and container. We can add some more info inside parenthesis, or in the logs.

If that's hard to discuss in these comments, we can also discuss this live instead.

apyrgio:
Oh sorry, I thought the parentheses were there for the discussion, and not something that would be added in the status bar. Yeah, let's discuss live so that we can wrap this up.

apyrgio:
Heads up: I have rebased and squashed some fixup commits. Also, I've added the following new commits:

* 1faf7d3d - gui: Reword our status messsages (for the rewording we were talking about)
* fbb46718 - gui: Open logs when clicking on a status message (in order to show the log window when clicking on a message in the status bar)

almet:
Perfect, to me this is ready to be squashed and merged!

PR #1261, opened 2025-10-06
Add a script to release a new container image

### LINKED ISSUE #1064: Add a script to attest+sign container images
We should provide a script to do the whole roundtrip:

- Verify the reproducibility of the image, comparing it with the one published by our CI jobs (see #1074)
- Verify that attestations match our expectations (see #1063)
- Sign the container image with a hardware key and publish the signature to the container registry
- Tag the container image as `latest`

almet:
This script glues together multiple other scripts to ease the general process of attesting, reproducing, signing and tagging a new container image.

Fixes #1064


almet:
Some notes about this PR:

- When trying to reproduce the image on my local copy of dangerzone it's failing and I'm not sure what to do to fix it (it's not finding the proper commit hash, it seems, but I'm not sure as to why). I think it's not important to do that right now, and that this issue will go away as soon as we target the main dangerzone repo and not my own fork)
- It requires some changes on the ghcr-signer, which was using some path-relative files (not relative to the ghcr-signer itself). I'll add some commits in https://github.com/freedomofpress/ghcr-signer/pull/7 to address this.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+
+def run(cmd, check=True, capture_output=False):
+    """Run a command and optionally capture output."""
+    click.echo(f"Running: {' '.join(cmd)}")
+    result = subprocess.run(cmd, check=check, capture_output=capture_output, text=True)
+    return result
+
+
+@click.command()
+@click.option("--commit", default=None)
```
apyrgio: Let's add a description here, and in the rest of the CLI options, to show how they will be used. We can say for instance that by default we will use the HEAD of the Git branch.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+@click.option(
+    "--ghcr-signer-path",
+    type=click.Path(exists=True),
+    required=True,
+    help="Path to the ghcr-signer repository",
+    default="../ghcr-signer",
+)
+@click.option("--repository", default="freedomofpress/dangerzone")
+@click.option("--branch", default="main")
+def attest_release_sign(commit, ghcr_signer_path, repository, branch):
```
apyrgio: One option that is present in `attest-provenance` but is missing here is the `--workflow` option.

dev_scripts/release-container-image.py
apyrgio: Also, specifically for this CLI option, we need to verify that the user passes the full SHA1 commit. Else, our `attest-provenance` script will fail.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+    """Get platform-specific full_image digests."""
+    click.echo(f"\n📋 Getting platform-specific digests for: {full_image}\n")
+
+    try:
+        result = run(
+            ["crane", "manifest", full_image],
+            capture_output=True,
+        )
+        manifest = json.loads(result.stdout)
+        click.echo("✅ Platform digests retrieved\n")
```
apyrgio: Can we perhaps mention here what are the platform digests that we have retrieved?

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+        commit = result.stdout.strip()
+
+    full_image = get_candidate_image(commit)
+    attest_provenance(full_image, commit, repository, branch)
+    digests = get_platform_digests(full_image)
+    debian_archive_date = extract_debian_archive_date()
+    for platform in ["linux/amd64", "linux/arm64"]:
+        pass  # reproduce_image(full_image, platform, debian_archive_date, digests[platform])
+
+    sign_image(full_image, ghcr_signer_path)
```
apyrgio: I would like the signing part to be optional, ideally, just in case we want to verify in one machine and sign in another.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+    image_name = IMAGE_NAME_PATH.read_text().strip()
+
+    click.echo(f"\n📦 Looking for images for commit: {short_commit}")
+    click.echo(f"   Repository: {image_name}\n")
+
+    # Get the latest container image with this commit
+    try:
+        result = run(
+            [
+                "crane",
```
apyrgio, line 221: I think we need a preflight check at the very start, where we will ensure that `uv`, `crane`, `git` and `poetry` are in the `PATH`.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+PROJECT_ROOT = Path(__file__).parent.parent
+IMAGE_NAME_PATH = PROJECT_ROOT / "share" / "image-name.txt"
+DOCKER_ENV_PATH = PROJECT_ROOT / "Dockerfile.env"
+
+
+def extract_debian_archive_date():
+    with DOCKER_ENV_PATH.open() as f:
+        content = f.read()
+
+    match = re.search(r"DEBIAN_ARCHIVE_DATE=(\d+)", content)
```
apyrgio: I'm afraid we can't do that. If you recall, we bump the `DEBIAN_ARCHIVE_DATE` in our CI runners every night, so it's not necessarily the same as the stored one. So, we have to parse this  value from the image tag, as I mentioned in the doc.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+        return {
+            f"{m['platform']['os']}/{m['platform']['architecture']}": m["digest"]
+            for m in manifest["manifests"]
+        }
+    except subprocess.CalledProcessError as e:
+        click.echo(f"❌ Error getting platform digests: {e}")
+        sys.exit(1)
+
+
+def reproduce_image(parent_image, platform, debian_archive_date, digest):
```
apyrgio: We have an issue here. This function will work only if two conditions apply:
1. The image we want to reproduce comes from the tip of the branch we're in right now.
2. The repo is clean, with no changes that affect the container image.

This _may_ be the case only when we release the container image, but not a few days later. In order to do it properly, we have to:
1. Git clone our Dangerzone repo to a temporary directory.
2. Checkout to the specific commit.
3. Run the `reproduce-image` command

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+    """Reproduce the image for a specific platform."""
+    # Format image with digest
+    image_base = parent_image.split("@")[0]
+    full_image = f"{image_base}@{digest}"
+
+    click.echo(f"\n🔄 Reproducing image for platform: {platform}")
+    click.echo(f"   Parent Image: {parent_image}")
+    click.echo(f"   Digest: {digest}")
+    click.echo(f"   Debian archive date: {debian_archive_date}")
+    click.echo(f"   Full image: {full_image}\n")
```
apyrgio: Minor, I would use slightly different terms:

Parent image -> Root manifest
Digest -> Platform image digest
Full image -> Platform image

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+import json
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import click
+
+PROJECT_ROOT = Path(__file__).parent.parent
+IMAGE_NAME_PATH = PROJECT_ROOT / "share" / "image-name.txt"
```
apyrgio, line 23: I would let the user specify their own image name, just in case we are at a point where we move from `v1` to `v2` or something, and we want to sign something older.

Also, it helps now for testing, and maybe in the future as well.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,246 @@
...
+            capture_output=True,
+        )
+        commit = result.stdout.strip()
+
+    full_image = get_candidate_image(commit)
+    attest_provenance(full_image, commit, repository, branch)
+    digests = get_platform_digests(full_image)
+    debian_archive_date = extract_debian_archive_date()
+    for platform in ["linux/amd64", "linux/arm64"]:
+        pass  # reproduce_image(full_image, platform, debian_archive_date, digests[platform])
```
apyrgio: Reminder to uncomment this before merging. Also, kind of an important thing: this step takes quite a while in most cases. We should probably have a flag that can bypass this step, just in case we have performed it already, provided that this flag's name is very strict (e.g, `--yes-i-have-reproduced <digest1>,<digest2>`)

dev_scripts/release-container-image.py
apyrgio: Actually, I think we should detect if we collected anything other than 2 platform digests, and fail.

dev_scripts/release-container-image.py
almet: Thanks for all the comments. I've implemented them in commit 3068705495f95b95f4d8d7bc4e23410be6cf95b9, let me know if that suits you :-)

dev_scripts/release-container-image.py
almet: Just did in 48757192388ad48578d40194ba66f38ebe5de715

almet:
Just triggered the macOSx86 job which was failing with:

> Stderr: b'Cannot connect to Podman. Please verify your connection to the Linux system using `podman system connection list`, or try `podman machine init` and `podman machine start` to manage a new Linux VM\nError: unable to connect to Podman socket: failed to connect: ssh: handshake failed: EOF\n'

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,368 @@
...
+
+    try:
+        # Run the reproduce-image command from the cloned repository
+        click.echo("🏗️  Running reproduce-image command...")
+        run(
+            [
+                "poetry",
+                "run",
+                "python",
+                f"{temp_dir}/dev_scripts/reproduce-image.py",
```
apyrgio, line 317: This doesn't seem to work, I'm afraid, because the underlying script uses the current working directory.

```
Running: poetry run python /tmp/dangerzone-reproduce-u9d6w7ie/dev_scripts/reproduce-image.py --debian-archive-date 20251008 --platform linux/arm64 ghcr.io/freedomofpress/dangerzone/v1@sha256:c6826c6a1622bdbb22214d26dba91bbc6a88eff05e32d66c22416a1ce7459018
2025-10-08 21:56:04 - INFO - Building container image      
2025-10-08 21:56:04 - DEBUG - Running command: python3 ./install/common/build-image.py --platform linux/arm64 --runtime podman --debian-archive-date 20251008
```

In general, the best way to circumvent this is to set `cwd=temp_dir` in `subprocess.run()`.

dev_scripts/release-container-image.py
```diff
@@ -0,0 +1,368 @@
...
+    click.echo(f"\n🔄 Reproducing image for platform: {platform}")
+    click.echo(f"   Root manifest: {root_manifest}")
+    click.echo(f"   Platform image digest: {platform_image_digest}")
+    click.echo(f"   Debian archive date: {debian_archive_date}")
+    click.echo(f"   Platform image: {platform_image}")
+    click.echo(f"   Repository path: {temp_dir}\n")
+
+    try:
+        # Run the reproduce-image command from the cloned repository
+        click.echo("🏗️  Running reproduce-image command...")
```
apyrgio, line 311: I'm sometimes getting garbled text when printing this (or similar lines):

```
🏗️  Runnvng reproduce-im/ge command...                                          e                                                                                                       e     d          d             0       l              w
```

It may be my terminal though. Do you get anything similar?

almet: Hmm, I'm not. What's the terminal you're using, so I can reproduce?

dev_scripts/release-container-image.py
almet: Good call, did so in 8eef463436c4780a7ca0e2560be540e5ef437987

almet:
There were an error (I can't find back the discussion unfortunately) due to the ghcr-signer not returning the proper return code. This should be fixed with https://github.com/freedomofpress/ghcr-signer/pull/10

dev_scripts/release-container-image.py
apyrgio: Gnome Terminal + Tmux. I wonder what's going on, since this hasn't happened to me before. It _could_ be that the ANSI escape sequences of Buildkit (from the `reproduce-image.py` script) are somehow messing up things, since I think I don't see the same issue in lines before that.

almet: Unfortunately, I haven't been able to reproduce such an issue with gnome-terminal and tmux on my machine.
Is it possible it's tied to something else? Can you try to run `reset` before running the command to check you still have the issue? Thanks!

Let me know if I can do anything to help.

apyrgio: Ok, looking into it, it seems that this is a Tmux thing, which happens while the build command runs. The tail end of the terminal seems fine, but if you use Tmux's scroll buffer to scroll up, you get those corrupted characters.

Most likely this unrelated with this PR, so let's resume...

apyrgio:
Feel free to merge, thanks!

PR #1268, opened 2025-10-07
Do not use already existing Podman credentials

### LINKED ISSUE #1266: Handle already logged-in users properly
Currently, if a user has logged-in to ghcr.io with either docker or podman in the past, on macOS a prompt will be shown to them, when trying to access their password manager. No matter what they enter after this prompt, the `podman pull` command the comes after it will fail, because github expects this call to be made unauthenticated.

- It's possible to retrieve the current user with `podman login --get-login ghcr.io`.
- We can overwrite the current logged-in user by using `REGISTRY_AUTH_FILE`
- Logging out manually with `podman logout ghcr.io`or `docker logout ghcr.io` solves the issue.

almet:
Two things to note:

- I haven't found how to reproduce exactly the error I've been seeing yesterday on this machine, that was solved by doing `podman logout ghcr.io`
- This seem a simple-enough fix, that was able to avoid using some broken credentials on my machine, that made the pull fail.

(I believe this) fixes #1266

PS: I've added the "no changelog" label because this refers to stuff not yet released, and as such I don't believe they need a specific entry. Let me know if you feel differently.

### DIFF
--- a/dangerzone/container_utils.py
+++ b/dangerzone/container_utils.py
@@ -421,3 +421,16 @@ def get_local_image_digest(image: Optional[str] = None) -> str:
         )
     image_digest = lines.pop().replace("sha256:", "")
     return image_digest
+
+
+def disable_registry_auth(env: dict) -> None:
+    """Disable registry authentication by setting auth-related environment variables
+    to non-existent files.
+
+    This prevents Podman/Docker from using any existing credentials that might be
+    configured on the system, avoiding potential authentication errors.
+    """
+    if "REGISTRY_AUTH_FILE" not in env:
+        env["REGISTRY_AUTH_FILE"] = "does-not-exist"
+    if "DOCKER_CONFIG" not in env:
+        env["DOCKER_CONFIG"] = "does-not-exist"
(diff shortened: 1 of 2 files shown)


almet:
I've just updated this PR, after finding that the issue was related to `cosign download signatures`.

What do you think?

apyrgio:
There are two things that come to mind now:
1. Is there something a user may have added in `DOCKER_CONFIG` that we may not want to override? This file seems to be read implicitly after all, and maybe some other settings are used from there.
2. This worries me a bit: https://github.com/freedomofpress/dangerzone/actions/runs/18349185396/job/52265615886?pr=1268#step:6:139

If we don't have a clean solution for this, it _may_ be ok to accept that users who have logged in to GHCR and no longer have valid credentials, will have to troubleshoot their installation. Most likely they will be power users after all...

almet:
One way to avoid this is to only deal with the credentials for `cosign  download signature` and not for all podman calls.

almet:
I've just updated this PR to reflect this, only the `cosign download signature` calls are now unauthenticated by default.

Issue #1267, opened 2025-10-06
Output from `podman machine raw` is swallowed by the capture_output mechanism


almet:
See the following bash history as a testimony to that:

```bash
alexis@Carnet ~/d/f/dangerzone (1245-ux)> poetry run ./dev_scripts/dangerzone-machine raw images
# (returns nothing)
alexis@Carnet ~/d/f/dangerzone (1245-ux)> /Users/alexis/dev/fpf/dangerzone/share/vendor/podman/podman --connection dz-internal-0.9.1 --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs images
REPOSITORY                                   TAG         IMAGE ID      CREATED       SIZE
ghcr.io/almet/dangerzone/dangerzone-testing  <none>      47bba9f1df4f  15 hours ago  1.05 GB

alexis@Carnet ~/d/f/dangerzone (1245-ux)> poetry run ./dev_scripts/dangerzone-machine --log-level debug raw images
DEBUG:root:Running: /Users/alexis/dev/fpf/dangerzone/share/vendor/podman/podman --connection dz-internal-0.9.1 --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs images
DEBUG:root:REPOSITORY                                   TAG         IMAGE ID      CREATED       SIZE

DEBUG:root:ghcr.io/almet/dangerzone/dangerzone-testing  <none>      47bba9f1df4f  15 hours ago  1.05 GB

DEBUG:root:Process returncode: 0
```

PR #1269, opened 2025-10-07
Do not swallow podman output

almet:
Output the captured streams to INFO or DEBUG

The standard streams are output with an INFO level so they can be
visible by default, and the name of the command and returncode are
sent to the DEBUG level.

Fixes https://github.com/freedomofpress/dangerzone/issues/1267

### DIFF
--- a/dangerzone/podman/cli.py
+++ b/dangerzone/podman/cli.py
@@ -22,7 +22,10 @@ logger = logging.getLogger(__name__)
 )
 def main(log_level: str) -> None:
     """Manage Dangerzone Podman machines."""
-    logging.basicConfig(level=getattr(logging, log_level.upper()), stream=sys.stderr)
+    logging.basicConfig(
+        level=getattr(logging, log_level.upper()),
+        format="%(message)s",
+    )
 
 
 @main.command()
--- a/dangerzone/capture_output.py
+++ b/dangerzone/capture_output.py
@@ -50,7 +50,7 @@ class PatchedPopen(original_subprocess_popen):
 
         def _consume_pipe(pipe: IOBase) -> None:
             for line in iter(pipe.readline, ""):
-                log.debug(_decode_if_needed(line))
+                log.info(_decode_if_needed(line))
 
         # Create threads to read the stdout and stderr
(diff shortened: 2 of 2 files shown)


dangerzone/podman/cli.py
```diff
@@ -22,7 +22,10 @@
 )
 def main(log_level: str) -> None:
     """Manage Dangerzone Podman machines."""
-    logging.basicConfig(level=getattr(logging, log_level.upper()), stream=sys.stderr)
```
almet, line 25: This removes `stream=sys.stderr`, as I believe it's the default behavior. Let me know if I'm missing anything here.

apyrgio:
If we are to deal with this issue properly (which I agree we should), I think we shouldn't fix this by converting the log calls to info. In the long run, what will help us more is to guard the `capture_output.patch_stdlib()` with a condition, e.g., a check if `argv[0] in ["dangerzone", "dangerzone-cli"]`.

Then, we can patch subprocess only if Dangerzone is called from these specific binaries, and not from `dangerzone-image` / `dangerzone-machine`, or any other import we choose to do in the future (or that some user is already doing...).

almet:
> If we are to deal with this issue properly (which I agree we should), I think we shouldn't fix this by converting the log calls to info. In the long run, what will help us more is to guard the `capture_output.patch_stdlib()` with a condition, e.g., a check if `argv[0] in ["dangerzone", "dangerzone-cli"]`.
> 
> Then, we can patch subprocess only if Dangerzone is called from these specific binaries, and not from `dangerzone-image` / `dangerzone-machine`, or any other import we choose to do in the future (or that some user is already doing...).

Thanks for the feedback. I kind of like having one clear path in general, which is why I went the way of a general fix (I believe we shouldn't swallow the output, in general).

I've implemented your feedback in ff4044969b8888ff361723c87c6eb775577aeb60.

I still believe [♫](https://www.youtube.com/watch?v=LKaXY4IdZ40) that using `INFO` in addition to `DEBUG` makes sense in general to avoid capturing only to debug, and would like to keep this commit.

How does that sound?

apyrgio:
Ok, let's go with logging to INFO, and we can reconsider lowering the verbosity once we see more how this little hack of ours behaves in the real world.

PR #1273, opened 2025-10-08
docs: Add installation instructions for Fedora Atomic

### LINKED ISSUE #1232: Add Fedora Atomic (Silverblue) installation instructions
**What is the feature you think should be a good addition to Dangerzone?**

https://fedoraproject.org/atomic-desktops/ are a slight variation on normal Fedora by using an immutable image for deployment. Notably it doesn't support `dnf`, you need to use rpm-ostree.

Steps I used:
* Download https://packages.freedom.press/yum-tools-prod/dangerzone/dangerzone.repo, save it to `/etc/yum.repos.d`
* `rpm-ostree install dangerzone` (no sudo needed)
* reboot!

And then it worked great.

Notably there was no PGP key confirmation, not sure if it just silently did it or what.

HastD:
Fixes #1232. Also add shell script syntax highlighting for a few code blocks.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -29,6 +29,12 @@ Notes:
 
 ◎ More information about where that points [in the runner-images repository](https://github.com/actions/runner-images/tree/main)
 
+### Note on unsupported Linux distros
+
+`.deb` and `.rpm` packages are provided for supported distributions. Users in **other** Debian-based or Fedora-based distros — that are not listed above — may be able to install Dangerzone through these packages. Unfortunately, Dangerzone is not tested against these distros, and might fail to install, update, run, or be broken in subtle ways.
+
+Please, proceed at your own risks, only if you know what you're doing.
+
 ## MacOS
 
 - Download [Dangerzone 0.9.1 for Mac (Apple Silicon CPU)](https://github.com/freedomofpress/dangerzone/releases/download/v0.9.1/Dangerzone-0.9.1-arm64.dmg)
@@ -203,6 +209,11 @@ After confirming that it matches, type `y` (for yes) and the installation should
 </table>
 
 ### Fedora Atomic (Silverblue, Kinoite, etc.)
+
+> [!WARNING]
+> This distribution is [not officially supported](#Note-on-unsupported-Linux-distros)
+> by the Dangerzone team. Please, proceed at your own risks,
+> only if you know what you're doing.
 
 Type the following commands in a terminal:
 


apyrgio:
Hey, thanks a lot for this contribution! I'm inclined to accept it, but I have a slight reservation, which I wanted to mention in the issue as well (but I completely forgot, sorry Kunal :face_with_open_eyes_and_hand_over_mouth:). Currently, we provide installation instructions for OSes which we [officially support](https://github.com/freedomofpress/dangerzone/blob/main/INSTALL.md#operating-system-support). This means that we include them in our CI, follow their point releases, test Dangerzone manually on them before having a release out, etc.

Fedora Silverblue is not officially supported by us, and adding it in our tests would be an overkill, unfortunately. So, it's possible that these instructions may break in the future, or that Dangerzone in general may break in Fedora Silverblue in the future. With that in mind, would you be ok if we referred to these instructions as "community-provided", or similar, to explain that we don't monitor them, and we'll need help from our users if there's an error?

HastD:
@apyrgio Yes, that makes sense. Should I edit this PR to add that language (and if so, do you have specific language I should use?), or will you make that change yourself?

apyrgio:
Nah, that's cool, I can add that comment and add a changelog entry as well, thanks.

@almet I'd like your two cents on this, since it's not something we've done in the past.

almet:
Hey,

Thanks for the installation instructions @HastD and request for comment @apyrgio.

On one side we want to enable simple instructions like this one; on the other side, we don't want people to use half-supported distributions, as it might mean security issues not being taken care of, making the whole point of Dangerzone moot.

The middle-ground is hard to find here (if we're to find any).

What about doing the following:

- Update [the OS support section](https://github.com/freedomofpress/dangerzone/blob/main/INSTALL.md#operating-system-support), adding a "community support" subsection, where we state why the users are on their own when using these distro, and should know what they're doing ;
- Refer to this section for "community supported" installation instructions, with a warning to discourage people of doing so, and stating that the security might be lower on these distributions ;

For Fedora Atomic, and some Ubuntu derivatives, the story might be different than with Arch Linux for instance, so the warning might not apply sometimes, as we're supporting the upstream distros, which can give a certain level of confidence.

If we want/need, given that we have more distributions like this, we could update our table accordingly, with a "security concerns" column, for instance.

apyrgio:
Fair point. How about:

> ## Note on unsupported Linux distros
> 
> For our supported Linux distros, we offer `.deb` and `.rpm` packages to install Dangerzone. This means that users in **other** Debian-based or Fedora-based distros, that are not listed above, may be able to install Dangerzone through these packages. Unfortunately, since we don't test Dangerzone against these distros, it's quite possible that Dangerzone may fail to install, update, run, or be broken in subtle ways. We advise you to proceed only if you know what you're doing, and report to our issue tracker if something is not working.

almet:
Neat! Here is a slightly updated version, where I changed the language to be more direct (well, you know me ;)) and removed the "report to our issue tracker if something is not working" as we are not officially supporting these platforms (we will receive and answer such requests, on a best-effort basis).

> ## Note on unsupported Linux distros
> 
> `.deb` and `.rpm` packages are provided for supported distributions. Users in **other** Debian-based or Fedora-based distros — that are not listed above — may be able to install Dangerzone through these packages. Unfortunately, Dangerzone is not tested against these distros, and might fail to install, update, run, or be broken in subtle ways.
> 
> Please, proceed at your own risks, only if you know what you're doing.

almet:
I've just updated your PR with the text we agreed on in https://github.com/freedomofpress/dangerzone/pull/1273/commits/37dc5f9e582fe6f8e1df0e1649f14ccc978e3dbf. @apyrgio let me know if that's okay for you

apyrgio:
Looks great. I'd add a changelog item, and then we can merge it.

HastD:
Are documentation changes normally recorded in the changelog? I don't see other changes that purely affect documentation in the changelog, and [keepachangelog.com](https://keepachangelog.com/en/1.1.0/) (which is linked at the top of the changelog) mentions documentation changes as an example of something *not* worth giving a changelog entry.

almet:
Nope you're right, changelog entries are not required in this case. They would have been useful if we were openly supporting these new distributions, but as that's not the case we can shortcut it. Just merged, thanks again!

PR #1277, opened 2025-10-13
Skip Podman tasks if a custom runtime was specified

### LINKED ISSUE #1151: Adapt startup tasks for external Podman usage and improve documentation
Now that we've decided we're supporting only Podman (either embedded or external) the only task here is to document the usage of external Podman.

**Old issue:**

We definitely want to embed Podman in Dangerzone to simplify the installation #1145. We should decide if we are going to continue to support:

- Using Docker Desktop instead
- Using external Podman instead

We do not want to support other options like Orbstack, Colima, etc.

apyrgio:
Do not initialize and start Podman machines if a custom runtime was specified, because that's a clear indication that the user does not want Dangerzone to take control over the Podman installation.

Closes #1151

> [!NOTE]
> The accompanying documentation changes will be sent once we release Dangerzone 0.10.0, so as not to confuse our users.

### DIFF
--- a/tests/test_startup.py
+++ b/tests/test_startup.py
@@ -160,3 +160,40 @@ def test_machine_stop_others_task(
         else:
             run_spy.assert_not_called()
             mock_podman_machine_manager.stop.assert_not_called()
+
+
+def test_startup_skips_podman_tasks_if_custom_runtime_is_specified(
+    mocker: MockerFixture,
+) -> None:
+    """Skip Podman startup tasks when a custom runtime is specified."""
+    mock_settings = mocker.patch("dangerzone.settings.Settings").return_value
+    mock_settings.custom_runtime_specified.return_value = True
+    mocker.patch(
+        "dangerzone.startup.ContainerInstallTask.should_skip", return_value=True
+    )
+    mocker.patch("dangerzone.startup.UpdateCheckTask.should_skip", return_value=True)
+
+    tasks = [
+        startup.MachineInitTask(),
+        startup.MachineStartTask(),
+        startup.MachineStopOthersTask(),
+        startup.ContainerInstallTask(),
+        startup.UpdateCheckTask(),
+    ]
+
+    init_run_spy = mocker.spy(tasks[0], "run")
+    init_skip_spy = mocker.spy(tasks[0], "should_skip")
+    start_run_spy = mocker.spy(tasks[1], "run")
+    start_skip_spy = mocker.spy(tasks[1], "should_skip")
+    stop_others_run_spy = mocker.spy(tasks[2], "run")
+    stop_others_skip_spy = mocker.spy(tasks[2], "should_skip")
+
+    runner = startup.StartupLogic(tasks=tasks)
+    runner.run()
+
+    init_run_spy.assert_not_called()
+    assert init_skip_spy.spy_return == True
+    start_run_spy.assert_not_called()
+    assert start_skip_spy.spy_return == True
+    stop_others_run_spy.assert_not_called()
+    assert stop_others_skip_spy.spy_return == True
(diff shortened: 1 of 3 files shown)


tests/test_startup.py
```diff
@@ -160,3 +160,40 @@ def test_machine_stop_others_task(
...
+    stop_others_run_spy = mocker.spy(tasks[2], "run")
+    stop_others_skip_spy = mocker.spy(tasks[2], "should_skip")
+
+    runner = startup.StartupLogic(tasks=tasks)
+    runner.run()
+
+    init_run_spy.assert_not_called()
+    assert init_skip_spy.spy_return == True
+    start_run_spy.assert_not_called()
+    assert init_skip_spy.spy_return == True
```
almet: You probably meant this instead:
```suggestion
    assert init_run_spy.spy_return == True
```

apyrgio:
> Since we're targeting the test/main-icu branch, I believe it would be okay to have the documentation changes here (even though I'm not sure which doc changes you are thinking of)

I'm thinking, something like this:

```diff
diff --git a/README.md b/README.md
index f46e6c37..a3894102 100644
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Dangerzone can convert these types of document into safe PDFs:
 - SVG (`.svg`)
 - other image formats (`.bmp`, `.pnm`, `.pbm`, `.ppm`)
 
-Dangerzone was inspired by [Qubes trusted PDF](https://blog.invisiblethings.org/2013/02/21/converting-untrusted-pdfs-into-trusted.html), but it works in non-Qubes operating systems. It uses containers as sandboxes instead of virtual machines (using Docker for macOS and Windows, and [podman](https://podman.io/) on Linux).
+Dangerzone was inspired by [Qubes trusted PDF](https://blog.invisiblethings.org/2013/02/21/converting-untrusted-pdfs-into-trusted.html), but it works in non-Qubes operating systems as well. It uses containers as sandboxes instead of virtual machines (Podman on macOS, Windows, and Linux) and disposable qubes on Qubes OS.
 
 Set up a development environment by following [these instructions](/BUILD.md).
 
@@ -85,4 +85,5 @@ Dangerzone gets updates to improve its features _and_ to fix problems. So, updat
 
 ### Can I use Podman Desktop?
 
-Yes! We've introduced [experimental support for Podman Desktop](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) on Windows and macOS.
+Yes! Starting with Dangerzone 0.10.0, Podman is used under the hood on Windows
+and macOS, but we have also added [experimental support for Podman Desktop](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) on Windows and macOS.
diff --git a/docs/podman-desktop.md b/docs/podman-desktop.md
index 2a3f0ddb..2fa9fbbe 100644
--- a/docs/podman-desktop.md
+++ b/docs/podman-desktop.md
@@ -6,7 +6,7 @@ experimental. If you try it out and encounter issues, please reach to us, we'll
 be glad to help.
 
 With [Podman Desktop](https://podman-desktop.io/) installed on your machine,
-here are the required steps to change the dangerzone container runtime.
+here are the required steps to change the Dangerzone container runtime.
 
 First, you need to start Podman, and make sure that it's running. Then, you will
 be required to open a terminal and follow these steps:
@@ -60,3 +60,8 @@ podman machine stop
 podman machine set --rootful=false
 podman machine start

+
+### Dangerzone seems to fail during startup
+
+Dangerzone will not manage the Podman machine's lifecycle, which means that you
+must make sure it's started, before Dangerzone starts.
```

I don't want to merge this in `test/main-icu` yet, because there will be a short time between merging that branch in `main` and releasing 0.10.0, in which people might get confused.

PR #1279, opened 2025-10-13
Do not allow to bypass logindex checks when updating

almet:
Here are some changes while reading the ICU source code. The main change is to removal of the logindex bypass mechanism when updating the container

### DIFF
--- a/dangerzone/updater/signatures.py
+++ b/dangerzone/updater/signatures.py
@@ -307,7 +307,13 @@ def upgrade_container_image_airgapped(
                 "The log index is not higher than the last known one"
             )
 
-    image_digest = dz_manifest["manifests"][0].get("digest").replace("sha256:", "")
+    if len(expected_manifest["manifests"]) > 1:
+        raise errors.InvalidDangerzoneManifest(
+            f"Contains more than one manifest: {expected_manifest['manifests']}"
+        )
+    image_digest = (
+        expected_manifest["manifests"][0].get("digest").replace("sha256:", "")
+    )
 
     runtime.load_image_tarball(container_tar)
     # Apply the tag manually here, since images downloaded with `cosign download`


apyrgio:
Looks good :slightly_smiling_face:

dangerzone/updater/signatures.py
```diff
@@ -307,7 +307,13 @@ def upgrade_container_image_airgapped(
                 "The log index is not higher than the last known one"
             )
 
-    image_digest = dz_manifest["manifests"][0].get("digest").replace("sha256:", "")
+    if len(expected_manifest["manifests"]) > 1:
```
almet, line 310: Thanks @redshiftzero for the comment in our code walk-trough session, this could have let to issues down the line (nothing major, but good to have it on check here).

We're now getting the digest from the filtered index.json and not from the dz one, to avoid getting the manifest of the signature or attestation.

dangerzone/updater/signatures.py
```diff
@@ -157,19 +157,18 @@ def check_signatures_and_logindex(
     remote_digest: str,
     pubkey: Path,
     signatures: List[Dict],
-    bypass_logindex_check: bool = False,
 ) -> tuple[int, int]:
+    """Check the validity of the signatures and ensures the logindex are only going upwards"""
```
apyrgio:
```suggestion
    """Check the validity of the signatures and ensure the log index is only going upwards"""
```

dangerzone/updater/signatures.py
```diff
@@ -246,10 +247,15 @@ def upgrade_container_image_airgapped(
...
     upgrade the image and retag it to the expected tag.
 
-    The logic supports "dangerzone archives" only, which have
+    The logic supports "dangerzone archives" only, which have a
     `dangerzone.json` file at the root of the tarball.
 
-    See `prepare_airgapped_archive` for more details.
+    This function is used on airgapped scenarios, but also when the
+    bundled container.tar file is used (e.g. for new releases, until
+    a new container image is availabel in the registry)
```
apyrgio:
```suggestion
    This function is used on airgapped scenarios, as well as when
    loading the bundled container.tar file (e.g. for new releases, until
    a new container image is available in the registry)
```

PR #1281, opened 2025-10-14
Ensure that Dangerzone can run in airgapped environments

### LINKED ISSUE #1280: Find a replacement for cosign's --offline flag
We were surprised to find out that the `--offline` flag in the `cosign verify*` command family does not mean that `cosign` will perform a fully offline verification. It means that it won't reach the Rekor log to verify the log index, if no proof was provided (see https://github.com/sigstore/cosign/issues/4454). It will _still_ make a call to the Sigstore CDN to receive the [`trusted_root`](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) file though, and fail if it can't.

That's something we don't want for our airgapped installations. When we did our initial tests with Cosign, we had assumed that this was working, but probably our tests were not strict enough. If we are to support airgapped installations, we need to do a better test, ideally in our CI, that will force Dangerzone to perform an aigrapped conversion.

apyrgio:
Ensure that Dangerzone can run in airgapped environments, by adding some instructions for power users, and a CI test that verifies them.

Closes #1280

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,10 @@ Dangerzone gets updates to improve its features _and_ to fix problems. So, updat
 ### Can I use Podman Desktop?
 
 Yes! We've introduced [experimental support for Podman Desktop](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) on Windows and macOS.
+
+### Can I run Dangerzone in an airgapped environment?
+
+Yes, Dangerzone is designed to run in airgapped environments without any
+configuration. If you want to update its container image, follow
+[our instructions](docs/developer/independent-container-updates.md#Installing-image-updates-to-airgapped-environments)
+to do so.
--- a/docs/developer/independent-container-updates.md
+++ b/docs/developer/independent-container-updates.md
@@ -6,7 +6,7 @@ This mechanism, known as Independent Container Updates (ICU) allows us to shorte
 
 This increases the security of the conversion process dramatically, making it harder for an attacker to rely on known and patched exploits in our stack.
 
-In order to ensure the sandbox image is trusted, we sign it with [cosign](https://www.sigstore.dev/) signatures, and check them against a key distributed in the Dangerzone application.
+In order to ensure the sandbox image is trusted, we sign it with [Cosign CLI](https://github.com/sigstore/cosign), which is part of the greater [Sigstore](https://www.sigstore.dev/) ecosystem, and verify it against a key distributed in the Dangerzone application.
 
 ## Install updates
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -86,3 +86,9 @@ Dangerzone gets updates to improve its features _and_ to fix problems. So, updat
 ### Can I use Podman Desktop?
 
 Yes! We've introduced [experimental support for Podman Desktop](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) on Windows and macOS.
+
+### Can I run Dangerzone in an airgapped environment?
+
+Yes, please follow [our instructions](docs/airgapped.md) to do so.
+
+Yes! We've introduced [experimental support for Podman Desktop](https://github.com/freedomofpress/dangerzone/blob/main/docs/podman-desktop.md) on Windows and macOS.
```
almet:
```suggestion
```

apyrgio:
@almet: This PR is once more ready for a review, based on our updated understanding in https://github.com/freedomofpress/dangerzone/issues/1280#issuecomment-3422977474.

.github/workflows/ci.yml
```diff
@@ -552,7 +556,10 @@ jobs:
           #     protocol specified". However, we have verified with strace(1)
           #     that the command in the Podman container can read the Xauthority
           #     file successfully.
-          xvfb-run -s '-ac' ./dev_scripts/env.py --distro ${{ matrix.distro }} --version ${{ matrix.version }} run --dev \
+          xvfb-run -s '-ac' \
+            ./dev_scripts/env.py \
+              --distro ${{ matrix.distro }} --version ${{ matrix.version }} \
+              run --dev --offline\
```
almet: Ah! I think we should test for both online and offline. I suggest adding another dimension to the matrix, with offline / online and test on one platform (only) for offline support.

dangerzone/updater/cosign.py
```diff
@@ -14,73 +14,97 @@
...
+        rekor_pub_key = str(get_resource_path("rekor.pub"))
+        log.debug(f"Pinning Rekor public key to {rekor_pub_key}")
+        custom_env["SIGSTORE_REKOR_PUBLIC_KEY"] = rekor_pub_key
+
+    # NOTE: This is an uncommon way to update envvars. We basically want to ensure that
+    # the environment variables we have set above will be passed to the command,
+    # provided that they don't override the ones that the user has set.
+    env = custom_env | os.environ.copy()
+
+    cmd = [_COSIGN_BINARY] + cmd
```
almet, line 43:
```suggestion
    cmd = [_COSIGN_BINARY,] + cmd
```

dev_scripts/env.py
```diff
@@ -745,6 +760,12 @@ def parse_args():
         action="store_true",
         help="Run the command into the dev variant of the Dangerzone environment",
     )
+    parser_run.add_argument(
+        "--offline",
```
almet: `--networkless` seem a better naming choice to me, to not mix things up with the flag we pass to `cosign verify`.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
   machine the Dangerzone uses under the hood
   ([#1172](https://github.com/freedomofpress/dangerzone/issues/1172))
 - Capture all the command outputs in the logs ([#1236](https://github.com/freedomofpress/dangerzone/issues/1172))
+- Allow Dangerzone to run in airgapped environments without extra configuration
```
almet: I don't think this needs a special entry in the changelog as it's part of the work we've done for ICU (dangerzone was already able to run in airgapped envs)

almet (commented):
LGTM, added a few comments, thanks!

dev_scripts/env.py
almet: `--no-network` it is :)

.github/workflows/ci.yml
apyrgio: We discussed about this live. The situation with this PR regarding the connectivity of CI tests is:

Platform | Smoke test | CI test
------------- | ---------------- | -----------
Windows | Online         | Online
macOS     | Online         | Online
Linux        | Online         | :x: **Offline**

This means that we currently have a mix of online and offline CI tests. Ideally, we would like to test every platform both online and offline, but there's the issue of test runtime, which is already large. So, we decided to roll with this for now, and reconsider it in the future.

dangerzone/updater/cosign.py
apyrgio: Not necessary for lists, but necessary for tuples. I think it's fine as is.

dev_scripts/env.py
apyrgio: I'll update it to `--no-network`.

PR #1290, opened 2025-10-22
Add the public key used to sign container updates

almet:
I've hesitated to include the `.asc` here, but since we'll be signing the resulting packages for windows, macOS and Linux, I don't think that's useful.

Let me know if you think otherwise, and I'm happy to add it here.

### DIFF
--- a/docs/developer/python.md
+++ b/docs/developer/python.md
@@ -7,9 +7,8 @@
   > 🚨 Do not use the one from Homebrew as it is known to
   > [cause issues](https://github.com/freedomofpress/dangerzone/issues/471))
 
-
 - [ ] Ensure the `~/.zprofile` of all accounts contains the following lines, and
-  remove any older ones, if they exist:
+      remove any older ones, if they exist:
 
   ```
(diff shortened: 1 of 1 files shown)


apyrgio:
Hm, you have a point, since it's not actually used.

I think we can include it in the docs for independent container updates, as in "just an fyi, our Cosign key is signed by our primary key!". But that's not critical for 0.10.0, unless we find a good use for this signature. So, feel free to merge, and we can communicate our trust for this key in a better way, once we start writing some blog posts.

almet:
Actually added this signature and info in the docs, see [the updated diff](https://github.com/freedomofpress/dangerzone/pull/1290/files).

apyrgio:
Looks good, and we can merge this. However, I have the feeling that I've started seeing this error way too often in our CI: https://github.com/freedomofpress/dangerzone/actions/runs/18745172374/job/53471131573?pr=1290#step:12:977. I might open an issue for this, just to understand if it affects us somehow.

PR #1291, opened 2025-10-22
Expand third-party notice with GitHub assets

### LINKED ISSUE #1224: Licenses for vendored assets
Now that we are vendoring some third-party assets in our packages, we should find a place to add their copyright notices,

apyrgio:
Expand our third-party notice file, with information about the GitHub assets that we bundle in our repo and installers. For concrete information about each binary (URL, checksum, version), we point to our `mazette.lock`.

Closes #1224


THIRD_PARTY_NOTICE
```diff
@@ -2,13 +2,59 @@ This project includes third-party components as follows:
...
+   - Included in: Dangerzone installers (Windows MSI and macOS DMG)
+
+4. Podman
+   - URL: See `podman` section in `mazette.lock`
+   - License: Apache-2.0
+   - Last updated: N/A
+   - Description: A tool for managing OCI containers and pods
+   - Included in: Dangerzone installers (Windows MSI and macOS DMG)
+
+5. gvproxy
```
almet: I wouldn't prefix with a number here to avoid issues. Here you are using twice items number 5.

Also, what about using a `.md` format here?

If we did this, I think you can prefix every number by `0.` and github will render the file with incremental numbers instead.

THIRD_PARTY_NOTICE
```diff
@@ -2,13 +2,59 @@ This project includes third-party components as follows:
...
+   - Included in: Dangerzone installers (macOS DMG)
+
+5. vfkit
+   - URL: See `vfkit` section in `mazette.lock`
+   - License: Apache-2.0
+   - Last updated: N/A
+   - Description: Command-line tool to start VMs on macOS
+   - Included in: Dangerzone installers (macOS DMG)
 
+6. Podman Machine OS disk image
```
almet: I would put this just right after Podman. Also, what about sorting things alphabetically?

almet (approved):
Looks good to me.

If possible I would like to use markdown instead of a raw file without extension, other than that looks good to me. I've added a few other comments. As usual, nothing major, so feel free to implement them or not.

README.md
```diff
@@ -64,6 +64,8 @@ Licensed under the AGPLv3: [https://opensource.org/licenses/agpl-3.0](https://op
 Copyright (c) 2022-2024 Freedom of the Press Foundation and Dangerzone contributors
 Copyright (c) 2020-2021 First Look Media
 
+See also [THIRD_PARTY_NOTICE](THIRD_PARTY_NOTICE) for more information regarding the third-party software that Dangerzone depends on.
```
almet: (Wee need to add the .md extension then)

THIRD_PARTY_NOTICE
apyrgio: > I wouldn't prefix with a number here to avoid issues. Here you are using twice items number 5.

D'oh!

I have significantly changed the structure of this file in 8c4b91e0457d015341f7a6c47c38a6c4d12c5b9f. Take a look.

THIRD_PARTY_NOTICE
apyrgio: Done in 8c4b91e0457d015341f7a6c47c38a6c4d12c5b9f.

almet:
Nice, LGTM, let's merge :-)

README.md
apyrgio: (Just did so)

THIRD_PARTY_NOTICE
apyrgio: Btw, this was based on this https://github.com/awslabs/git-remote-s3/blob/5775df2bd9d1375baa95b0b29d02c2776f583dbc/THIRD_PARTY_NOTICE.md?plain=1. Unfortunately, a lot of the projects did not have copyright notices (just licenses), so I ditched that field.

PR #1298, opened 2025-10-25
[FIX] - invalid path found in Docker file

compilable:
- Bug:
When using the Docker file in Ubuntu, Debian using `docker build -t dangerzone .` below error occurred due to an invalid path.

- Initial error:

image

- Fix: 
The `entrypoint.py` file exists, but it's located at `dangerzone` folder, not in `container_helpers/entrypoint.py`. The Dockerfile is looking in the wrong path.

Fixed container_helpers mount paths:
- ./container_helpers/repro-sources-list.sh → repro-sources-list.sh
- ./container_helpers/gvisor.key → gvisor.key

Fixed conversion COPY path:
- COPY conversion/*.py → COPY dangerzone/conversion/*.py
Fixed entrypoint COPY path:
- COPY container_helpers/entrypoint.py → COPY dangerzone/container_helpers/entrypoint.py

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ RUN mkdir -p /opt/dangerzone/dangerzone
 RUN touch /opt/dangerzone/dangerzone/__init__.py
 
 # Copy only the Python code, and not any produced .pyc files.
-COPY conversion/*.py /opt/dangerzone/dangerzone/conversion/
+COPY dangerzone/conversion/*.py /opt/dangerzone/dangerzone/conversion/
 
 # Create a directory that will be used by gVisor as the place where it will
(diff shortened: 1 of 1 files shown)


almet:
Hi, thanks for the feedback.

To build the image, you can use the script in `install/common/build-image.py`, as specified in https://github.com/freedomofpress/dangerzone/blob/main/BUILD.md#building-a-local-container-image.

Does that work for you?

almet:
I'm closing this PR, thanks for proposing the changes, but I believe the instructions I provided explain why we're using these paths right now.

Don't hesitate to let us know if think we're missing something here!

PR #1301, opened 2025-10-27
Bump the logindex of the bundled image

almet:
https://search.sigstore.dev/?logIndex=634403782

### DIFF
--- a/dangerzone/updater/signatures.py
+++ b/dangerzone/updater/signatures.py
@@ -36,7 +36,7 @@ def appdata_dir() -> Path:
 
 # RELEASE: Bump this value to the log index of the latest signature
 # to ensure the software can't upgrade to container images that predates it.
-BUNDLED_LOG_INDEX = 207673688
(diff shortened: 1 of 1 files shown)


apyrgio:
The change looks good, so feel free to merge.

What I think would help for the next PR is to have a way to ensure that this log index is indeed the correct one. This requires going to the GHCR signer repo, looking at the last signed image, and grabbing the log index from there. A more automated way to do so is:

```
$ ls -t SIGNATURES \
  | grep -v README.md \
  | head -1 \
  | xargs -I {} find "SIGNATURES/{}" -name MANIFEST \
  | xargs cat \
  | jq -r '.layers[].annotations."dev.sigstore.cosign/bundle"' \
  | jq .Payload.logIndex \
  | sort
634403782
634404089
634404358
```

The log index we need is the smallest one, else we risk ignoring an image for an architecture. We can automate this check, by factoring out this index to a file, and having a CI test that runs just for this file to clone GHCR signer and do the above.

PR #1304, opened 2025-10-30
Update dodo.py file to use prepare-image

almet:


dodo.py
```diff
@@ -184,22 +180,18 @@ def task_install_assets():
...
-            ["cp", img_id_src, img_id_dst],
+            f"poetry run ./dev_scripts/dangerzone-image prepare-archive --output {img_src}",
         ],
-        "file_dep": IMAGE_DEPS,
-        "targets": [img_src, img_dst, img_id_src, img_id_dst],
-        "task_dep": ["init_release_dir", "check_container_runtime"],
+        "targets": [
+            img_src,
+        ],
+        "task_dep": ["init_release_dir", "check_container_runtime", "install_assets"],
```
apyrgio:
```suggestion
        "task_dep": ["init_release_dir", "install_assets"],
```

apyrgio:
Approved with a minor fix, thanks!

dodo.py
almet: Updated and rebased

PR #1323, opened 2025-11-06
Set UTF-16 encoding when running WSL

### LINKED ISSUE #1319: Fix logs copy-paste on Windows
On Windows, the wsl command seem to b0rk the output, and make it impossible to copy/paste. We should find a way to solve this.

almet:
Set the encoding to non-BOM UTF16, which is what WSL uses. If not set, stdout will contain invisible null characters which renders the output impossible to read and copy/paste.

Fixes #1319


apyrgio:
Slick, let's merge.

PR #1333, opened 2025-11-13
Add startup task that installs WSL

### LINKED ISSUE #1312: 0.10.0-rc1 Fails to initialize on windows (wsl --update)
### What happened?

Failed to initialize.

### operating system version

Windows 10 (newly installed) and I had never messed (or installed / updated) wsl.exe

### Dangerzone version

v0.10.0-rc1

### Podman info

```shell
Starting Podman machine: dz-internal-0.10.0
Error: dz-internal-0.10.0: VM does not exist

(I didn't bother copying and pasting the command, but it failed with error 125)
```

### Document conversion logs

```shell

```

### Additional info

It also showed the following in the logs and it appears to be a wsl.exe not getting called with the right parameters:

Image

But for some reason I could not copy and paste this text.

apyrgio:
This PR adds a startup task that can install WSL, and do some vital things so that users understand what's going on (prompt the user to install it and reboot, show a helpful message in case of an error, etc.)

To achieve this, we introduce a new `dangerzone/windows` directory, in which we add two Python modules that perform some windows only tasks:
* `dangerzone.windows.shellexec`: This module provides some convenience wrappers on top of [ShellExecuteEx](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexw).
* `dangerzone.windows.wsl`: This module is responsible for the invocation of `wsl.exe` commands, as well as providing a thin layer or logic above them.

Fixes #1312

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,14 @@ jobs:
           enableCrossOsArchive: true
           fail-on-cache-miss: true
           key: v6-container-${{ needs.build-container-image.outputs.image_uri }}
+      # NOTE: We have noticed that the `windows-2022` runner image has WSL
+      # installed, but its kernel is old (plus, the default version is WSL1).
+      # This means that `wsl --status` fails, which by extension means that our
+      # `dangerzone-cli` command and its WSL detection will fail.
+      #
+      # For this reason, we prefer to run `wsl --update` before the tests.
+      - name: Update WSL
+        run: wsl --update
       - name: Smoke test
         run: poetry run .\dev_scripts\dangerzone-cli.bat .\tests\test_docs\sample-pdf.pdf --ocr-lang eng --debug
       - name: Run CLI tests
--- a/dangerzone/podman/machine.py
+++ b/dangerzone/podman/machine.py
@@ -81,12 +81,6 @@ class PodmanMachineManager:
             logger.info(f"Podman machine '{self.name}' already exists.")
             return
 
-        if platform.system() == "Windows":
-            # Set the encoding to non-BOM UTF16, which is what WSL uses.
-            # If not set, stdout will contain invisible null characters which
-            # renders the output impossible to read and copy/paste.
-            util.subprocess_run(["wsl", "--update"], encoding="UTF-16LE", check=False)
-
         logger.info(f"Initializing Podman machine: {self.name}")
         self.podman.machine.init(
             name=self.name,
--- a/tests/podman/test_machine.py
+++ b/tests/podman/test_machine.py
@@ -50,8 +50,6 @@ def test_initialize_machine_no_existing(
     version = get_version()
     machine_name = f"dz-internal-{version}"
     image_path = str(machine_manager._get_machine_image_path())
-    if platform.system() == "Windows":
-        rec_wsl_update = fp.register(["wsl", "--update"])
 
     rec_list = podman_register(
         ["machine", "list", "--format", "json"], stdout=json.dumps([])
@@ -68,8 +66,6 @@ def test_initialize_machine_no_existing(
         ]
     )
     machine_manager.init()
-    if platform.system() == "Windows":
-        assert rec_wsl_update.call_count() == 1
     assert rec_list.call_count() == 1
     assert rec_init.call_count() == 1
 
@@ -84,8 +80,6 @@ def test_initialize_machine_stale_exists(
     machine_name = f"dz-internal-{version}"
     stale_machine_name = "dz-internal-stale"
     image_path = str(machine_manager._get_machine_image_path())
-    if platform.system() == "Windows":
-        rec_wsl_update = fp.register(["wsl", "--update"])
 
     rec_list = podman_register(
         ["machine", "list", "--format", "json"],
@@ -105,8 +99,6 @@ def test_initialize_machine_stale_exists(
         ]
     )
     machine_manager.init()
-    if platform.system() == "Windows":
-        assert rec_wsl_update.call_count() == 1
     assert rec_list.call_count() == 1
     assert rec_rm.call_count() == 1
     assert rec_init.call_count() == 1
@@ -291,8 +283,6 @@ def test_initialize_machine_with_timezone(
     version = get_version()
     machine_name = f"dz-internal-{version}"
     image_path = str(machine_manager._get_machine_image_path())
-    if platform.system() == "Windows":
-        rec_wsl_update = fp.register(["wsl", "--update"])
     rec_list = podman_register(
         ["machine", "list", "--format", "json"], stdout=json.dumps([])
     )
@@ -308,7 +298,5 @@ def test_initialize_machine_with_timezone(
         ]
     )
     machine_manager.init(timezone="America/New_York")
-    if platform.system() == "Windows":
-        assert rec_wsl_update.call_count() == 1
     assert rec_list.call_count() == 1
     assert rec_init.call_count() == 1


dangerzone/gui/main_window.py
```diff
@@ -78,6 +78,47 @@
...
 """
 
+WSL_NEEDS_INSTALL_MSG = """\
+<p>Dangerzone requires a Microsoft Windows feature called
+<a href="https://learn.microsoft.com/en-us/windows/wsl/">
+Windows Subsystem for Linux (WSL)
+</a>, which is not present in your system.
+<p>You can let Dangerzone download and install it now, or quit Dangerzone and install it
+on your own.</p>
+"""
```
almet, line 88: In this message and the one below, I find this is a lot of times "Dangerzone". 

```suggestion
WSL_NEEDS_INSTALL_MSG = """\
<p>Dangerzone requires a Microsoft Windows feature called
<a href="https://learn.microsoft.com/en-us/windows/wsl/">
Windows Subsystem for Linux (WSL)
</a>, which is not present in your system.
<p>Should we download and install it now, or quit Dangerzone and install it
on your own?</p>
"""
```

dangerzone/gui/main_window.py
```diff
@@ -78,6 +78,47 @@
...
+
+WSL_NEEDS_REBOOT_MSG = """\
+<p>Dangerzone has installed the Windows Subsystem for Linux (WSL), but you need to
+reboot your computer before Dangerzone can use it.</p>
+<p>You can either reboot your computer now, or quit Dangerzone and reboot it
+later.</p>
+"""
```
almet, line 95: Same, plus I removed the last sentence as it's a duplicate with the options shown to the user.
```suggestion
WSL_NEEDS_REBOOT_MSG = """\
<p>The Windows Subsystem for Linux (WSL) is now installed, but you need to reboot your computer
to make Dangerzone work.</p>
"""
```

dangerzone/startup.py
```diff
@@ -168,6 +169,47 @@ def run(self) -> None:
             raise RuntimeError("Failed to stop all other running Podman machines.")
 
 
+class WSLInstallTask(Task):
+    name = "Installing Windows Subsystem for Linux"
+
+    def should_skip(self) -> bool:
+        return platform.system() != "Windows" or wsl.is_wsl_installed()
```
almet: since we use a namespace, let's simplify the name of the functions:

```suggestion
        return platform.system() != "Windows" or wsl.is_installed()
```

dangerzone/startup.py
```diff
@@ -168,6 +169,47 @@ def run(self) -> None:
...
+
+    def run(self) -> None:
+        if not self.prompt_install():
+            raise errors.WSLNotInstalled("User chose to quit instead of installing WSL")
+
+        try:
+            wsl.install_wsl_and_check_reboot()
```
almet:
```suggestion
            wsl.install_and_check_reboot()
```

dangerzone/util.py
```diff
@@ -1,16 +1,27 @@
...
+from typing import Any
 
 try:
     import platformdirs
 except ImportError:
     import appdirs as platformdirs  # type: ignore[no-redef]
 
 
+# NOTE: We originally used `subprocess.STARTF_USESHOWWINDOW` here, but there's a more
+# modern way since Python 3.7 (see also https://github.com/python/cpython/issues/85785)
```
almet, line 18: Thanks. I'm a bit reluctant to change this behavior this close to the final release. OK if we stick with the old way for now, and fix it right after `0.10.0` is out?

dangerzone/windows/shellexec.py
```diff
@@ -0,0 +1,119 @@
...
+
+from .. import errors
+
+log = logging.getLogger(__name__)
+
+
+SEE_MASK_NOCLOSEPROCESS = 0x00000040
+TIMEOUT_INFINITE = 0xFFFFFFFF
+
+
```
almet, line 15: Adding a link to this URL might prove helpful: https://learn.microsoft.com/en-us/windows/win32/api/shellapi/ns-shellapi-shellexecuteinfoa

dangerzone/windows/shellexec.py
```diff
@@ -0,0 +1,119 @@
...
+    Returns:
+        The exit code of the process.
+
+    Raises:
+        errors.WinShellExecStartFailure: If ShellExecuteExW fails.
+        errors.WinShellExecNoHandle: If the process handle is not returned.
+        errors.WinShellExecTimeoutExpired: If the process does not finish within the timeout.
+
+    Further reading:
+    - ShellExecuteExW: https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexw
```
almet:
```suggestion
    - https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexw
```

dangerzone/windows/wsl.py
```diff
@@ -0,0 +1,117 @@
...
+log = logging.getLogger(__name__)
+
+# In the following commands, we set the encoding to non-BOM UTF16, which is what WSL
+# uses. If not set, stdout will contain invisible null characters which renders the
+# output impossible to read and copy/paste.
+# Note that this is not necessary for commands initiated with ShellExecuteEx, since they
+# don't print to a handle that Python controls, but to a terminal spawned by Windows.
+
+
+def wsl_list() -> str:
```
almet: I think these functions should be named `install`, `status`, etc, and used with their namespaces:

```python
from dangerzone.windows import wsl

wsl.install() # etc.
```

dangerzone/errors.py
```diff
@@ -158,5 +158,40 @@ class ContainerPullException(ContainerException):
     pass
 
 
-class OtherMachineRunningError(ContainerException):
+class OtherMachineRunningError(Exception):
+    pass
+
+
+class WSLInstallFailed(Exception):
```
almet: Probably should create a `WSLException` parent class and inherit it in the other exceptions, so it's easier to catch.

almet (approved):
Nice, well done Alex!

The wizardry to do a shell exec and ask for privileges on Windows explains why you seemed to lose some mental health in the past few days 😁 

Taking some perspective, I'm a bit concerned about introducing a way to have a windows prompt with elevated privileges from Dangerzone.

My fear is that an attacker could leverage this ability to trick the user into thinking they're updating their WSL installation, whereas they would be using this to do something else with elevated privileges.

Probably, if the attacker has a way to run python code on the host, it's already game over, but let's be sure we are fine with this before merging.

Alternative options might require the users to download an utility to enable WSL instead of shipping this inside the app. I'm not sure the trouble of building and distributing such a binary is worth it, but it might be.

Other than that, I added a few comments, but I think nothing major. Let me know when you've had a look!

dangerzone/gui/main_window.py
apyrgio: Hm, I want to explain to the user that rebooting is the only option to move forward, but they can do it later on, if they want to. I fear that "Quit Dangerzone" on its own does not get this message across. So, how about this rewording:

```suggestion
WSL_NEEDS_REBOOT_MSG = """\
<p>The Windows Subsystem for Linux (WSL) is now installed, but you need to
reboot your computer to make Dangerzone work.</p>
<p>You can either reboot your computer now, or quit Dangerzone and reboot it
later.</p>
"""
```

dangerzone/util.py
apyrgio: Makes sense, I can keep the startupinfo approach, but keep the move to `dangerzone.util`.

dangerzone/gui/main_window.py
almet: Hmm, I think the same: the first text conveys the meaning just fine, no need to repeat.

> "but you need to reboot your computer to make Dangerzone work"

Not a bit deal though, use the version you prefer ;)

apyrgio: I kept the middle ground one (and now I wonder if I should turn it into a question, like the previous prompt, but I don't know what's best UX-wise).

dangerzone/startup.py
apyrgio: Done in 040169ab.

dangerzone/startup.py
apyrgio: Done in 040169ab

dangerzone/util.py
apyrgio: Done in 12266ed6

dangerzone/windows/shellexec.py
apyrgio: Done in dd8dce47

dangerzone/errors.py
apyrgio: Done in e282fc43

apyrgio:
> Taking some perspective, I'm a bit concerned about introducing a way to have a windows prompt with elevated privileges from Dangerzone.

One quick thing I wanted to point here is that Dangerzone does not elevate its privileges, nor does it call a command with elevated privileges. It actually calls `wsl --install` / `wsl --update` via `ShellExecuteEx` **without** elevated privileges, since the `runAs` verb is not used. Then, `wsl` command is free to elevate its privileges, and prompt the user via UAC.

One of the reasons we don't want to call the command with elevated privileges is to allow installation by users that are not administrators (not checked though, plus WSL must already be installed).

dangerzone/windows/wsl.py
```diff
@@ -0,0 +1,117 @@
...
+log = logging.getLogger(__name__)
+
+# In the following commands, we set the encoding to non-BOM UTF16, which is what WSL
+# uses. If not set, stdout will contain invisible null characters which renders the
+# output impossible to read and copy/paste.
+# Note that this is not necessary for commands initiated with ShellExecuteEx, since they
+# don't print to a handle that Python controls, but to a terminal spawned by Windows.
+
+
+def ls() -> str:
```
almet: This doesn't seem to be used anywhere in the code?

apyrgio: You're right, it's not used. I originally tested WSL detection with `wsl --list --quiet` (which is `ls()`), but then switched to `wsl --status`. I can remove it if you want.

almet:
> Then, wsl command is free to elevate its privileges, and prompt the user via UAC.

Thanks for the clarification. I though that UAC was only allowed because we run via `ShellExecuteEx`, but that's probably me not understanding the how this security feature works.

If you feel it's okay to do that here, let's go :-)

dangerzone/windows/wsl.py
almet: Yup, that makes sense to remove it 👍

apyrgio: Just did so and pushed.

Issue #1327, opened 2025-11-06
Test 0.10.0 RC1 in Tails


almet:
We want to ensure that new container images aren't blocking the user from just using dangerzone. In addition, we might want to find ways to unblock the `podman pull` from Tor, by either using a mirror and/or asking MS Github about it.

almet:
I've successfully installed the 0.10.0 RC1 version of Dangerzone in Tails 7.1.

Installation works as intended, but as we noted previously, it is not possible to contact `ghcr.io` when checking for new container updates (see logs at the end of this message)

It's interesting to note that the interface currently doesn't display anything to the user, and so they don't know that something failed in the background. This can lead them to think that they are running the latest version of the container image, where in fact they are using the bundled one, even though they've enabled automatic sandbox updates.

We could do the following to fix this situation:

1. Catch the exceptions and display an error in the interface, so that users are notified that a failure happened.
2. Notify this to Microsoft GitHub, as they're blocking access to the ghcr.io from users coming from the Tor network.
3. Find alternative ways to download container images from within Tails / when using a Tor connection. This might be us mirroring the `ghcr.io/freedomofthepress/dangerzone/v1` package to somewhere else, or any other solution.


<details><summary>Logs</summary>
<p>

```traceback
Starting Dangerzone background tasks
Performing some Dangerzone startup tasks
Task 'Stopping other Podman VMs' will be skipped
Task 'Initializing Dangerzone VM' will be skipped
Task 'Starting Dangerzone VM' will be skipped
Task 'Check for updates' is starting...
Get manifest digests
Logging to the remote registry
Starting new HTTPS connection (1): ghcr.io:443
Encountered an error while checking for upgrades
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/urllib3/connection.py", line 198, in _new_conn
    sock = connection.create_connection(
        (self._dns_host, self.port),
    ...<2 lines>...
        socket_options=self.socket_options,
    )
  File "/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 85, in create_connection
    raise err
  File "/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 73, in create_connection
    sock.connect(sa)
    ~~~~~~~~~~~~^^^^
ConnectionRefusedError: [Errno 111] Connexion refusée

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #1336, opened 2025-11-14
Make the `podman pull` command and registry lookups go trough Tor on Tails

### LINKED ISSUE #1335: Make requests use a Socks proxy on Tails
If we detect we're running in Tails (we can check with `cat /etc/os-release`), we might want to make our `requests` requests go trough the socks proxy, by specifying an environment variable like this (taken [from the Tails bug tracker](https://gitlab.tails.boum.org/tails/tails/-/merge_requests/1123/diffs)):

```
username=curl-$(head -c 12 /dev/urandom | xxd -p)
ALL_PROXY="socks5://${username}:0@127.0.0.1:9050"
```

Or [passing the `proxies` argument](https://requests.readthedocs.io/en/latest/user/advanced/#proxies) in the python library.

To make this possible, the PySocks library[ is required](https://github.com/psf/requests/blob/main/setup.py#L100C9-L100C46). Fortunately, this is packaged for [Debian](https://packages.debian.org/trixie/python3-socks) and [Fedora](https://packages.fedoraproject.org/pkgs/python-socks/python3-socks/) under the `python3-socks` name.

almet:
On Tails, requests are blocked by default and need to go trough a SOCKS5 proxy to do requests over Tor; This implements it for `podman pull` and registry lookups.

Fix #1335
Refs #1327

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,8 @@ Rules-Requires-Root: no
 
 Package: dangerzone
 Architecture: any
-Depends: ${misc:Depends}, podman, python3, python3-pyside6.qtcore | python3-pyside2.qtcore, python3-pyside6.qtgui | python3-pyside2.qtgui, python3-pyside6.qtwidgets | python3-pyside2.qtwidgets, python3-pyside6.qtsvgwidgets | python3-pyside2.qtsvg, python3-platformdirs | python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-markdown, python3-packaging, tesseract-ocr-all
+# The dependency to python3-socks is only present here because it's only useful on Tails.
+Depends: ${misc:Depends}, podman, python3, python3-pyside6.qtcore | python3-pyside2.qtcore, python3-pyside6.qtgui | python3-pyside2.qtgui, python3-pyside6.qtwidgets | python3-pyside2.qtwidgets, python3-pyside6.qtsvgwidgets | python3-pyside2.qtsvg, python3-platformdirs | python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-socks, python3-markdown, python3-packaging, tesseract-ocr-all
 Description: Take potentially dangerous PDFs, office documents, or images
   Dangerzone is an open source desktop application that takes potentially dangerous PDFs, office documents, or images and converts them to safe PDFs. It uses disposable VMs on Qubes OS, or container technology in other OSes, to convert the documents within a secure sandbox.
   .


dangerzone/util.py
```diff
@@ -142,3 +144,20 @@ def format_exception(e: Exception) -> str:
...
+        if os_release_path.exists():
+            os_release = os_release_path.read_text()
+            return any([name in os_release for name in names])
+    return False
+
+
+def get_tails_socks_proxy() -> str:
+    """Generate a SOCKS5 proxy connection address that works on Tails"""
+    username = os.urandom(8).hex()
+    return f"socks5://{username}:0@127.0.0.1:9050"
```
apyrgio: I wonder, is the username necessary at all?

debian/control
```diff
@@ -9,7 +9,7 @@ Rules-Requires-Root: no
 
 Package: dangerzone
 Architecture: any
-Depends: ${misc:Depends}, podman, python3, python3-pyside6.qtcore | python3-pyside2.qtcore, python3-pyside6.qtgui | python3-pyside2.qtgui, python3-pyside6.qtwidgets | python3-pyside2.qtwidgets, python3-pyside6.qtsvgwidgets | python3-pyside2.qtsvg, python3-platformdirs | python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-markdown, python3-packaging, tesseract-ocr-all
+Depends: ${misc:Depends}, podman, python3, python3-pyside6.qtcore | python3-pyside2.qtcore, python3-pyside6.qtgui | python3-pyside2.qtgui, python3-pyside6.qtwidgets | python3-pyside2.qtwidgets, python3-pyside6.qtsvgwidgets | python3-pyside2.qtsvg, python3-platformdirs | python3-appdirs, python3-click, python3-xdg, python3-colorama, python3-requests, python3-socks, python3-markdown, python3-packaging, tesseract-ocr-all
```
apyrgio, line 13: Should we add a comment here about why this dependency is not in `pyproject.toml`?

dangerzone/util.py
almet: A different username is useful to use different Tor circuits, in their implementation.

apyrgio: Oh, that's cool. Feel free to add a comment, if there's some doc on this.

almet: I've added some docs to the function to make it clearer, thanks for the notice:

https://github.com/freedomofpress/dangerzone/pull/1336/files#diff-1525bcad87e5d0373c42b20400eb1b770a3121d31debb5ecc9192e539ba07786R160-R172

debian/control
almet: Yep, added a comment, thanks!

dangerzone/util.py
```diff
@@ -142,3 +144,29 @@ def format_exception(e: Exception) -> str:
...
+            return any([name in os_release for name in names])
+    return False
+
+
+def get_tails_socks_proxy() -> str:
+    """
+    Generate a SOCKS5 proxy connection address that works on Tails.
+
+    the username part makes it use stream isolation [0], which allow to isolate
+    unrelated streams, putting them on serparate circuits so that semantically
```
apyrgio:
```suggestion
    The username part makes it use stream isolation [0], which allows to isolate
    unrelated streams, putting them on separate circuits so that semantically
```

dangerzone/util.py
```diff
@@ -142,3 +144,29 @@ def format_exception(e: Exception) -> str:
...
+    unrelated traffic is not inadvertently made linkable [1].
+
+    [0] https://spec.torproject.org/socks-extensions.html#extended-auth
+    [1] https://spec.torproject.org/proposals/171-separate-streams.txt
+    """
+    username = os.urandom(8).hex()
+    return f"socks5://dz-{username}:0@127.0.0.1:9050"
```
apyrgio: I'm missing something here. The spec says that:

> New client implementations SHOULD NOT use legacy isolation. 

So, should we use `<torS0X>` here?

dangerzone/util.py
almet: Thanks. I've updated the PR with this.

dangerzone/util.py
almet: Thanks for spotting. Just did so (was un-commited for some reason)

apyrgio:
Thanks for the changes Alexis, looks good.

PR #1343, opened 2025-11-18
Ask the user before downloading new container images

### LINKED ISSUE #1334: Always ask the user before downloading a new container image
We want to prompt the user for each new container image, before doing the `podman pull`.
We are currently thinking about adding a `[ ] Always download new container images without asking` option in this prompt.

almet:
Also introduce a new setting to automatically download new container images in the future.

Fix #1334

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -119,6 +119,13 @@ report an issue ↗️
 </p>
 """
 
+DOWNLOAD_NEW_CONTAINER_IMAGE_MSG = """\
+<p>A new version of the secure sandbox is available and should be downloaded.</p>
+<p>The download size is usually about 500MiB.</p>
+ <p>Should we proceed now? (If you skip, you will be asked the next time your run Dangerzone)</p>
+<p>Keeping the sandbox up to date helps avoiding security vulnerabilities.
+ It's recommended to always apply updates.</p>
+"""
 
 DEBIAN_BULLSEYE_DEPRECATION_MSG = """\
 <p><b>Warning:</b> Debian Bullseye systems and their derivatives will
@@ -488,7 +495,9 @@ class MainWindow(QtWidgets.QMainWindow):
             self.handle_container_update_available
         )
         task_update_check.completed.connect(self.handle_update_check_completed)
-        task_update_check.needs_user_input.connect(self.handle_needs_user_input)
+        task_update_check.needs_user_input.connect(
+            self.handle_needs_user_input_enable_updates
+        )
 
         task_container_install.load_container.connect(
             self.status_bar.handle_task_container_install_local
@@ -506,6 +515,9 @@ class MainWindow(QtWidgets.QMainWindow):
             self.log_window.handle_task_container_install_failed
         )
         task_container_install.failed.connect(self.handle_startup_error)
+        task_container_install.needs_user_input.connect(
+            self.handle_needs_user_input_install_remote_container
+        )
 
         self.show()
 
@@ -659,7 +671,7 @@ class MainWindow(QtWidgets.QMainWindow):
         self.dangerzone.settings.set("updater_errors", 0)
         self.dangerzone.settings.save()
 
-    def handle_needs_user_input(self) -> None:
+    def handle_needs_user_input_enable_updates(self) -> None:
         check = prompt_for_checks(self.dangerzone)
         if check is not None:
             self.dangerzone.settings.set("updater_check_all", check, autosave=True)
@@ -734,6 +746,33 @@ class MainWindow(QtWidgets.QMainWindow):
     def handle_wsl_install_failed(self, msg: str) -> None:
         self.status_bar.handle_startup_error()
         self.show_wsl_error_widget()
+
+    def handle_needs_user_input_install_remote_container(
+        self, req: startup.PromptRequest
+    ) -> None:
+        log.debug("Prompting user to download new remote container")
+        question = Question(
+            self.dangerzone,
+            title="Download Sandbox",
+            message=DOWNLOAD_NEW_CONTAINER_IMAGE_MSG,
+            ok_text="Yes, download update",
+            cancel_text="Skip for now",
+            checkbox_text="Always download sandbox updates",
+        )
+        assert question.checkbox is not None
+        result = question.launch()
+        if question.checkbox.isChecked():
+            self.dangerzone.settings.set(
+                "updater_ask_before_download",
+                False,
+                autosave=True,
+            )
+        if result == Question.Accepted:
+            log.debug("Accepted")
+            req.reply(True)
+            return
+        log.debug("Refused")
+        req.reply(False)
 
     def handle_startup_error(self, msg: str) -> None:
         self.status_bar.handle_startup_error()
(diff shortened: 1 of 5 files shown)


dangerzone/gui/main_window.py
```diff
@@ -119,6 +119,12 @@
 </p>
 """
 
+DOWNLOAD_NEW_CONTAINER_IMAGE_MSG = """\
+<p>A new version of the secure sandbox is available and should be downloaded.
+ Should we proceed now?</p>
+<p>Keeping the sandbox up to date helps avoiding security vulnerabilities.
+ It's recommended to always apply updates.</p>
```
apyrgio: I want to address two things here:

1. Mention that the size of the download is usually around 500MiB, so that the users can make a more educated guess.
2. Mention that if they skip the update, they will be asked the next time they use Dangerzone.

Proposed text:

```suggestion
<p>A new version of the secure sandbox is available and should be downloaded.
The download size is usually around 500MiB.
Should we download and install it now, or skip it and ask the next time you run Dangerzone?</p>
<p>Keeping the sandbox up to date helps avoiding security vulnerabilities.
 It's recommended to always apply updates.</p>
```

dangerzone/gui/main_window.py
```diff
@@ -735,6 +746,34 @@ def handle_wsl_install_failed(self, msg: str) -> None:
...
+        self, req: startup.PromptRequest
+    ) -> None:
+        log.debug("Prompting user to download new remote container")
+        question = Question(
+            self.dangerzone,
+            title="Download Sandbox",
+            message=DOWNLOAD_NEW_CONTAINER_IMAGE_MSG,
+            ok_text="OK",
+            cancel_text="Skip for now (don't)",
+            checkbox_text="Always download Sandbox updates",
```
apyrgio:
```suggestion
            ok_text="Yes, download update",
            cancel_text="No, skip for now",
            checkbox_text="Always download sandbox updates",
```

dangerzone/gui/main_window.py
```diff
@@ -735,6 +746,34 @@ def handle_wsl_install_failed(self, msg: str) -> None:
...
+        result = question.launch()
+        if question.checkbox.isChecked():
+            self.dangerzone.settings.set(
+                "updater_ask_before_download",
+                False,
+                autosave=True,
+            )
+        if result == Question.Accepted:
+            log.debug("Accepted")
+            req.req_data["signal"].emit()
```
apyrgio: Do we need to manually emit a signal, if the `_download_container` method will do it for us?

dangerzone/gui/startup.py
```diff
@@ -109,16 +112,33 @@ def prompt_reboot(self) -> bool:
...
+                self._download_container()
+
+    def _download_container(self):
+        self.download_container.emit()
+        super().run()
+
+    def prompt_user(self):
+        return PromptRequest().ask(
+            self.needs_user_input, data={"signal": self.download_container}
+        )
```
apyrgio: If we end up not emitting a `download_container` signal on the main window, then we don't need to pass the signal here.

(Most likely, we can keep the entire `run()` method as is, and just do a `return` if the response is False, but I'll leave that to you)

apyrgio: Also, do we actually need a `PromptRequest` here? Can't we simply prompt from the startup thread?

dangerzone/gui/main_window.py
almet: Thanks for the feedback!

I updated the language in e8ce90c133b737f9b32271760a3ae16931ea247c

dangerzone/gui/main_window.py
almet: Same, see e8ce90c133b737f9b32271760a3ae16931ea247c

dangerzone/gui/startup.py
almet: Following your comment, I tried a few different ways to handle this:

## Directly pass a callback

```diff
diff --git a/dangerzone/gui/main_window.py b/dangerzone/gui/main_window.py
index 5e2ca483..27aa4850 100644
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -748,7 +748,7 @@ class MainWindow(QtWidgets.QMainWindow):
         self.show_wsl_error_widget()
 
     def handle_needs_user_input_install_remote_container(
-        self, req: startup.PromptRequest
+        self, on_confirmation: callable
     ) -> None:
         log.debug("Prompting user to download new remote container")
         question = Question(
@@ -768,12 +768,10 @@ class MainWindow(QtWidgets.QMainWindow):
                 autosave=True,
             )
         if result == Question.Accepted:
-            log.debug("Accepted")
-            req.req_data["signal"].emit()
-            req.reply(True)
-            return
-        log.debug("Refused")
-        req.reply(False)
+            log.debug("User accepted")
+            on_confirmation()
+        else:
+            log.debug("User declined, skipping new container download")
 
     def handle_startup_error(self, msg: str) -> None:
         self.status_bar.handle_startup_error()
diff --git a/dangerzone/gui/startup.py b/dangerzone/gui/startup.py
index e73192ee..56255a5f 100644
--- a/dangerzone/gui/startup.py
+++ b/dangerzone/gui/startup.py
@@ -123,11 +123,7 @@ class ContainerInstallTask(
             return super().run()
         elif strategy == InstallationStrategy.INSTALL_REMOTE_CONTAINER:
             if Settings().get("updater_ask_before_download"):
-                resp = self.prompt_user()
-                if resp:
-                    self._download_container()
-                else:
-                    log.debug("Skipping new container download (asked by the user)")
+                self.needs_user_input.emit(self._download_container)
             else:
                 self._download_container()
 
@@ -135,11 +131,6 @@ class ContainerInstallTask(
         self.download_container.emit()
         super().run()
 
-    def prompt_user(self) -> PromptRequest:
-        return PromptRequest().ask(
-            self.needs_user_input, data={"signal": self.download_container}
-        )
-
 
 class UpdateCheckTask(
     GUIMixin, startup.UpdateCheckTask, metaclass=_MetaConflictResolver
```

Unfortunately, this moves the download back to the main thread, and blocks the UI while the download is ongoing. We don't want that.

## Emitting a signal rather than using `PromptRequest`

I didn't went further down that road, as it seem it would be similar as reimplementing the `PromptRequest` somehow.

dangerzone/gui/main_window.py
almet: You're right, removed in 58be14677906fa52cadc7939359599c0724c5f25

dangerzone/gui/startup.py
apyrgio: The question was basically if we can move this question to `dangerzone.gui.startup`, instead of prompting the main thread to ask it. But that's minor, we don't have to clarify this now.

dangerzone/gui/main_window.py
apyrgio: Looks good!

apyrgio:
The changes look great Alexis, so feel free to merge when the tests pass.

dangerzone/gui/startup.py
almet: After quick live discussion, we decided that it makes sense to keep it as is, to avoid having different ways of doing things. It's fine having the main thread display the question to the user.

PR #1345, opened 2025-11-19
Install H2Orestart when reproducing container image

apyrgio:


almet:
Thanks. Let's wait to see if we catch some other issues before merging :-)

almet (approved):

PR #1348, opened 2025-11-20
Clear old images when loading a container image from a tarball

apyrgio:


tests/updater/test_installer.py
```diff
@@ -0,0 +1,45 @@
+from unittest.mock import MagicMock
```
almet, line 2: Right now there is no "updater" module, and the `test_installer.py` already exists at the root of the tests. I think these tests should go there instead.

apyrgio: Just to make sure I got it straight, you mean to merge the `tests/test_installer.py` file into `tests/updater/test_installer.py`, right?

almet: Yep, exactly.

PR #1355, opened 2025-11-26
Remove test_large_set.py to fix pytest collection errors

### LINKED ISSUE #1271: Remove test_large_set.py from the tests folder
Currently, a `test_large_set.py` file is present in the test folder, making the tests fail if the large test data is not present. This prevents to run the tests with a simple command like `poetry run pytest` as we're used to do with python programs.

The large test suite is really a great tool, but it's impractical to run on a day to day basis, removing it from this folder and having another way to run it (cf #969) seems a better way to handle these.

```traceback
=========================================================================== ERRORS ===========================================================================
__________________________________________________________ ERROR collecting tests/test_large_set.py __________________________________________________________
tests/test_large_set.py:42: in <module>
    docs_10K = get_test_docs(min_size=0, max_size=10 * 2**10)
tests/test_large_set.py:29: in get_test_docs
    ensure_test_data_exists()
tests/test_large_set.py:25: in ensure_test_data_exists
    sys.exit(1)
E   SystemExit: 1
---------------------------------------------------------------------- Captured stdout -----------------------------------------------------------------------
Test data repository it empty. Skipping large tests.
```

kl2400033266:
The test_large_set.py file causes pytest collection errors when running tests without the large test data repository. Removing this file allows running the standard test suite with 'poetry run pytest' without needing the external test data.

Fixes #1271


almet:
Hi, and thanks for opening the pull request.

As mentioned in the original issue, we might want to have a way to run the large tests, still. I believe this could be done in freedomofpress/dangerzone-image#13 instead.

@apyrgio, what do you think? Okay for you to remove this file?

apyrgio:
I'm really sorry for the delay here, I forgot I was the blocker. I don't think we should remove this file, for two reasons:
1.  You can ignore it with the `--ignore tests/test_large_set.py` option of `pytest`
2. If you run `make test`, it uses the aforementioned switch and ignores it by default.

I'll close this PR since there are better ways around it, but I'll have it in mind for my freedomofpress/dangerzone#1098 PR. Perhaps there's a way to make `pytest` skip this file by default.

almet:
I still think that we should be able to run pytest without having to exclude some files, and so renamed the original issue to point that out, thanks!

Issue #1329, opened 2025-11-09
Improve Conversion Speeds


oregonpillow:
**What is the feature you think should be a good addition to Dangerzone?**

I'm new to the project and only skimmed the code so far, but the conversion of PDF to bitmap pixels and OCR stand out to me as large potential bottlenecks for the slow conversion speeds I observed when running a large test PDF.

It looks like neither pixel conversion or OCR is multi-processed. On small docs, streaming to bitmap is fine, but on large documents (100+ pages) the streaming time could be improved considerably i think.

However, since OCR conversion is more demanding I think this would be the first candidate for optimizing. Was there a specific reason this wasn't previously implemented? Would you be open to this?

**Is your feature request related to a problem? Please describe.**

Slow conversion speeds

almet:
Hi, and thanks for opening an issue. That's a good question!

Right now, what's happening is that the document is streamed to the `stdin` of a process inside the container, where then it's opened and converted to pixels which are sent back to the host using `stdout`.

Then, on the host, the PDF is reconstructed and OCR is optionally run on top.

It could be very interesting to do profiling to know where most of the time is spent.

If data-streaming takes longer than what we want, we might want to parallelize the streaming of pixels from the sandbox to the host. Right now it's sequential. We're currently using stdin and stdout to lower the attack surface of the container: it's not able to read files, or write files on the host, or connect to any network.

If OCR is the bottleneck, doing it in parallel might speed up things, as you mention. We're currently doing the OCR as we receive the data, and just after we reconstruct the PDF page. Don't hesitate if you want to give it a go!

With all that said, I think it would be interesting to have a profiling done (in the container and in the host) to see what are the current bottlenecks.

apyrgio:
I agree with what Alexis said, and I wanted to add some extra observations:
1. There's some expected slowdown due to our isolation layers. On Windows and macOS, we need a Linux VM, so there's a virtualization slowedown. Then, on all platforms, there's gVisor involved, which incurs a performance penalty on system calls. Granted, this affects only the rasterization part, not the reconstruction part.
2. One of the reasons why we don't spawn multiple OCR threads, or convert multiple documents at the same time, is that the Tesseract library spawns multiple threads for OCR under the hood, and these threads use up all the CPU. So, while I do agree that OCR takes a while, I'm not sure if multi-threading is the way to make it faster. It could be that the models we are using are not the fastest, or that we need to tweak Tesseract some other way.

almet:
[...]

PR #1358, opened 2025-11-27
Queue OCR tasks to speedup conversions

almet:
This allows to run the OCR at the same time we do the conversion,
and using all the available CPU.

This results in a x2.4 speedup on my quite beefy machine (AMD Ryzen, 16 CPUs)

Fixes https://github.com/freedomofpress/dangerzone/issues/1329

After this change (`time`, on a 220 pages PDF, with OCR):

```
________________________________________________________
Executed in   44.78 secs    fish           external
   usr time  622.67 secs  608.00 micros  622.66 secs
   sys time    2.94 secs   62.00 micros    2.94 secs

```

Before this change:

```
________________________________________________________
Executed in  274.08 secs    fish           external
   usr time  271.77 secs    1.06 millis  271.77 secs
   sys time    0.79 secs    0.01 millis    0.79 secs
```

### DIFF
--- a/dangerzone/updater/signatures.py
+++ b/dangerzone/updater/signatures.py
@@ -476,7 +476,7 @@ def store_signatures(
     """
     Store signatures locally in the SIGNATURE_PATH folder, like this:
 
-    ~/.config/dangerzone/signatures/
+    ~/.local/share/dangerzone/signatures/
     ├── <pubkey-digest>
     │   ├── <image-digest>.json
     │   ├── <image-digest>.json


apyrgio:
Hey Alexis! I'm looking into your PR, and I have a couple of comments:

1. Your branch does not seem to work on my machine. The error I get is the following, which I believe happens in a worker context:

    ```
    Usage: dangerzone-cli [OPTIONS] [FILENAMES]...
    Try 'dangerzone-cli --help' for help.
    
    Error: Missing argument 'FILENAMES...'
    [INFO ] Running: /usr/bin/podman kill dangerzone-doc-to-pixels-AJSZCc
    [WARNING] Could not kill container 'dangerzone-doc-to-pixels-AJSZCc' within 5 seconds
    ```
    
    I think we have encountered something similar before, I need to look into it though. The good thing is that the tests work, so probably it's something on my side.
2. I'm wondering if we should disable multithreading on Tessearct's side, with `OMP_THREAD_LIMIT=1`.
3. Regarding the worker queue, I also want to check if [`concurrent.futures.ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) offers a nicer interface, maybe even a map-like one.
4. I was trying to understand how we can set the `tessedit_do_invert=0` config value that I mentioned in https://github.com/freedomofpress/dangerzone/issues/1329#issuecomment-3587263765, just to see if it's worth the hassle. Then I realized that we have a divergence in the Tesseract data. On Linux they contain the `tessconfigs` / `configs` file you see in https://github.com/tesseract-ocr/tessdata_fast, but on Windows / macOS we strip them away. So, I'm not sure if this causes any issue.

almet:
I think I found the issue related to your first point.

When `--ocr-lang` is provided, the code creates a `multiprocessing.Pool`. With Python 3.14's forkserver, the worker processes reimport the dangerzone module, and imports the CLI module because the basename is still "dangerzone-cli".

This triggers Click to parse arguments in the worker process, which fails.
Making sure we use the `fork` start method is actually what we want for multiprocessing, as they state in the docs that it doesn't play well with frozen binaries:

> The 'spawn' and 'forkserver' start methods generally cannot be used with “frozen” executables (i.e., binaries produced by packages like PyInstaller and cx_Freeze) on POSIX systems. The 'fork' start method may work if code does not use threads. 
> https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing:~:text=Warning,threads

I'm not sure how it wasn't failing for me at first though!

almet:
I've updated this PR with the following changes:

Progress is now shown after the OCR is done. In previous versions it was showing the progress prior to OCR, and that was not really matching the actual progress.

I'm now using the "spawn" `mp_context`. It required some changes in the architecture of the code I will outline just in the next bullet point. We could use different strategies based on the OS, but as the benefit seem little, I believe it's best to keep with one strategy everywhere.

The `__init__.py` file was doing the CLI argument parsing and acting accordingly, which was problematic because it made it impossible to import some python modules without actually summoning a conversion (which was failing due to the lack of arguments). I actually went ahead and moved the code that's doing the CLI parsing in some specific `run` function (or module).

As a bonus, I've also removed the `dev_script/dangerzone-` scripts, which were doing some path manipulation that is not really needed anymore. The downside of this is that when running as devs, we will have to set the `DANGERZONE_DEV=1` environment variable. It's used in a bunch of places, but mainly that's useful to specify a different location for some resources required to run dangerzone (OCR data and fonts).

Finally, I've made some additional benchmarks, on a mac silicon m1, for a book of 220 pages, the changes are as follow (in both cases, I started the podman machine first, to take its startup time out):

On the branch:

```
________________________________________________________
Executed in   58.76 secs    fish           external
   usr time  300.44 secs    0.47 millis  300.44 secs
   sys time    6.03 secs    1.62 millis    6.03 secs
```

On the main branch:

```
________________________________________________________
Executed in  141.46 secs    fish           external
   usr time  137.44 secs    0.54 millis  137.44 secs
   sys time    2.57 secs    1.97 millis    2.57 secs
```

That's a ~2.4 speedup as well!

Let me know what you think :-)

apyrgio:
Thanks a lot for this PR Alexis. I have tested your PR locally, and I can verify that the speedup I see is x2.5 as well.

... and that makes me :raised_eyebrow:

My test machine has 4 physical cores, 8 logical ones. That's half of what you're seeing in your own setup, and yet we get the same speedup. This means that past a certain point, we get diminishing returns from our parallelization efforts.

That's expected when there's some sort of communication between processes, but we don't have this here. In theory, the OCR tasks should be fully parallelizable. So. what. is. it?

I don't know the answer yet, unfortunately, but I want to write down my assumptions that turned out to be wrong

_(for the following computations, I used the [short version of the Intel manual](https://github.com/user-attachments/files/25366326/intel_sdm_short.pdf), and wrote some code in the `test/queue-ocr-apyrgio` branch)_


## Is it pickling overhead?

Here are some Python stats from the main process:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100    0.002    0.000  110.716    1.107 /usr/lib/python3.13/concurrent/futures/process.py:405(wait_result_broken_or_wakeup)
      250    0.005    0.000  109.620    0.438 /usr/lib/python3.13/multiprocessing/connection.py:1134(wait)
      250    0.002    0.000   50.994    0.204 /usr/lib/python3.13/selectors.py:385(select)
      250   40.045    0.160   50.991    0.204 {method 'poll' of 'select.poll' objects}
    33/26    0.000    0.000   42.888    1.650 /usr/lib/python3.13/threading.py:327(wait)
  299/198    0.004    0.000   42.888    0.217 {method 'acquire' of '_thread.lock' objects}
      2/1    0.002    0.001   42.885   42.885 /usr/lib/python3.13/threading.py:1000(_bootstrap)
   100/56    0.001    0.000   14.309    0.256 /usr/lib/python3.13/multiprocessing/connection.py:182(send_bytes)
   100/56    0.003    0.000   14.296    0.255 /usr/lib/python3.13/multiprocessing/connection.py:406(_send_bytes)
       51    0.002    0.000    2.088    0.041 /home/user/dangerzone/dangerzone/isolation_provider/base.py:240(iter_untrusted_pixels)
```

We do see that `/usr/lib/python3.13/multiprocessing/connection.py:182(send_bytes)` takes 14.3s cumulative time, but most likely that's just the main process blocked while sending bytes to a busy OCR worker. If anything, a bit more down in the stats, we see this line:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)  
       50    0.114    0.002    0.114    0.002 {method 'dump' of '_pickle.Pickler' objects} 
```

So, the pickling overhead is negligible.

## Is it some sort of per-page bootstrapping overhead?

I was not super sure what Python does internally. Maybe an interpreter starts on every request, or PyMuPDF gets bootstrapped per request. For this reason, I've benchmarked the time it takes the worker function to perform OCR. This time duration does not include any Python initializations, pickling, etc., which means that if there's any per-request overhead, this time should be significantly less than the total time for the conversion.

For a single worker, I got these results:

* Total conversion time: **115.9575s**
* Total worker time: **115.6175s**

Again, we don't see any overhead from Python or `multiprocessing` here.

## Maybe the cores are not fully utilized?

I've benchmarked the conversions with perf, to see how the CPU behaves:

### 1 worker

```
# perf stat

         141931.48 msec task-clock:u                     #    1.057 CPUs utilized
                 0      context-switches:u               #    0.000 /sec
                 0      cpu-migrations:u                 #    0.000 /sec
            143846      page-faults:u                    #    1.013 K/sec
      413679360155      cycles:u                         #    2.915 GHz
     1402706370072      instructions:u                   #    3.39  insn per cycle
      129520512139      branches:u                       #  912.557 M/sec
        1833344862      branch-misses:u                  #    1.42% of all branches
                        TopdownL1                 #      7.2 %  tma_backend_bound
                                                  #     60.4 %  tma_bad_speculation
                                                  #      1.1 %  tma_frontend_bound
                                                  #     31.3 %  tma_retiring
                                                  
# perf stat -B -e cache-references,cache-misses,cycles,instructions,branches,faults,migrations

         544375824      cache-references:u
         215207435      cache-misses:u                   #   39.53% of all cache refs
      414974131672      cycles:u
     1403930378813      instructions:u                   #    3.38  insn per cycle
      129775805652      branches:u
            144653      faults:u
                 0      migrations:u
```

### 7 workers

```
# perf stat

         279928.46 msec task-clock:u                     #    5.409 CPUs utilized
                 0      context-switches:u               #    0.000 /sec
                 0      cpu-migrations:u                 #    0.000 /sec
            218115      page-faults:u                    #  779.181 /sec
      603429509398      cycles:u                         #    2.156 GHz
     1401267328027      instructions:u                   #    2.32  insn per cycle
      129204412004      branches:u                       #  461.562 M/sec
        1859287407      branch-misses:u                  #    1.44% of all branches
                        TopdownL1                 #      2.9 %  tma_backend_bound
                                                  #     68.7 %  tma_bad_speculation
                                                  #      0.4 %  tma_frontend_bound
                                                  #     28.0 %  tma_retiring
                                                  
# perf stat -B -e cache-references,cache-misses,cycles,instructions,branches,faults,migrations

         470291011      cache-references:u
         210736557      cache-misses:u                   #   44.81% of all cache refs
      585523062393      cycles:u
     1403278318509      instructions:u                   #    2.40  insn per cycle
      129619940634      branches:u
            223586      faults:u
                 0      migrations:u
```

We can see that `perf` reports 5.4 utilized cores when we use 7 workers. The ratio of bad speculation and cache misses is slightly higher than the single worker version, but it doesn't seem too alarming (I may be wrong though).

## What if we have less workers?

Workers | Total conversion time | Total worker time
------------- | ------------------------------ | -----------------------------
1               | 115.9575                      | 115.6175
2               | 75.6986                        | 143.2468
4               | 47.5609                         | 162.5934
6               | 48.2093                        | 260.7044
7               | 43.2773                         | 272.1581

Something starts to smell fishy past the 4th worker.

## What if we use ThreadPoolExecutor?

While PyMuPDF is thread safe for different objects, it seems that it doesn't release the GIL when performing OCR:

Screenshot From 2026-02-17 10-26-28

almet:
We've had a discussion yesterday about this, here is a summary:

- The original math in the first comment is wrong and misleading. The speedup on my AMD Ryzen machine is of x6, and on a macbook m1 is x2.4.
- I did some more benchmarks, that are summarized in the table below
- We don't want to use all the available CPUs to a void hogging the resources of the machine. We want `workers = max(1, round(mp.cpu_count() / 2))`

The current way things are processed is this:

1. Spawn N workers
2. For each page
   2.1 Retrieve the pixels from the container
   2.2 Send an OCR job to a worker
   2.3 Check if OCR jobs are finished, and in this case process them
   
@apyrgio mentioned two things:

1. We might want to cap the number of pages we store in memory to 2 x workers, in order to avoid filling up the RAM for nothing (one page ~= 8MB, so that would eat 1.5GB RAM in case the OCR is really slow to process). In practice, I don't think it's an issue, but having a cap should not add too much complexity ;
2. We might want deflate to be done in the workers to do it in parallel as well, adding another speedup gain.

---
<details><summary>Benchmarks</summary>
<p>


| Branch | Cores | Wall time (s) | User time (s) | Sys time (s) | Speedup vs main |
|--------|-------|---------------|---------------|--------------|-----------------|
| `main`| -        | 273.49      | 270.88        | 1.15         | 1.0x            |
| `queue-ocr` | 15 | 41.67       | 566.95        | 2.75         | 6.6x |
| `queue-ocr` | 12 | 42.63       | 469.42        | 2.45         | 6.4x |
| `queue-ocr` | 8  | 45.21       | 337.08        | 2.26         | 6.0x |
| `queue-ocr` | 4  | 77.46       | 299.13        | 1.96         | 3.5x |
| `queue-ocr` | 2  | 148.55      | 292.38        | 1.90         | 1.8x |

Also, here is the same thing, but on the `queue-ocr-apyrgio` branch:

| Cores | Wall time (s) | User time (s) | Sys time (s) | Speedup vs main |
|-------|---------------|---------------|--------------|-----------------|
| 16 | 48.65 | 1.96 | 1.44 | 5.6x |
| 12 | 50.36 | 1.86 | 1.54 | 5.4x |
| 8 | 52.95 | 2.01 | 1.66 | 5.2x |
| 4 | 91.07 | 1.98 | 1.55 | 3.0x |

</p>
</details>

almet:
> We might want deflate to be done in the workers to do it in parallel as well, adding another speedup gain.

The page worker does the following call, which contains `compress=True`, and so I believe we are covered. This is what was done previously in the OCR-conversion code path, it just now happens in the worker:

```python
return pixmap.pdfocr_tobytes(
    compress=True,
    language=ocr_lang,
    tessdata=tessdata_dir,
)
```

> We might want to cap the number of pages we store in memory to 2 x workers, in order to avoid filling up the RAM for nothing. In practice, I don't think it's an issue, but having a cap should not add too much complexity

I implemented that by:

- renaming the `collect_ready_futures` function to `drain_ocr_futures`, which now has a way to block until the number of completed futures reaches a certain threshold.
- collecting the futures and blocking the read of the new image buffers while we are still doing the OCR. We accept up to `2 * max_workers` futures, and then wait for the OCR queue to drain to `max_workers`.

It slows a bit the conversion run, but the result is that the maximum resident memory has drained from ~600MB to ~170MB.

apyrgio:
> > We might want deflate to be done in the workers to do it in parallel as well, adding another speedup gain.
> 
> The page worker does the following call, which contains `compress=True`, and so I believe we are covered. 

I'm actually referring to the non-OCR path, i.e., `pixels_to_pdf_page`. In my machine, parallelizing this part has the same performance benefits as parallelizing the OCR one. I've done so in my branch, so feel free to check it out there.

> > We might want to cap the number of pages we store in memory to 2 x workers, in order to avoid filling up the RAM for nothing. In practice, I don't think it's an issue, but having a cap should not add too much complexity
> 
> I implemented that by:
> 
>     * renaming the `collect_ready_futures` function to `drain_ocr_futures`, which now has a way to block until the number of completed futures reaches a certain threshold.
> 
>     * collecting the futures and blocking the read of the new image buffers while we are still doing the OCR. We accept up to `2 * max_workers` futures, and then wait for the OCR queue to drain to `max_workers`.
> 
> 
> It slows a bit the conversion run, but the result is that the maximum resident memory has drained from ~600MB to ~170MB.

Yay for the lower memory usage!

I saw the new code, and I do see the slowdown as well. I think it has to do with the way we pop items from the queue; your code tries to pop as many as possible and then read one page from the stdin. This could result to smaller queue depths if the leftmost jobs take a while. What I did in my branch is to make sure that the queue depth maxes out first, and then process the next item. This makes things slightly faster, but again, that's not important to me.

What's more important to me is to separate the queue/submission logic from the data handling logic. Currently, they are a bit intertwined. For example, some issues I see with `drain_ocr_futures`:

1. It handles the queue logic AND reconstructs the PDF.
2. It's an inline function in a bigger `convert_with_proc` method.
3. It is used in three different places in the code.

I think it's possible to factor out the queue logic in some other method, and keep `convert_with_proc` focused on the data handling part. This will help with debugging this in a few months from now.

apyrgio:
I have opened a [separate issue](https://github.com/freedomofpress/dangerzone/issues/1428) for the things I mentioned here. I propose rebasing this PR, making sure tests pass, and then merge it. We can handle the rest afterwards.

apyrgio:
Thanks a lot, feel free to merge once the tests pass.

PR #1364, opened 2025-12-02
ci: Fix security scans of released images

apyrgio:
We no longer include the container tarball in our releases anymore, since it can change from time to time. Instead, it's best to run security scans against ghcr.io/freedomofpress/dangerzone/v1:latest, which is guaranteed to be the most recently released image.


apyrgio:
Here's a passing run of this security scan: https://github.com/freedomofpress/dangerzone/actions/runs/19855474577/job/56892213614

PR #1370, opened 2025-12-08
Update INSTALL.md with Windows on Arm note

hwine:
Added note about Windows on Arm support status.

Related to #1228


almet:
Thanks!

I've slightly updated your changes. Will merge as soon as the CI is green.

Issue #1379, opened 2025-12-17
Make the CI run for external contributors


almet:
Currently, external contributors cannot run the CI, as it will (rightfully?) fail to upload the images to `freedomofpress/dangerzone`.

Errors can be seen in https://github.com/freedomofpress/dangerzone/pull/1370/checks for instance.

almet:
I've been able to reproduce this specific error [in a smaller repository](https://github.com/almet/external-ghcr/pull/1): it's not possible to have an upstream repo push to a ghcr on behalf of the forker.

Additionally, we don't want to use `pull_request_target` for security reasons.

We've discussed that together with @apyrgio and given that a split between this repo and dangerzone-image will happen soon, we believe the best way forward in the short term is to add a warning in our pull request template, stating that we know this issue and are working on a fix.

In the meantime, it will require us to recreate a branch on the forker's behalf and run the CI this way.

almet:
Just to clarify, the current process I follow is:

- Look at the PR and review it, with a specific eye on security and how the CI could be leveraged to access/exfiltrate some secrets ;
- Checkout the branch locally (with `gh pr checkout ID`)
- Push it to a different branch with `git push origin branch-name:test/branch-name`, which will trigger the CI
- Look if the CI passes, potentially fix some issues by force-pushing to the fork and test branch.
- Merge with `git merge -` from the `main` branch.

PR #1384, opened 2026-01-01
Fix dark mode switch not updating UI mid-session

### LINKED ISSUE #1238: Switching to dark mode in the middle of a session makes the interface ugly
### What happened?

When switching from light mode to dark mode while using Dangerzone leads to a half-themed window, see the attached screenshot:

Image

### operating system version

macOS sequoia 15

### Processor type

Apple Silicon

### Dangerzone version

Development version

### Docker info

```shell

```

### Document conversion logs

```shell

```

### Additional info

_No response_

heathdutton:
When users switch between light and dark mode while Dangerzone is running, the UI now updates dynamically.

Changes:
- Listen to Qt's `paletteChanged` signal to detect theme changes
- Emit a `color_scheme_changed` signal when the OS color mode changes
- Update the `OSColorMode` property and repolish widgets in MainWindow
- Reload theme-specific spinner SVGs in StatusBar and ConversionWidget

Fixes #1238

### DIFF
--- a/tests/gui/test_main_window.py
+++ b/tests/gui/test_main_window.py
@@ -130,31 +130,25 @@ def window(
     window = create_main_window(qtbot, mocker, tmp_path)
     yield window
 
-    # FIXME: There's something in pytest that:
-    #
-    # 1. calls the `.closeEvent()` method of the `MainWindow` fixture, and
-    # 2. does not respect the `.ignore()` method of the event, and closes immediately
-    #
-    # Because we spawn a thread to perform the shutdown tasks, and no one waits this
-    # thread, this chain of events leads to the following behavior:
-    #
-    #   PASSED
-    #
-    #   =============== 1 passed in 3.18s ==================
-    #   QThread: Destroyed while thread '' is still running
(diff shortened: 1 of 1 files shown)


almet:
That is so cool, thanks! That looks to do the trick in a nice way.

We'll return from break next week and will find the time to review and test these changes on different platforms.

almet:
Thanks,

I've checked on some of the systems I have handy: this works on macOS, but not on Linux.

To be fair, on Linux, the detection of the dark theme doesn't work like we want even before these changes, as it seems bound to the state the session was in before starting the application, and so it's not directly related to the changes you're doing here.

As your patch improves the situation for macOS (and probably windows, I will report after I check that), I'm tempted to merge this as is, and open an issue to track dark/light mode themes for Linux separately.

almet:
> and probably windows, I will report after I check that

It also works on the Windows 11 station I used :sunglasses:

We currently have an issue with the CI (see #1379), and so I'll close this PR and open one myself instead to have the CI run properly in the meantime.

PR #1389, opened 2026-01-07
Fix dark mode switch not updating UI mid-session

### LINKED ISSUE #1238: Switching to dark mode in the middle of a session makes the interface ugly

almet:
This is a copy of https://github.com/freedomofpress/dangerzone/pull/1384 from our own branch to have the CI running while #1379 blocks us from having external contributors run the CI for themselves.

I've checked on some of the systems I have handy: this works on macOS and Windows 11, but not on Linux.

To be fair, on Linux, the detection of the dark theme doesn't work like we want even before these changes, as it seems bound to the state the session was in before starting the application, and so it's not directly related to the changes you're doing here.

As the patch improves the situation for macOS (and probably windows, I will report after I check that), I'm tempted to merge as is, and open an issue to track dark/light mode themes for Linux separately. I'm curious if that would float your boat @apyrgio.

Below is a copy of the the text in the original PR ↓

---

When users switch between light and dark mode while Dangerzone is running, the UI now updates dynamically. This is achieved by:

- Listening to Qt's paletteChanged signal to detect theme changes
- Emitting a color_scheme_changed signal when the OS color mode changes
- Updating the OSColorMode property and repolishing widgets in MainWindow
- Reloading theme-specific spinner SVGs in StatusBar and ConversionWidget

Fixes #1238

### DIFF
--- a/tests/gui/test_main_window.py
+++ b/tests/gui/test_main_window.py
@@ -130,31 +130,25 @@ def window(
     window = create_main_window(qtbot, mocker, tmp_path)
     yield window
 
-    # FIXME: There's something in pytest that:
-    #
-    # 1. calls the `.closeEvent()` method of the `MainWindow` fixture, and
-    # 2. does not respect the `.ignore()` method of the event, and closes immediately
-    #
-    # Because we spawn a thread to perform the shutdown tasks, and no one waits this
-    # thread, this chain of events leads to the following behavior:
-    #
-    #   PASSED
-    #
-    #   =============== 1 passed in 3.18s ==================
-    #   QThread: Destroyed while thread '' is still running
-    #   Aborted (core dumped)
-    #
-    # It looks like `QtBot` is the culprit, but I haven't found a better way to affect
-    # it's behavior. In order to circumvent it, we can eagerly wait for the shutdown
-    # thread to finish, which is not pretty nor stable, but works for now.
+    # Prevent shutdown thread creation during cleanup. When pytest-qt closes the
+    # widget, it triggers closeEvent which calls begin_shutdown. By setting
+    # requires_install to False, begin_shutdown exits immediately without
+    # creating the shutdown_thread, avoiding the "QThread: Destroyed while
+    # thread is still running" error.
+    window.dangerzone.isolation_provider.requires_install.return_value = False  # type: ignore [attr-defined]
+
+    # Wait for any running threads to complete
     if hasattr(window, "shutdown_thread"):
         window.shutdown_thread.wait()
+    if hasattr(window, "startup_thread"):
+        window.startup_thread.wait()
 
 
 def test_default_menu(
     qtbot: QtBot, mocker: MockerFixture, tmp_path: pathlib.Path
 ) -> None:
     """Check that the default menu entries are in order."""
+    # Set the setting BEFORE creating the window so the menu action is initialized correctly
     mocker.patch("dangerzone.settings.get_config_dir", return_value=tmp_path)
     settings.Settings().set("updater_check_all", True)
     window = create_main_window(qtbot, mocker, tmp_path)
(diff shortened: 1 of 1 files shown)


dangerzone/gui/__init__.py
```diff
@@ -92,6 +93,8 @@ def monkeypatch_event(arg__1: QtCore.QEvent) -> bool:
         self.os_color_mode = self.infer_os_color_mode()
         log.debug(f"Inferred system color scheme as {self.os_color_mode}")
 
+        self.paletteChanged.connect(self._handle_palette_change)
```
apyrgio: This is a deprecated `Application` member since 6.0: https://doc.qt.io/qt-6/qguiapplication-obsolete.html#paletteChanged

The suggestion is to handle the `QEvent.Type.ApplicationPaletteChange` (see https://doc.qt.io/qtforpython-6/PySide6/QtCore/QEvent.html).

apyrgio:
As a general comment, I'm worried about the multiple handlers for the palette change, and the duplicated code for loading the spinner and polishing/unpolishing the Qt elements (it still baffles me that there's no better way for Qt to handle that).

In any case, that's just Qt being Qt, and if we agree that we will refactor our UI in the mid-future, then a change like this should be fine, even if difficult to maintain.

To conclude, once we replace the deprecated signal, I'm fine with merging this PR.

dangerzone/gui/__init__.py
almet: I've been thinking the same thing when looking at this code, but as we're not uniformly using Qt6 in our apps, and we plan on keeping it this way for the forseable future, thought it's fine to keep this way to avoid having multiple code paths for the same feature without much benefit.

… Then I found out that Pyside2 provides and recommend this signal as well, and implemented the change in 00d0420d1dea8a8ab343ddf756d6ed887a836a62.

apyrgio (approved):

apyrgio:
Great, feel free to merge once the CI works.

.github/workflows/build-push-image.yml
```diff
@@ -213,6 +213,12 @@ jobs:
           # NOTE: The following command does NOT actually add annotations to the root manifest yet.
           # See https://github.com/freedomofpress/dangerzone/issues/1270#issuecomment-3382134379
           docker buildx imagetools create \
+              --annotation "manifest-descriptor:rocks.dangerzone.debian_archive_date=${{ needs.build.outputs.debian_archive_date }}" \
+              -t ${{ needs.build.outputs.image }} \
+              ${DIGESTS} || \
+          # In case of failures, try to --append instead \
```
apyrgio, line 219: Question, why can't we always `--append`?

tests/gui/test_main_window.py
```diff
@@ -184,6 +177,14 @@ def test_default_menu(
...
     assert window.dangerzone.settings.get("updater_remote_log_index") == 1000
 
+    # Prevent shutdown thread creation during cleanup (see window fixture for details)
+    window.dangerzone.isolation_provider.requires_install.return_value = False  # type: ignore [attr-defined]
+
+    if hasattr(window, "shutdown_thread"):
+        window.shutdown_thread.wait()
+    if hasattr(window, "startup_thread"):
+        window.startup_thread.wait()
+
```
apyrgio: Should we use the `window` fixture instead here?

.github/workflows/build-push-image.yml
almet: Append needs to have something to append to, per my understanding.

tests/gui/test_main_window.py
almet: If it's only for these two tests, I'm okay getting along with this duplication. Let's revisit if we need this more in the tests at some point? My main reason for not changing this is that the tests take forever to run and we have a PR that passes the tests right now, and so I'm tempted to merge as is.

apyrgio: Well, if we replace this code with the `window` fixture and it doesn't work, then I don't mind adding a note and saying that there's duplication. But if it does work, then that's a simple change that can happen while squashing commits.

.github/workflows/build-push-image.yml
apyrgio: Got it, thanks.

tests/gui/test_main_window.py
almet: Did so in a fixup commit (that I squashed in there), let's wait for the CI now 🫣

almet: We actually can't directly use the fixture here because we need to create the window **after** the settings have been changed, as they have an incidence on what's displayed.

I added a comment mentioning this.

apyrgio: Thanks a lot Alexis, I've approved the changes again.

PR #1390, opened 2026-01-08
Update the grype ignore list

almet:
Add a few CVEs that doesn't affect us. A new container image will be out soon nevertheless.

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -152,3 +152,20 @@ ignore:
   # Clients might be, but it's orthogonal to what we can do to solve this
   # issue.
   - vulnerability: CVE-2025-68973
+
+  # CVE-2025-68972
+  # ##############
+  #
+  # In GnuPG through 2.4.8, if a signed message has \f at the end of a
+  # plaintext line, an adversary can construct a modified message that places
+  # additional text after the signed material, such that signature verification
+  # of the modified message succeeds (although an "invalid armor" message is
+  # printed during verification). This is related to use of \f as a marker to
+  # denote truncation of a long plaintext line.
+  #
+  # Debian Tracker: https://security-tracker.debian.org/tracker/CVE-2025-68972
+  #
+  # Verdict: We do not use GnuPG or dirmngr in a way that is problematic
+  # related to this issue (e.g. we do not trust the text output in a way), and
+  # no fixes have been provided by upstream so far.
+  - vulnerability: CVE-2025-68972


.grype.yaml
```diff
@@ -119,3 +120,78 @@ ignore:
...
+  # a malicious server to cause the client to read large amounts of data into
+  # memory, potentially causing OOM or other DoS.
+  #
+  # Debian tracker: https://security-tracker.debian.org/tracker/CVE-2025-13836
+  #
+  # Verdict: we are not affected because we don't use any HTTP server
+  # capabilities inside the container. Outside the container, this could lead
+  # to OOM attacks when looking for updates on the client in case of Github
+  # servers compromise, which we deem acceptable until the next client release
+  # (0.11.0)
```
apyrgio:
```suggestion
  # (0.11.0). Moreover, the Debian Security team deems it as a minor issue for Trixie, and there's currently no fix out for it.
```

.grype.yaml
```diff
@@ -119,3 +120,78 @@ ignore:
...
+  # * https://security-tracker.debian.org/tracker/CVE-2025-66293
+  #
+  # Verdict: The bug is in libpng simplified API¹, which is not in use by
+  # libreoffice, which uses the "normal" API (the simplified API, which starts
+  # with `png_image_and` is not present in the libreoffice code²).
+  #
+  # ¹ https://github.com/pnggroup/libpng/blob/0f07f70aa55e4771d00b0f8a238dca992fbb60e4/manuals/libpng-manual.txt
+  # ² https://github.com/LibreOffice/core/blob/22beba9c9ff17292ee956985fd3eea208231d6af/vcl/source/filter/png/PngImageReader.cxx#L750
+  - vulnerability: CVE-2025-65018
+  - vulnerability: CVE-2025-66293
```
apyrgio: Given that this is a High severity bug with a fix, I think we should not add it in our ignore list for two reasons:
1. It won't affect our current security scans, which error out on Critical CVEs.
2. We will produce a new image out soon, and we should make sure that this CVE is not there. If for some unexpected reason it's still there, the ignore list will mask it.

Same reasoning applies to the following CVE.

.grype.yaml
almet: Just did so, thanks.

.grype.yaml
almet: I removed the libpng CVEs which are fixed [by our just released image](https://github.com/freedomofpress/ghcr-signer/pull/29), but let the ones related to GPG.

We're also discussing in parallel, so let's see how that goes!

apyrgio (approved):

apyrgio:
Thanks for the changes Alexis, feel free to merge.

PR #1391, opened 2026-01-08
Fix default PDF reader selection on Linux

### LINKED ISSUE #814: Detect default application for opening PDFs
Dangerzone has some logic for detecting which application (`*.desktop` file) can open a PDF:

https://github.com/freedomofpress/dangerzone/blob/65776d8c0506535c20f74f4375193287e1e8a78b/dangerzone/gui/logic.py#L97-L123

The result on Tails 6.2 is the following application list:

image

However, what's missing here is to present to the user the default application for opening PDFs. Instead, the first choice is picked up at random. Here, it's GIMP, whereas it should be Document Viewer.

almet:
`xdg-mime` outputs `\n`, causing the default PDF reader to not be recognized.

Fixes #814

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## [Unreleased](https://github.com/freedomofpress/dangerzone/compare/v0.9.0...HEAD)
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


apyrgio (approved):

apyrgio:
Nice one, feel free to merge once the tests pass.

Issue #1069, opened 2025-01-30
Provide a slim version of dangerzone


almet:
Once we will have independent container updates, we will be able to distribute dangerzone with and without container images bundled, making it possible for users to chose what they want to download.

On Linux, the provided package will probably always be the slim one, with dependencies on the container images.

almet:
At least the Linux packages could by default not ship the `container.tar` file, making it so that we end up with `dangerzone` and `dangerzone-full` packages (instead of `-slim`).

This will require changes in how we prompt the user about the updates, because if there is no `container.tar`, we actually need to reach to the container registry (at least once) to get it.

On Linux, we could display a message stating that it's possible to use `-full` variants in case the user doesn't want to enable the updates.

On Windows and macOS, we have been thinking about what the "slim" version would be, because it might also mean that we don't want to distribute the Podman binaries, for instance. I believe this should be covered by a separate issue to keep the progress, and so created #1407.

PR #1400, opened 2026-01-15
Make container.tar optional

almet:
This removes the need to have a `container.tar` file packaged with Dangerzone.

Related to #1069, but is not enough to close it for now, as a `dangerzone-slim` version will be required at some point, which will probably need an update of our Debian and Fedora packages as well.

### DIFF
--- a/dangerzone/updater/signatures.py
+++ b/dangerzone/updater/signatures.py
@@ -19,6 +19,7 @@ from .. import container_utils as runtime
 from .. import errors as dzerrors
 from ..util import get_resource_path, subprocess_run
 from . import cosign, errors, log, registry
+from .log_index import LAST_KNOWN_LOG_INDEX
 
 try:
     import platformdirs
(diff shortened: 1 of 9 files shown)


dangerzone/updater/signatures.py
```diff
@@ -33,9 +33,20 @@ def appdata_dir() -> Path:
...
+    """
+    return get_resource_path("container.tar").exists()
+
+
 # RELEASE: Bump this value to the log index of the latest signature
 # to ensure the software can't upgrade to container images that predates it.
-BUNDLED_LOG_INDEX = 732661252
+# If no container.tar is bundled, this is set to 0 so that the installer
+# falls back to remote container installation.
+BUNDLED_LOG_INDEX = 732661252 if is_container_tar_bundled() else 0
```
apyrgio: I get what you're trying to do here, but unfortunately this nerfs our protection against being instructed to install older versions of the image.

What we should do instead is:
* Rename `BUNDLED_LOG_INDEX` to `LAST_KNOWN_LOG_INDEX`.
* Move this constant in a separate file (optional, mostly so that we can script the update of this value when we build a new container image)
* In `updater/installer.py`:
   - Construct a new `bundled_log_index` variable that is effectively `LAST_KNOWN_LOG_INDEX if is_container_tar_bundled() else 0`
   - Proceed with this value for the rest of the checks.

This way, we will both have this value as the lowest log index that we can accept, but also dismiss it from our update checks if the `container.tar` file is not present.

almet: Thanks, that makes sense, and I like the idea of having the log index in a separate file as well. I implemented all that and push/squashed the commits. Let me know if that's better from your point of view!

apyrgio: \m/

apyrgio:
Nice, feel free to merge!

PR #1406, opened 2026-01-22
docs: Update copyright notice formatting in README.md

paulschreiber:
## Summary
Update copyright notice formatting in README.md
• Use © instead of (c). The latter has no legal meaning.
• Use en dashes to separate time periods.


almet (approved):
Thanks Paul! (Merging as-is without running the CI as it's currently broken for external contributors #1379 and is "just" a change in the README)

Issue #1404, opened 2026-01-21
Do not ship dangerzone-machine on Linux


almet:
The `dangerzone-machine` utility is only useful on Windows and macOS, where we manage the Podman Linux virtual machine. Linux builds (RPMs and DEB files) should not contain this utility.

PR #1412, opened 2026-01-26
Exclude dangerzone-machine from Linux packages

NielDuysters:
Resolves: https://github.com/freedomofpress/dangerzone/issues/1404

I decided to make a minimal change on the level of the package builders themselves instead of more upstream in e.g `setup.py`. This to avoid breaking other build-flows and development ergonomics.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -10,4 +10,5 @@ export DH_VERBOSE=1
 
 override_dh_builddeb:
 	./install/linux/debian-vendor-pymupdf.py --dest debian/dangerzone/usr/lib/python3/dist-packages/dangerzone/vendor/
+	rm -f debian/dangerzone/usr/bin/dangerzone-machine
 	dh_builddeb $@
--- a/install/linux/dangerzone.spec
+++ b/install/linux/dangerzone.spec
@@ -225,6 +225,7 @@ convert the documents within a secure sandbox.
 %install
 %pyproject_install
 %pyproject_save_files dangerzone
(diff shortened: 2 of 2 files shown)


apyrgio:
(argh, we need to fix our CI for contributors soon :grimacing: )

Looks good to me Niel. I'll test your changes in a separate branch just to make sure our CI is green. Out of curiosity, did you encounter any issue with this binary in your Linux installation or something? To be clear, I agree that we should remove it because it's redundant.

apyrgio:
The Debian/Fedora tests pass: https://github.com/freedomofpress/dangerzone/actions/runs/21364285122/job/61492503082

Some Windows tests are failing, and Intel macOS is in progress, but that's part of our CI's flakiness.

NielDuysters:
> (argh, we need to fix our CI for contributors soon 😬 )
> 
> Looks good to me Niel. I'll test your changes in a separate branch just to make sure our CI is green. Out of curiosity, did you encounter any issue with this binary in your Linux installation or something? To be clear, I agree that we should remove it because it's redundant.

I did not do an extensive test of the binary, but it seemed to run fine.

almet:
To answer to your questions Niel (even if you redacted them ;)): the dangerzone-machine CLI command is only useful on Windows and macOS to initialize the Virtual Machine used by Podman.

On Linux, since Podman doesn't require a VM, it's not useful and we should avoid shipping it. Thanks for the code changes!

Issue #1413, opened 2026-01-26
Look for Desktop files in XDG_DATA_DIR


almet:
It could make sense to look at Desktop files located in the XDG_DATA_DIR, rather than using hardcoded locations, on Linux.

See https://github.com/freedomofpress/dangerzone/discussions/1410 for more info.

PR #1414, opened 2026-01-29
Look for desktop files in XDG_DATA_DIRS

NielDuysters:
Resolves https://github.com/freedomofpress/dangerzone/issues/1413

In this change I split the contents of the `XDG_DATA_DIRS` environment variable and concat it with `~/.local/share`. I use `dict.fromkeys` to remove duplicates, otherwise we would potentially search the same directory affecting performance. If `XDG_DATA_DIRS` is empty we default to the paths specified before this change avoiding regression and matching the [expected behavior](https://wiki.archlinux.org/title/XDG_Base_Directory#System_directories).

I tested it by installing indiPDF through flatpack. After this change indiPDF was listed as option:
image

When unsetting `XDG_DATA_DIRS` the same options as before this change are listed. This fixes the issue raised here https://github.com/freedomofpress/dangerzone/discussions/1410

### DIFF
--- a/dangerzone/gui/logic.py
+++ b/dangerzone/gui/logic.py
@@ -119,11 +119,16 @@ class DangerzoneGui(DangerzoneCore):
                 log.debug(f"xdg-mime query failed: {e}")
 
             # Find all .desktop files
-            for search_path in [
-                "/usr/share/applications",
-                "/usr/local/share/applications",
-                os.path.expanduser("~/.local/share/applications"),
-            ]:
+            # Use dict.fromkeys (rather than a set) to retain paths order
+            # (only dict keys are used, values are set to `None`)
+            for search_path in dict.fromkeys(
+                os.environ.get(
+                    "XDG_DATA_DIRS",
+                    "/usr/local/share:/usr/share",
+                ).split(":")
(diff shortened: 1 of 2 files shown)


dangerzone/gui/logic.py
```diff
@@ -119,11 +119,14 @@ def _find_pdf_viewers(self) -> OrderedDict[str, str]:
                 log.debug(f"xdg-mime query failed: {e}")
 
             # Find all .desktop files
-            for search_path in [
-                "/usr/share/applications",
-                "/usr/local/share/applications",
-                os.path.expanduser("~/.local/share/applications"),
-            ]:
+            for search_path in dict.fromkeys(
```
almet, line 124: Thanks! I would add a comment here explaining why `dict.fromkeys` is used in place of `set()` (to keep the order in which we walk the paths)


```suggestion
            # Uses dict.fromkeys (rather than a set) to retain paths order.
            # Only dict keys are used, values are set to `None`.
            for search_path in dict.fromkeys(
```

almet:
(Just merged, the CI was passing excepting on macOS which was still pending, but as this code path is not used there, we're fine). Thanks again Niel!

PR #1423, opened 2026-02-11
Provide `*-full` packages for Linux, and remove `container.tar` from default ones

### LINKED ISSUE #1069: Provide a slim version of dangerzone

almet:
This removes the container.tar from the Linux packages, providing `-full` variants for the versions with the container.

This doesn't currently change how windows and macOS installers are created. Currently they continue shipping with the container.tar inside.

Fixes #1069

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
           poetry run dangerzone-image prepare-archive
           --output share/container.tar
 
-      - name: Build Dangerzone .deb
+      - name: Build Dangerzone .deb packages
         run: |
           ./dev_scripts/env.py --distro ${{ matrix.distro }} \
               --version ${{ matrix.version }} \
@@ -248,8 +248,17 @@ jobs:
           if-no-files-found: error
           compression-level: 0
 
+      - name: Upload Dangerzone-full .deb
+        if: matrix.distro == 'debian' && matrix.version == 'bookworm'
+        uses: actions/upload-artifact@v6
+        with:
+          name: dangerzone-full.deb
+          path: "deb_dist/dangerzone-full_*_*.deb"
+          if-no-files-found: error
+          compression-level: 0
+
   install-deb:
-    name: "install-deb (${{ matrix.distro }} ${{ matrix.version }})"
+    name: "install-deb (${{ matrix.distro }} ${{ matrix.version }} ${{ matrix.package }})"
     runs-on: ubuntu-latest
     needs:
       - build-deb
@@ -259,16 +268,28 @@ jobs:
         include:
           - distro: ubuntu
             version: "22.04"
+            package: "dangerzone.deb"
           - distro: ubuntu
             version: "24.04"
+            package: "dangerzone.deb"
           - distro: ubuntu
             version: "26.04"
+            package: "dangerzone.deb"
           - distro: ubuntu
             version: "25.10"
+            package: "dangerzone.deb"
           - distro: debian
             version: bookworm
+            package: "dangerzone.deb"
+          - distro: debian
+            version: bookworm
+            package: "dangerzone.deb"
+          - distro: debian
+            version: bookworm
+            package: "dangerzone-full.deb"
           - distro: debian
             version: trixie
+            package: "dangerzone.deb"
 
     steps:
       - name: Checkout
@@ -281,7 +302,7 @@ jobs:
       - name: Download Dangerzone .deb
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
-          name: dangerzone.deb
+          name: ${{ matrix.package }}
           path: "deb_dist/"
 
       - name: Build end-user environment
@@ -289,6 +310,13 @@ jobs:
           ./dev_scripts/env.py --distro ${{ matrix.distro }} \
               --version ${{ matrix.version }} \
               build
+
+      - if: matrix.package == 'dangerzone-full.deb'
+        name: Download container image
+        run: |
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} \
+              --version ${{ matrix.version }} \
+              run dangerzone-image upgrade
 
       - name: Run a test command
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,6 +309,7 @@ jobs:
       matrix:
         distro: ["fedora"]
         version: ["42", "43", "44"]
+        full: [true, false]
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -351,31 +352,48 @@ jobs:
         id: image-digest
         run: echo "digest=$(crane digest $(cat share/image-name.txt):latest)" >> $GITHUB_OUTPUT
 
-      - name: Cache container image archive
+      - if: matrix.full
+        name: Cache container image archive
         id: cache-container
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: share/container.tar
           key: v1-container-tar-${{ runner.arch }}-${{ steps.image-digest.outputs.digest }}
 
-      - name: Download the container image as a tar archive
-        if: steps.cache-container.outputs.cache-hit != 'true'
+      - if: matrix.full and steps.cache-container.outputs.cache-hit != 'true'
+        name: Download the container image as a tar archive
         env:
           DANGERZONE_DEV: "1"
         run: >-
-          poetry run dangerzone-image prepare-archive
-          --output share/container.tar
+          poetry run dangerzone-image prepare-archive --output share/container.tar
 
-      - name: Build Dangerzone .rpm
+      - if: not matrix.full
+        name: Build dangerzone(slim).rpm
         run: |
           ./dev_scripts/env.py --distro ${{ matrix.distro }} --version ${{ matrix.version }} \
               run --dev --no-gui ./dangerzone/install/linux/build-rpm.py
 
-      - name: Upload Dangerzone .rpm
+      - if: matrix.full
+        name: Build dangerzone-full.rpm
+        run: |
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} --version ${{ matrix.version }} \
+              run --dev --no-gui ./dangerzone/install/linux/build-rpm.py --full
+
+      - if: not matrix.full
+        name: Upload dangerzone.rpm
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: dangerzone-${{ matrix.distro }}-${{ matrix.version }}.rpm
           path: "dist/dangerzone-*.x86_64.rpm"
+          if-no-files-found: error
+          compression-level: 0
+
+      - if: matrix.full
+        name: Upload dangerzone-full.rpm
+        uses: actions/upload-artifact@v7
+        with:
+          name: dangerzone-full-${{ matrix.distro }}-${{ matrix.version }}.rpm
+          path: "dist/dangerzone-full-*.x86_64.rpm"
           if-no-files-found: error
           compression-level: 0
 
@@ -385,11 +403,25 @@ jobs:
       - name: Reclaim some storage space
         run: podman system reset -f
 
-      - name: Build end-user environment
+      - if: not matrix.full
+        name: Build (slim) end-user environment
         run: |
           ./dev_scripts/env.py --distro ${{ matrix.distro }} \
               --version ${{ matrix.version }} \
               build
+
+      - if: matrix.full
+        name: Build (full) end-user environment
+        run: |
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} \
+              --version ${{ matrix.version }} \
+              build --full
+
+      - if: not matrix.full
+        name: Download container image
+        run: |
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} --version ${{ matrix.version }} \
+              run dangerzone-image upgrade
 
       - name: Run a test command
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,6 +330,50 @@ jobs:
               --version ${{ matrix.version }} \
               run dangerzone --help
 
+
+  install-deb-full:
+    name: "install-deb-full (${{ matrix.distro }} ${{ matrix.version }})"
+    runs-on: ubuntu-latest
+    needs:
+      - build-deb
+    strategy:
+      matrix:
+        include:
+          - distro: debian
+            version: bookworm
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+
+      - name: Download Dangerzone full .deb
+        uses: actions/download-artifact@v6
+        with:
+          name: dangerzone-full.deb
+          path: "deb_dist/"
+
+      - name: Build end-user environment with full package
+        run: |
+          ./dev_scripts/env.py --distro ${{ matrix.distro }} \
+              --version ${{ matrix.version }} \
+              build --full
+
+      - name: Run a test command (container bundled)
+        run: |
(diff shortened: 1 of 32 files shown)


dangerzone/gui/updater.py
```diff
@@ -33,6 +33,23 @@
 OK_TEXT = "Yes, enable sandbox updates"
 CANCEL_TEXT = "No, disable sandbox updates"
 
+MSG_CONFIRM_DOWNLOAD_CONTAINER = """\
+<p>
+    <b>Enable sandbox download?</b>
```
apyrgio:
```suggestion
    <b>Download conversion sandbox?</b>
```

dangerzone/gui/updater.py
```diff
@@ -33,6 +33,23 @@
...
+</p>
+
+<p>Dangerzone needs to download the sandbox from the internet to convert documents.</p>
+
+<p>If you enable this option, Dangerzone will download the sandbox now and
+periodically check for updates.</p>
+<p>This is <b>required</b> for Dangerzone to work. If you prefer an offline install,
+you can download the full version with an embedded container.</p>
+
+<p>If you need to run Dangerzone in an air-gapped environment, see <a href="https://github.com/freedomofpress/dangerzone/blob/main/docs/developer/independent-container-updates.md#installing-image-updates-to-air-gapped-environments">our documentation</a>.</p>
```
apyrgio: What's the difference between an offline install and an airgapped install? Can we merge these somehow?

dangerzone/updater/releases.py
```diff
@@ -155,10 +157,19 @@ def should_check_for_updates(settings: Settings) -> bool:
     if settings.get("updater_last_check") is None:
         log.debug("Dangerzone is running for the first time, updates are stalled")
         settings.set("updater_last_check", 0, autosave=True)
+        # If no container is bundled and none is installed, we must prompt the user
+        # to enable updates immediately, otherwise Dangerzone won't work.
+        if not is_container_tar_bundled() and not LAST_LOG_INDEX.exists():
```
apyrgio: What happens if last log index exists, but the Podman storage was erased in the meantime?

dangerzone/cli.py
```diff
@@ -144,7 +144,19 @@ def cli_main(
...
+            click.echo(
+                "\n"
+                + Fore.RED
+                + Style.BRIGHT
+                + "No container image found."
+                + Style.RESET_ALL
+                + " Please initialize Dangerzone by running:\n\n"
+                "    dangerzone-image upgrade\n"
+            )
+            sys.exit(1)
```
apyrgio, line 159: If we don't offer a way to auto-download the image via the CLI, then this should be recorded as a breaking change, at least for new installations.

install/linux/build-deb.py
```diff
@@ -20,6 +22,60 @@ def run(cmd):
...
+    rules_orig = rules_path.read_text()
+
+    control_new = (
+        control_orig.replace("Source: dangerzone", "Source: dangerzone-full")
+        .replace("Package: dangerzone", "Package: dangerzone-full")
+        .replace(
+            "Conflicts: dangerzone-full, dangerzone-qubes",
+            "Conflicts: dangerzone, dangerzone-qubes",
+        )
+    )
```
apyrgio: Isn't there a way to have two control files for separate packages? I haven't dug into it yet, just asking.

dangerzone/cli.py
almet: Good call, and even more: we'll need to record this as a breaking change in any case, unless we decide it's an opt-out feature.

I was thinking about providing an `--upgrade` option to the cli, but having `dangerzone-image upgrade` might be better? Any thoughts?

apyrgio: Hm, no clue if one or the other is more useful. I'd keep them separate for now (so just use `dangerzone-image upgrade`) and check for user requests.

almet: I've added a mention in the changelog about that as well  0c45f9fdef21b686c4eef3f0707753ed55e9a706

dangerzone/updater/releases.py
almet: Good catch! I'm now checking for the image presence, in 8f91d632ad3fea50aab3b1c9d0bc75956e6db63f

dangerzone/gui/updater.py
almet: I couldn't find what this comment applies to, unfortunately. I've reworded these messages and also reworked how and when they show up. I think we can consider this one done, let me know if not!

.github/workflows/ci.yml
```diff
@@ -305,6 +331,50 @@ jobs:
               --version ${{ matrix.version }} \
               run dangerzone --help
 
+
+  install-deb-full:
```
apyrgio: I'm a bit worried of the repetition here between the "full" and regular variants. I actually think that without complicating things too much, we can merge this logic into the `install-deb` job with the following changes:

1. In the matrix field of `install-deb`, add a `full: True|False`. For Debian Bookworm, we can have two entries:

```
- distro: debian
  version: bookworm
  variant: ""
- distro: debian
  version: bookworm
  variant: "full"
```

2. Use `${{ matrix.variant }}` only on the steps that currently have the word "full".

.github/workflows/ci.yml
```diff
@@ -404,6 +494,49 @@ jobs:
           ./dev_scripts/env.py --distro ${{ matrix.distro }} --version ${{ matrix.version }} \
               run dangerzone --help
 
+  install-rpm-full:
```
apyrgio: What I mentioned in https://github.com/freedomofpress/dangerzone/pull/1423/changes#r3219532167 also applies here, with the added benefit that:
1. This job will no longer need to wait for `build-install-rpm` to finish.
2. Having a `variant: "full"` option in the matrix will help build just the packages we need per job, which means we will build the full variant for just one of our the Fedora versions (Fedora 43 works fine).

dangerzone/gui/updater.py
```diff
@@ -33,6 +33,22 @@
...
+
+<p>Welcome to Dangerzone, in order to convert documents, it needs to download a sandbox from the internet.</p>
+
+<p>If you enable this option, Dangerzone will download the sandbox now and
+periodically check for updates.</p>
+<p>This is <b>required</b> for this version of Dangerzone to work.</p>
+
+<p>If you need to run Dangerzone in an air-gapped environment, see <a href="https://github.com/freedomofpress/dangerzone/blob/main/docs/developer/independent-container-updates.md#installing-image-updates-to-air-gapped-environments">our documentation</a>.</p>
+"""
+OK_TEXT_DOWNLOAD = "Yes, download sandbox"
```
apyrgio:
```suggestion
OK_TEXT_DOWNLOAD = "Yes, download sandbox and enable updates"
```

debian/control
```diff
@@ -9,8 +9,19 @@ Rules-Requires-Root: no
 
 Package: dangerzone
 Architecture: any
+Conflicts: dangerzone-full, dangerzone-qubes
```
apyrgio: Is there any `dangerzone-qubes` package though for Debian?

dangerzone/gui/main_window.py
```diff
@@ -685,10 +688,30 @@ def handle_update_check_completed(self) -> None:
...
+
+        This is a blocking prompt - the user must accept or the app will shut down.
+        """
+        check = prompt_for_checks(self.dangerzone, download_required=True)
+        if check is True:
+            req.reply(True)
+        else:
+            # User declined or pressed X - shutdown
+            req.reply(False)
+            QtCore.QTimer.singleShot(0, self._handle_download_declined)
```
apyrgio, line 709: Why did we use this method of exiting the app versus `self.begin_shutdown(ret=2)`, which is used by the other tasks? I recall that there were some weird Python tracebacks if we didn't exit gracefully.

dangerzone/gui/updater.py
```diff
@@ -33,6 +33,22 @@
...
+    <b>Download conversion sandbox?</b>
+</p>
+
+<p>Welcome to Dangerzone, in order to convert documents, it needs to download a sandbox from the internet.</p>
+
+<p>If you enable this option, Dangerzone will download the sandbox now and
+periodically check for updates.</p>
+<p>This is <b>required</b> for this version of Dangerzone to work.</p>
+
+<p>If you need to run Dangerzone in an air-gapped environment, see <a href="https://github.com/freedomofpress/dangerzone/blob/main/docs/developer/independent-container-updates.md#installing-image-updates-to-air-gapped-environments">our documentation</a>.</p>
```
apyrgio: The link points to our updates instructions, but I'm not sure if we want to ask the user to go through those. I think we need a section that suggests people to install the `-full` variant, and explain that this is already the case for Windows/macOS, but not Linux.

dangerzone/updater/releases.py
```diff
@@ -155,10 +157,19 @@ def should_check_for_updates(settings: Settings) -> bool:
     if settings.get("updater_last_check") is None:
         log.debug("Dangerzone is running for the first time, updates are stalled")
         settings.set("updater_last_check", 0, autosave=True)
+        # If no container is bundled and none is installed, we must prompt the user
+        # to enable updates immediately, otherwise Dangerzone won't work.
+        if not is_container_tar_bundled() and not is_container_image_installed():
```
apyrgio, line 162: In case where we have an **existing** user that:
* Downloads the slim version of Dangerzone (most likely a Linux user)
* Does not have the Dangerzone image installed locally (maybe because they cleared their Podman storage)
* Has indicated that they don't wish to get security updates

shouldn't we prompt them one more time with our new message, instead of letting Dangerzone crash down the line? Alternatively, we could just raise `errors.UpdaterDisabledNoContainer`, right?

dangerzone/startup.py
```diff
@@ -228,9 +243,29 @@ def should_skip(self) -> bool:
...
+                    return False
+                elif accepted is False:
+                    # User declined - raise an error to stop startup
+                    raise errors.UpdaterDisabledNoContainer()
+                # User pressed X - treat as decline
+                raise errors.UpdaterDisabledNoContainer()
+            else:
+                # Container available: non-blocking prompt, handler saves setting
+                # Skip update check, user will be prompted again next run if needed
+                return True
```
apyrgio: (Optional, solely because of the extra level of indentation which made things a bit more confusing to me)

```suggestion
        except updater_errors.NeedUserInput:
            self.prompt_user()
            return True
        except updater_errors.NeedUserInputNoContainer as e:
                if self.prompt_user(download_required=True) is True:
                    settings.Settings().set("updater_check_all", True, autosave=True)
                    # Accepting here is consent for this download, so suppress
                    # ContainerInstallTask's redundant prompt for this run.
                    ContainerInstallTask.skip_prompt_once = True
                    # Proceed with update check immediately so the remote
                    # container can be downloaded.
                    return False
                    
                # User declined or pressed X - raise an error to stop startup
                raise errors.UpdaterDisabledNoContainer()
```

dangerzone/startup.py
```diff
@@ -228,9 +243,29 @@ def should_skip(self) -> bool:
...
+            download_required = isinstance(e, updater_errors.NeedUserInputNoContainer)
+            accepted = self.prompt_user(download_required=download_required)
+
+            if download_required:
+                # No container available: blocking prompt, handle response
+                if accepted is True:
+                    settings.Settings().set("updater_check_all", True, autosave=True)
+                    # Accepting here is consent for this download, so suppress
+                    # ContainerInstallTask's redundant prompt for this run.
+                    ContainerInstallTask.skip_prompt_once = True
```
apyrgio: If I understand correctly, here we change the `skip_prompt_once` attribute of the whole `ContainerInstallTask` class, in order to affect the subsequent of the execution of that task?

Hm, messing with global state is a bit risky, I'd like to avoid this if possible. I think there's a more elegant way to do that though. I'll comment in the relevant task.

dangerzone/gui/startup.py
```diff
@@ -123,7 +123,11 @@ def run(self) -> None:
             self.load_container.emit()
             return super().run()
         elif strategy == InstallationStrategy.INSTALL_REMOTE_CONTAINER:
-            if Settings().get("updater_ask_before_download"):
+            ask = Settings().get("updater_ask_before_download")
+            if startup.ContainerInstallTask.skip_prompt_once:
+                startup.ContainerInstallTask.skip_prompt_once = False
+                ask = False
+            if ask:
```
apyrgio: Could we rely on this check instead?

```suggestion
            ask = Settings().get("updater_ask_before_download")
            can_proceed = is_container_tar_bundled() or is_container_image_installed()
            if ask and can_proceed:
```

install/windows/build-app.py
apyrgio, line 1: Does this file replace `build-app.bat`, or is it an experiment?

.github/workflows/ci.yml
almet: Thanks, I agree. Let's keep things together. I've done so in 26f92cc6945718979b98d1a4ed134b46d0b53bb0 (hoping that this passes the CI!)

.github/workflows/ci.yml
almet: I'm actually not sure we want this here. For Debian, because the packages is built once and installed everywhere the story is a bit different, but since here we need to build on each system, I believe we could just build the `-full` variant and try it on each system.

We could try to depend on some more specific jobs to finish, but I'm not sure the benefit would be worth the cost. Having things run in a separate job is useful to ensure we're not using the local caches (when testing that conversions work)

dangerzone/gui/startup.py
almet: Hmm, this won't actually work, and it will ask the user a second time.

This was done to avoid asking them again. It introduces the in-memory `skip_prompt_once` variable as a way to remember the choice the user just entered in a previous prompt.

Adding the conditions you propose in `can_proceed` won't change this, because at this stage it's not already downloaded, unfortunately.

dangerzone/gui/updater.py
almet: This should only display on Linux (Windows and macOS variants ship with the container already). I've updated the link to our Linux installation instructions in 6d881b25dd6adb5b85b85452727d29738d16fcdb

dangerzone/gui/updater.py
almet: I've actually updated this in 8c6f33cec9631f19f96c2a0834d125769fe1d4b4. The text and message in the button now look like this: 

Image

dangerzone/updater/releases.py
almet: Good call! I extended the check just below to do that for us: 4de09480aaa9cada63f41fb73b22e007ae3a8a2e

debian/control
almet: Ah, you're correct! I'm one step ahead 📦 =D See c1743329a8eab9edd81099e398fd2fd0b7a70cc5

dangerzone/startup.py
almet: I haven't seen a comment about this elsewhere. Let me know how you'd prefer to do things :-)

.github/workflows/ci.yml
apyrgio: That's awesome, thanks!

.github/workflows/ci.yml
almet: Actually implemented this as 430f317f4556111e8417c7d16d7a1573f08cd29c

dangerzone/startup.py
apyrgio: We discussed about this offline, and we decided to show a prompt based on the current state of the system, instead of the result of a previous step. So, we're good here.

dangerzone/gui/startup.py
apyrgio: We have landed on a way to do this, as I've commented [here](https://github.com/freedomofpress/dangerzone/pull/1423#discussion_r3342391991), so we're good here too.

apyrgio (approved):

apyrgio:
Yay, looks awesome! I've approved this PR, but remember to do two things before you merge this:
1. Remove the stale `build-app.py` and resolve [my comment](https://github.com/freedomofpress/dangerzone/pull/1423#discussion_r3221598664).
2. Squash fixups.

almet:
Just rebased, let's hope the CI will still be green!

install/windows/build-app.py
apyrgio: We had discussed this live, and we said that we can remove it, so I went ahead and did so.

PR #1451, opened 2026-04-07
test: fuzzing harnesses for IPC pixel stream protocol

jw409:
## Context

Initial attempt at fuzzing the IPC pixel stream protocol between the sandbox and host. After review, we realized the harnesses targeted the wrong layer — they fuzzed inlined Python parsing code (which cannot buffer overrun or integer overflow) rather than the actual C attack surface in PyMuPDF/MuPDF.

## What was here

- Atheris (libFuzzer) coverage-guided fuzzer for the IPC wire protocol
- Standalone random fuzzer with 7 input generators (boundary values, truncated streams, etc.)
- Both inlined the parsing functions from base.py to avoid importing PyMuPDF

## Why it was closed

Fuzzing int.from_bytes() and simple integer comparisons in pure Python is testing that Python works, not finding security bugs. The real attack surface is fitz.Pixmap() and fitz.Document operations — C code in MuPDF where integer overflow (e.g., CVE-2026-3308), heap buffer mismatch, and use-after-free bugs actually live.

## Superseded by

- Hardening fixes: #1454
- Security test harness targeting C boundary: #1455

### DIFF
--- a/tests/fuzz_ipc_protocol.py
+++ b/tests/fuzz_ipc_protocol.py
@@ -1,0 +1,139 @@
+"""
+Atheris-based coverage-guided fuzzer for the Dangerzone IPC pixel stream protocol.
+
+The IPC protocol between the container and the host transmits rasterized pages
+over stdout as a binary stream:
+
+    page_count: 2 bytes, big-endian unsigned int
+    Per page:
+        width:  2 bytes, big-endian unsigned int
+        height: 2 bytes, big-endian unsigned int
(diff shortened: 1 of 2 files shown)


jw408:
Superseded — retargeting to test the actual C attack surface (PyMuPDF/MuPDF) rather than inlined Python parsing. New PR incoming.

jw409:
Superseded — retargeting to test the actual C attack surface (PyMuPDF/MuPDF) rather than inlined Python parsing. New PR incoming.

PR #1453, opened 2026-04-07
test: Hypothesis property tests for IPC pixel stream protocol

jw409:
## Context

Initial attempt at property-based testing of the IPC pixel stream protocol using Hypothesis. After review, we realized the tests verified Python stdlib correctness rather than exercising the actual security-relevant C code.

## What was here

- 30 Hypothesis property-based tests across 8 categories (read_int round-trip, bounds checking, truncated streams, edge cases, pixel buffer size)
- Inlined parsing functions from base.py to avoid pulling in PyMuPDF dependency chain

## Why it was closed

The properties being tested — "read_int decodes what write_int encodes," "width > MAX_PAGE_WIDTH is rejected" — are tautologies in Python. int.from_bytes() cannot overflow, BytesIO.read() cannot buffer overrun, and the bounds checks are single-line comparisons.

The real attack surface is the C layer: fitz.Pixmap(colorspace, width, height, pixel_data) where MuPDF allocator computes w*depth*n in 32-bit math (see CVE-2026-3308). 

## Superseded by

- Hardening fixes: #1454
- Security test harness targeting C boundary: #1455

### DIFF
--- a/tests/test_ipc_protocol.py
+++ b/tests/test_ipc_protocol.py
@@ -1,0 +1,437 @@
+"""Hypothesis property-based tests for the IPC pixel stream protocol.
+
+The dangerzone IPC protocol sends pixel data from a sandboxed converter
+process to the host over stdout as a binary stream:
+
+    [page_count: 2 bytes big-endian unsigned]
+    for each page:
+        [width:  2 bytes big-endian unsigned]
+        [height: 2 bytes big-endian unsigned]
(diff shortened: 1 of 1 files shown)


jw409:
Superseded — retargeting to test the actual C attack surface (PyMuPDF/MuPDF) rather than inlined Python parsing. New PR incoming.

PR #1454, opened 2026-04-07
security: convert env var forwarding from denylist to allowlist

jw409:
## Summary

The gVisor entrypoint forwards environment variables from the outer container into the sandbox. Previously it used a denylist of 8 specific variables — everything else passed through. In CI or automated deployments, sensitive variables (`AWS_ACCESS_KEY_ID`, `GITHUB_TOKEN`, etc.) would propagate into the sandbox.

This converts to an allowlist: only the locale and timezone variables that LibreOffice needs for text rendering are forwarded (`LANG`, `LC_ALL`, `LC_CTYPE`, `LANGUAGE`, `TZ`).

## Changes

- `entrypoint.py`: Replace denylist (`not_forwarded_env`) with allowlist (`allowed_env`)
- `test_entrypoint_env.py`: New tests verifying allowlist behavior, with AST-based extraction to stay in sync with the source automatically

## Test plan

- [ ] Existing CI passes (no conversion behavior change — locale vars were already forwarded)
- [ ] `test_entrypoint_env.py` verifies allowed vars pass, sensitive vars are blocked, and the test allowlist matches the source via AST parsing

### DIFF
--- a/tests/test_entrypoint_env.py
+++ b/tests/test_entrypoint_env.py
@@ -1,0 +1,73 @@
+"""Tests for the environment variable allowlist in entrypoint.py.
+
+The entrypoint builds an OCI config for gVisor and forwards only allowlisted
+environment variables into the sandbox.  Since entrypoint.py has module-level
+side effects and can't be imported, we parse its AST to extract the allowlist
+and test the filtering logic here.
+"""
+
+import ast
+import pathlib
+
+ENTRYPOINT = pathlib.Path(__file__).parent.parent / (
+    "dangerzone/container_helpers/entrypoint.py"
+)
+
+
+def _read_allowlist() -> set[str]:
+    """Extract the allowed_env set from entrypoint.py via AST."""
+    tree = ast.parse(ENTRYPOINT.read_text())
+    for node in ast.walk(tree):
+        if (
+            isinstance(node, ast.Assign)
+            and len(node.targets) == 1
+            and isinstance(node.targets[0], ast.Name)
+            and node.targets[0].id == "allowed_env"
+            and isinstance(node.value, ast.Set)
+        ):
+            return {
+                elt.value
+                for elt in node.value.elts
+                if isinstance(elt, ast.Constant) and isinstance(elt.value, str)
+            }
+    raise AssertionError("Could not find 'allowed_env = {...}' in entrypoint.py")
+
+
+ALLOWED_ENV = _read_allowlist()
+
+
+def _apply(env: dict[str, str]) -> set[str]:
+    """Reproduce the allowlist logic — return forwarded key names."""
+    return {k for k in env if k in ALLOWED_ENV}
+
+
+class TestEnvAllowlist:
+    def test_forwards_locale_and_tz(self) -> None:
+        env = {"LANG": "en_US.UTF-8", "LC_ALL": "C", "TZ": "UTC"}
+        assert _apply(env) == {"LANG", "LC_ALL", "TZ"}
+
+    def test_blocks_sensitive_vars(self) -> None:
+        env = {
+            "AWS_SECRET_ACCESS_KEY": "x",
+            "GITHUB_TOKEN": "x",
+            "ANTHROPIC_API_KEY": "x",
+        }
+        assert _apply(env) == set()
+
+    def test_blocks_unknown_vars(self) -> None:
+        env = {"MY_CUSTOM_VAR": "x", "SOME_SECRET": "x"}
+        assert _apply(env) == set()
+
+    def test_mixed_env_only_passes_allowed(self) -> None:
+        env = {
+            "LANG": "en_US.UTF-8",
+            "HOME": "/home/user",
+            "PATH": "/usr/bin",
+            "AWS_SECRET_ACCESS_KEY": "x",
+            "TZ": "America/Los_Angeles",
+        }
+        assert _apply(env) == {"LANG", "TZ"}
+
+    def test_allowlist_contains_expected_vars(self) -> None:
+        """Sanity check: the allowlist has exactly the locale + TZ vars."""
+        assert ALLOWED_ENV == {"LANG", "LC_ALL", "LC_CTYPE", "LANGUAGE", "TZ"}
--- a/tests/test_entrypoint_env.py
+++ b/tests/test_entrypoint_env.py
@@ -1,44 +1,11 @@
-"""Tests for the environment variable allowlist in entrypoint.py.
+"""Tests for the environment variable allowlist in entrypoint.py."""
 
-The entrypoint builds an OCI config for gVisor and forwards only allowlisted
-environment variables into the sandbox.  Since entrypoint.py has module-level
-side effects and can't be imported, we parse its AST to extract the allowlist
-and test the filtering logic here.
-"""
-
-import ast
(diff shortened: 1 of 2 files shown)


jw409:
[old comment about wsl2 stuff that has been removed from this PR edited]

jw409:
[old comment about wsl2 stuff that has been removed from this PR edited]

apyrgio:
Converting the denylist to an allowlist is generally useful, so thanks for this PR. There are two things though that I'm not comfortable with merging:

* The WSL2 detection logic: We know that the WSL2 kernel is capable of running `ptrace(2)`. After all, Dangerzone runs in WSL2. Although I haven't tried, Dangerzone should also be able to work within a WSL2 environment as a regular Linux application, provided that you run it with the accompanying seccomp policy, and not using `podman` commands directly. So, it would be preferable if:
   * The WSL2 logic was moved into a different PR (or issue), since it's not related with envvars after all.
   * We investigate whether Podman machine does something wildly different than WSL2 with regards to `ptrace(2)`
* There's _looots_ of testing going on in this PR. This should be a relatively simple change, and the tests for it should also not sprawl over the codebase. Note that by design, envvars from the host are not supposed to leak into the container. With that in mind, I believe we can greatly simplify the tests, and make them more maintainable.

Thanks again!

jw409:
> Converting the denylist to an allowlist is generally useful, so thanks for this PR. There are two things though that I'm not comfortable with merging:
> 
> * The WSL2 detection logic: We know that the WSL2 kernel is capable of running `ptrace(2)`. After all, Dangerzone runs in WSL2. Although I haven't tried, Dangerzone should also be able to work within a WSL2 environment as a regular Linux application, provided that you run it with the accompanying seccomp policy, and not using `podman` commands directly. So, it would be preferable if:
>   
>   * The WSL2 logic was moved into a different PR (or issue), since it's not related with envvars after all.
>   * We investigate whether Podman machine does something wildly different than WSL2 with regards to `ptrace(2)`

Ack, I have removed that. I need to research how I footgun'd myself here, and if I find something that is not just a "jw409 problem" I'll make a seperate PR. 

> * There's _looots_ of testing going on in this PR. This should be a relatively simple change, and the tests for it should also not sprawl over the codebase. Note that by design, envvars from the host are not supposed to leak into the container. With that in mind, I believe we can greatly simplify the tests, and make them more maintainable.
> 
> Thanks again!

Ack, it should be much simplified. Thank you for the kind redirect and simplification, I think this PR is ready for review now.

tests/test_entrypoint_env.py
```diff
@@ -0,0 +1,73 @@
...
+
+import ast
+import pathlib
+
+ENTRYPOINT = pathlib.Path(__file__).parent.parent / (
+    "dangerzone/container_helpers/entrypoint.py"
+)
+
+
+def _read_allowlist() -> set[str]:
```
almet: Rather than using the ast here, copy/pasting seems cleaner

tests/test_entrypoint_env.py
```diff
@@ -0,0 +1,73 @@
...
+            }
+    raise AssertionError("Could not find 'allowed_env = {...}' in entrypoint.py")
+
+
+ALLOWED_ENV = _read_allowlist()
+
+
+def _apply(env: dict[str, str]) -> set[str]:
+    """Reproduce the allowlist logic — return forwarded key names."""
+    return {k for k in env if k in ALLOWED_ENV}
```
almet: It seems that what we're testing here is t he dict comprehension doing an "if". Instead, we could check the behavior of the entrypoint matches our expectations.

tests/test_entrypoint_env.py
jw408: [nevermind; I think I figured out what you meant]

jw409: Good call. I've wrapped the entrypoint's side effects in `if __name__ == "__main__":` and moved `allowed_env` to module level so the test can import it directly — no AST parsing, no duplication. The set is defined once in `entrypoint.py` and the test reads it from there.

tests/test_entrypoint_env.py
jw409: I'm not 100% sure I'm reading this correctly — let me know if I'm off base. My understanding is you'd prefer the test to exercise the actual module rather than reimplementing the filter logic separately. I've refactored `entrypoint.py` with a `__main__` guard so the allowlist can be imported directly, which eliminates the parallel implementation. The tests still construct synthetic envs and check filtering against the real `allowed_env` set, since fully running the entrypoint would need gVisor/container infrastructure. Happy to take a different approach if you had something else in mind.

jw409:
The latest commit wraps the entrypoint's side effects in `if __name__ == "__main__":` so the test can import `allowed_env` directly — no more AST parsing or duplication. The entrypoint diff looks large because every line under the guard gained 4 spaces of indent; append `?w=1` to the URL (or toggle "Hide whitespace" in the diff view) to see the actual semantic changes are just a few lines.

apyrgio:
Thanks for these changes @jw409. I have a few more comments if you don't mind, to make this PR more helpful to us:

1. I looked again at the existing implementation in `main`, and I saw that for the envvars that are not supposed to be forwarded, we mask them with `<ENV>=`, instead of dropping them. The proposed implementation in this PR is different though, and will simply ignore these variables. I'm not sure if this makes any difference to gVisor, but probably it's best to retain the original behavior.
2. Turning the entrypoint.py in a module with `if __name__ == "__main__":` is a cool idea, and allows us to test the `allowed_env` list. However, it doesn't test how this list applies to the final OCI config. Since we started going down that path, why not factor out also the generation of the OCI config, and check that the envvars are actually masked in it?
3. We want the `test/` structure to reflect the structure of our sources, so I would move the tests under `tests/container_helpers/test_entrypoint.py`, to match the rest of our tests.
4. Due to the upcoming split of Dangerzone into two repos, where the second repo will hold only the image/converter logic (see https://github.com/freedomofpress/dangerzone/pull/1459/changes), I think this PR should target the newly-created https://github.com/freedomofpress/dangerzone-image repo.

jw409:
Thank you @apyrgio for the detailed feedback — all four points landed and reshaped the change substantially. Reposting in the right place:

**→ https://github.com/freedomofpress/dangerzone-image/pull/2**

Summary of what changed per your comments:

1. **Mask vs drop** — went with **drop**, not mask. Allowlist + drop removes the key *name* as a signal (`AWS_SECRET_ACCESS_KEY`, `SSH_AUTH_SOCK`, `KUBECONFIG` and friends never appear in the OCI config at all). gVisor behavior is identical either way; drop is the stricter position and that's the one the sandbox boundary should take.
2. **Test the final OCI config, not just the allowlist set** — done. Factored out `build_oci_config(command, env)` as a pure function; the test drives it with hostile env dicts and asserts on the actual `config["process"]["env"]` array reaching gVisor. 20 tests covering baseline presence, allowlist forward, sensitive-name-and-value absence, sandbox-control vars, baseline shadowing, and hostile values (newlines/nulls/shell injection).
3. **Mirror source layout** — tests now at `tests/helpers/test_entrypoint.py` (matching `src/helpers/entrypoint.py`'s location in dangerzone-image).
4. **Target dangerzone-image** — done; moving this PR to draft.

PR #1456, opened 2026-04-08
Feat/deepseek ocr

jw409:
Deepseek-3B-OCR


jw409:
created in error on wrong repo.

PR #1457, opened 2026-04-09
gui: elide long filenames with tooltip in document list

### LINKED ISSUE #143: Word-wrap the document filename label
If the filename is very long, you can't read the whole thing.

paste

moaaz-ae:
When a file with a very long name is selected, the application window would resize awkwardly or push elements out of the viewport.

For example (_on Fedora Linux_), without this fix, the **"Start another conversion"** button gets pushed completely off-screen when selecting a file with a long filename, making it impossible for the user to trigger another conversion. See the "Before" image below.

This PR improves the UI by:

1. Using `QFontMetrics.elidedText()` with `ElideMiddle` to shorten the displayed filename if it exceeds the available width.
2. Adding a tooltip to the filename label so that the full path is still visible on hover.

Fixes #143

**Before**

image

**After**

image

### DIFF
--- a/dangerzone/gui/main_window.py
+++ b/dangerzone/gui/main_window.py
@@ -1667,9 +1667,14 @@ class DocumentWidget(QtWidgets.QWidget):
         self.dangerous_doc_label.setAlignment(
             QtCore.Qt.AlignVCenter | QtCore.Qt.AlignLeft
         )
-        self.dangerous_doc_label.setText(os.path.basename(self.document.input_filename))
+        doc_name = os.path.basename(self.document.input_filename)
+        self.dangerous_doc_label.setToolTip(doc_name)
         self.dangerous_doc_label.setMinimumWidth(200)
         self.dangerous_doc_label.setMaximumWidth(200)
+        # Elide text with '...' when the filename is too long for the label width
+        metrics = self.dangerous_doc_label.fontMetrics()
+        elided = metrics.elidedText(doc_name, QtCore.Qt.ElideMiddle, 200)
+        self.dangerous_doc_label.setText(elided)
 
         # Conversion status images
         self.img_status_unconverted = self.load_status_image("status_unconverted.png")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 - Update the theme correctly on dark/light mode changes
   ([#1238](https://github.com/freedomofpress/dangerzone/issues/1238))
 - Update Linux .desktop file with missing MIME types for HWP and ODF formats ([#646](https://github.com/freedomofpress/dangerzone/issues/646))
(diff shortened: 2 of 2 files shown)


dangerzone/gui/main_window.py
```diff
@@ -1660,9 +1660,14 @@ def __init__(
...
             QtCore.Qt.AlignVCenter | QtCore.Qt.AlignLeft
         )
-        self.dangerous_doc_label.setText(os.path.basename(self.document.input_filename))
+        doc_name = os.path.basename(self.document.input_filename)
+        self.dangerous_doc_label.setToolTip(doc_name)
         self.dangerous_doc_label.setMinimumWidth(200)
         self.dangerous_doc_label.setMaximumWidth(200)
+        # Elide text with '...' when the filename is too long for the label width
+        metrics = self.dangerous_doc_label.fontMetrics()
+        elided = metrics.elidedText(doc_name, QtCore.Qt.ElideMiddle, 195)
```
almet: Given that we set `self.dangerous_doc_label.setMinimumWidth(200)`, I wonder why this is `195`? Is it to let the place for the `...`?

almet:
Thanks a lot for the changes, that looks great. I've added a small question, and then we can merge!

dangerzone/gui/main_window.py
moaaz-ae: Yep, https://doc.qt.io/qt-6/qfontmetrics.html#elidedText isn’t really clear about that. I thought I needed to make space for `...`, but I just tried using 200 and it looks the same (I can’t notice any difference). It seems I don’t need to reduce it to 195

apyrgio:
The addition of ellipsis is a welcome one, and I'll merge it once the tests pass. One thing I didn't manage to reproduce though is the push of the "Start another conversion" button to the right.

Truth is, I have tested it only on a Fedora container, not a VM, but since it works in your machine and is in the correct direction, I think it's safe to merge.

moaaz-ae:
If this helps with reproduction:

OS: Fedora 43
Desktop Environment: GNOME (wayland)
Scaling Factor: 1.0
System Font: Adwaita Sans 11
Install Method: RPM (dangerzone-0.10.0-1.fc43.x86_64)
Window State: Default size
Filename: This_is_an_extremely_long_filename_created_specifically_to_hit_the_absolute_maximum_limit_of_two_hundred_and_fifty_five_characters_on_a_standard_linux_filesystem_like_ext4_if_you_add_one_more_character_to_this_string_the_system_will_throw.pdf

apyrgio:
Ok, I tried it out in a Fedora 43 VM as well, using Gnome, but I still can't reproduce it.

But, wait a sec:

>  Install Method: RPM (dangerzone-0.10.0-1.fc43.x86_64)

The "Start another conversion" button is not present in the 0.10.0 release yet. So... could it be that this is what you're observing?

In any case, the [tests have passed](https://github.com/freedomofpress/dangerzone/actions/runs/24444304208/job/71431098246), so I'll merge this PR.

apyrgio (approved):

moaaz-ae:
The version was actually the problem. Feel a bit silly for missing that. Thanks for merging.

PR #1458, opened 2026-04-09
fix: Add missing MIME types to .desktop file

### LINKED ISSUE #646: Update Linux .desktop file with new supported formats (HWP)
When we added [HWP file format support](https://github.com/freedomofpress/dangerzone/pull/460) we forgot to update the following line:

https://github.com/freedomofpress/dangerzone/blob/7f50ad2e4836b243326a1e2483e7c3fb5c2f3b78/install/linux/press.freedom.dangerzone.desktop#L10

This affects only Linux platforms and the consequence is that users won't be able to easily add Dangerzone as default application for the new file formats until we fix this.

moaaz-ae:
The Linux desktop entry was missing several supported MIME types. This PR adds the following to `press.freedom.dangerzone.desktop`:

- `application/x-hwp`
- `application/x-hwp-template`
- `application/vnd.oasis.opendocument.text-template`
- `application/vnd.oasis.opendocument.spreadsheet-template`
- `application/vnd.oasis.opendocument.presentation-template`

These formats were already handled by the conversion logic but wouldn't appear in "Open With" menus for these file types.

Fixes #646


almet (approved):
Great, thanks!

apyrgio (approved):

PR #1459, opened 2026-04-10
Split dangerzone image to a separate repository

almet:
> [!NOTE]
> These changes come together with the https://github.com/freedomofpress/dangerzone-image repository, so be sure to review the code there as well.

This branch removes the in-tree container image (build, conversion code, Dockerfiles) and rewires the client-side code to consume the published image from the new freedomofpress/dangerzone-image repository. The bulk of the diff is deletions plus the test/CI changes, which don't test every supported format locally (it's done on the dangerzone-image repo now)

Look at the individual commits for more details.

Worth noting: this makes the tests stop popping on our dev machines, by setting the `QT_QPA_PLATFORM` env var to `offscreen` (there is still the possibility to pass the `--onscreen` flag to pytest if required)


THIRD_PARTY_NOTICE.md
```diff
@@ -15,17 +18,6 @@ Code signing and transparency for containers and binaries.
...
-This is the public key used for verifying packages from the gVisor repository.
-
-- License: N/A
-- Repository: https://github.com/google/gvisor
-- Download link: https://gvisor.dev/archive.key
-- Included in: Dangerzone container image and Dangerzone Git repo
-
-_(Last updated: 2025-01-21)_
-
 ## gvproxy
```
almet, line 21: I should remove this as well.

pyproject.toml
```diff
@@ -275,9 +275,9 @@ destination = "share/vendor/podman/gvproxy"
 # sign their releases (see https://github.com/crc-org/vfkit/issues/437).
 [tool.mazette.asset.vfkit]
 repo = "crc-org/vfkit"
-version = "==0.6.1"
-platform."darwin/amd64" = "vfkit"
-platform."darwin/arm64" = "vfkit"
+version = ">=0.6.1"
+platform."darwin/amd64" = "vfkit-unsigned"
+platform."darwin/arm64" = "vfkit-unsigned"
```
almet, line 280: Had to do this because vfkit stopped providing signed binaries. They are now -unsigned, and I think that's what we might want, even though I don't really like it.

.github/workflows/ci.yml
```diff
@@ -120,16 +85,33 @@ jobs:
...
+          key: v1-container-tar-${{ runner.arch }}-${{ steps.image-digest.outputs.digest }}
+
+      - name: Download the container image as a tar archive
+        if: steps.cache-container.outputs.cache-hit != 'true'
+        env:
+          DANGERZONE_DEV: "1"
+        run: >-
+          poetry run dangerzone-image prepare-archive
+          --output share/container.tar
+
```
apyrgio, line 114: Quick question here: since this invocation is repeated in several places in the code, should we move it in a single place, as we did for the `build-container-image` job, and make it a dependency for the rest of the steps?

Other than that, wow, 4 seconds to restore the image from cache, that's amazing!

.gitignore
```diff
@@ -149,7 +149,7 @@ tests/test_docs_large/
 install/windows/Dangerzone.wxs
 share/container.tar
 share/container.tar.gz
-share/image-id.txt
+share/image-name.txt
```
apyrgio, line 152: I agree we should ignore `share/image-id.txt`, but isn't `share/image-name.txt` necessary?

CHANGELOG.md
```diff
@@ -33,6 +33,34 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
   ([#1338](https://github.com/freedomofpress/dangerzone/issues/1338))
 - Pin Podman to v5.x.x, to support Windows 10 installations for one last release
   ([#1357](https://github.com/freedomofpress/dangerzone/issues/1357))
+- Move the conversion error definitions out of the now-removed `conversion/`
+  package and into `dangerzone.conversion_errors`, so the host code keeps a
+  single source of truth for the error codes raised inside the sandbox.
+- Split the container image build out into a separate
```
apyrgio, line 39:
```suggestion
- Move the code for the container image into a separate
```

pyproject.toml
apyrgio: Just an fyi, I had opened an issue about it, and pinged the maintainer: https://github.com/crc-org/vfkit/issues/437. Other than that, I guess we don't have any other option.

dangerzone/isolation_provider/qubes.py
```diff
@@ -85,22 +84,19 @@ def terminate_doc_to_pixels_proc(
...
         with zipfile.ZipFile(temp_file, "w") as z:
-            z.mkdir("dangerzone/")
-            z.writestr("dangerzone/__init__.py", "")
             for root, _, files in os.walk(_conv_path):
                 for file in files:
                     if file.endswith(".py"):
                         file_path = os.path.join(root, file)
-                        relative_path = os.path.relpath(file_path, _src_root)
+                        relative_path = os.path.relpath(file_path, _conv_path.parent)
                         z.write(file_path, relative_path)
```
apyrgio, line 100: Note to self to test this part in Qubes.

tests/conftest.py
```diff
@@ -155,12 +150,9 @@ def noop(*args: Any, **kwargs: Any) -> bool:
 
 SAMPLE_DIRECTORY = "test_docs"
 BASIC_SAMPLE_PDF = "sample-pdf.pdf"
-BASIC_SAMPLE_DOC = "sample-doc.doc"
-SAMPLE_EXTERNAL_DIRECTORY = "test_docs_external"
-SAMPLE_COMPRESSED_DIRECTORY = "test_docs_compressed"
+BASIC_SAMPLE_PDF2 = "sample-pdf2.pdf"
```
apyrgio, line 153: Minor, feel free to ignore it:

Why are we using a copy of the original PDF, with a different name? If we opted to keep the `sample-doc.doc` without including it in our conversion tests, we would have the following benefits:
1. We wouldn't need these extra changes in the code.
2. We would have an office document handy for our manual tests.

.github/workflows/ci.yml
```diff
@@ -489,53 +472,45 @@ jobs:
...
-            share/image-name.txt
-          enableCrossOsArchive: true
-          fail-on-cache-miss: true
-          key: v6-container-${{ needs.build-container-image.outputs.image_uri }}
-
-      - name: Setup xvfb (Linux)
-        run: |
-          sudo apt update
-          # Stuff copied wildly from several stackoverflow posts
-          sudo apt-get install -y xvfb libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xinput0 libxcb-xfixes0 libxcb-shape0 libglib2.0-0 libgl1-mesa-dev '^libxcb.*-dev' libx11-xcb-dev libglu1-mesa-dev libxrender-dev libxi-dev libxkbcommon-dev libxkbcommon-x11-dev
```
apyrgio, line 507: 🥳

apyrgio (commented):
Amazing work Alexis! My review comments are pretty minor, I think we're very close to merging this.

.github/workflows/ci.yml
```diff
@@ -581,20 +553,22 @@ jobs:
...
-          key: v6-container-${{ needs.build-container-image.outputs.image_uri }}
+      - name: Use the testing image and key
+        run: |-
+          cp tests/assets/dangerzone-testing.pub share/freedomofpress-dangerzone.pub
+          echo "ghcr.io/freedomofpress/dangerzone-testing/main/v1" > share/image-name.txt
+
+      - name: Download the old container image as an archive
+        run: >-
+          poetry run dangerzone-image prepare-archive
+          --image $(cat share/image-name.txt):old-latest
```
apyrgio, line 565: Question: Instead of relying on `old-latest`, and adding code in the `dangerzone-image` repo in order to tag the previous `latest` tag, can't we just use here a known image digest, which is guaranteed to be older than latest? My understanding is that we don't lose something, right?

pyproject.toml
cfergeau: I’ve finally added a signed binary, and will make sure to provide signed binaries in future releases. Thanks for the issue, the reminder and your patience !

tests/conftest.py
almet: I think my rationale was the following:

- Dangerzone only deals with PDFs now
- Let's use PDFs. Because we already have one handy, let's just copy it.

I would be okay with keeping a different format, ultimately. I see it mainly as a matter of taste.

But because you mention it's minor feedback, I'll let it as it is: it's already implemented and doesn't change dramatically the testing, so let's keep things simple :).

.github/workflows/ci.yml
almet: Ah, interesting!

Not having to tag `old-latest` would be great, I agree. It's been done to avoid the "clients" to deal with the complexity themselves.

I don't immediately see a way to do this easily, but you might have something in mind that I don't see at first pass? 

PS: This other older image digest need to be signed with the testing keys, and ideally keep the same image name.

.gitignore
almet: `image-id.txt` should be gone with these new changes, hence the removal.

You're correct, that we should keep `image-name.txt` though!

.github/workflows/ci.yml
almet: We discussed this live, and while it could make sense on long run, we don't plan to change this right now, as it's just a one liner.

.github/workflows/ci.yml
almet: We agreed together that this is easy enough, and couldn't think about better ways atm.

pyproject.toml
almet: Thanks; I now use the signed binary!

CHANGELOG.md
almet: Thanks, applied!

almet:
I've fixed the CI issue doing the E2E update, and I did this by changing how the old-latest images are tagged on the dangerzone-image repository.

It now tags `old-latest` only if it's different from the `latest`, and only if there are changes. For feature branches, we copy the `old-latest` from the `main` git branch.

Last item pending is the test on Qubes, and we should be good to merge!

dangerzone/isolation_provider/qubes.py
apyrgio: This has been tested successfully!

apyrgio (approved):

apyrgio:
I have:
* signed the commits once again,
* rebased on top of `main`,
* force-pushed the branch and
* approved the changes.

Let's merge once the CI tests pass, with the exception of the security scans. We have discussed internally and we will push a new image soon.

PR #1463, opened 2026-04-23
Make Qubes conversion work after the dangerzone-image split

apyrgio:
Update the Dangerzone Qubes logic and build instructions to work with the new server-side component (see also https://github.com/freedomofpress/dangerzone-image/pull/7).


dangerzone/isolation_provider/qubes.py
```diff
@@ -25,7 +25,8 @@ def get_max_parallel_conversions(self) -> int:
         return 1
 
     def start_doc_to_pixels_proc(self, document: Document) -> subprocess.Popen:
-        dev_mode = getattr(sys, "dangerzone_dev", False) is True
+        conv_mod_path = os.environ.get("DANGERZONE_INSECURE_CONVERTER_PATH", None)
```
almet, line 28: Should we move this variable down in the section where it's used, and avoid passing it along?

BUILD.md
```diff
@@ -297,15 +305,22 @@ And when creating a `.rpm` you'll need to enable the `--qubes` flag.
...
-For changes in the server side components, you can simply edit them locally,
-and they will be mirrored to the disposable qube through the `dz.ConvertDev`
-RPC call.
-
-The only reason to build a new Qubes RPM and install it in the `fedora-42-dz`
-template for development is if:
-
-1. The project requires new server-side components.
-2. The code for `qubes/dz.ConvertDev` needs to be updated.
+> [!TIP]
```
almet, line 314: This makes me see that when we actually change how we write documentation, we could have sections about where to handle things for developers specifically, and that will be different from what it is to build the project yourself.

E.g. I don't expect anybody building the project to have to do this, unless they want to make edits to how we handle the Qubes conversion (still, it's good to have these info somewhere).

It's just an info / comment, nothing to change here.

almet (commented):
I've added a few notes. I think we're almost good to go with this. Mainly a few changes related to how we do the teleport, and that's it!

dangerzone/isolation_provider/qubes.py
apyrgio: It's also used to figure out if we will enable dev mode, right after this line. So, it needs to stay up here.

BUILD.md
apyrgio: Makes sense.

dangerzone/isolation_provider/qubes.py
almet: Ah, I missed the `and conv_mod_path` part further down. Okay!

apyrgio:
I've squashed the fixups in this branch, and tested the Qubes tests locally. I think it's ready to merge once the CI turns green.

almet:
The linter reports this:

```
dangerzone/isolation_provider/qubes.py:94: error: Value of type variable "AnyStr" of "walk" cannot be "str | None"  [type-var]
dangerzone/isolation_provider/qubes.py:96: error: Item "None" of "str | None" has no attribute "endswith"  [union-attr]
dangerzone/isolation_provider/qubes.py:97: error: Argument 1 to "join" has incompatible type "str | None"; expected "str"  [arg-type]
dangerzone/isolation_provider/qubes.py:97: error: Argument 2 to "join" has incompatible type "str | None"; expected "str"  [arg-type]
Found 4 errors in 1 file (checked 69 source files)
```

Otherwise, it's good! Since this targets the dangerzone-split branch, if you don't have the time to fix it before I get another look at it, I'll merge and fix in this other branch :-)

Thanks a lot :+1:

apyrgio:
Fixed it:

```diff
diff --git a/dangerzone/isolation_provider/qubes.py b/dangerzone/isolation_provider/qubes.py
index 2f99e892..4fd8f49d 100644
--- a/dangerzone/isolation_provider/qubes.py
+++ b/dangerzone/isolation_provider/qubes.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 import zipfile
 from pathlib import Path
-from typing import IO, Callable, Optional
+from typing import IO

 from ..document import Document
 from ..updater.signatures import is_container_tar_bundled
@@ -49,6 +49,7 @@ class Qubes(IsolationProvider):

         if dev_mode:
             assert p.stdin is not None
+            assert conv_mod_path is not None
             # Send the dangerzone module first.
             self.teleport_dz_module(conv_mod_path, p.stdin)

@@ -84,7 +85,7 @@ class Qubes(IsolationProvider):
         if p.stdout:
             p.stdout.close()

-    def teleport_dz_module(self, conv_path: Optional[str], wpipe: IO[bytes]) -> None:
+    def teleport_dz_module(self, conv_path: str, wpipe: IO[bytes]) -> None:
         """Send the dangerzone module to another qube, as a zipfile."""
         # Grab the absolute file path of the dangerzone module.
         temp_file = io.BytesIO()
```

Issue #549, opened 2023-09-21
Ability to add more documents after conversion finished


deeplow:
After converting a document a user may remember they want to add another one (or not realize multiple docs can be added in the first place).

PR #1464, opened 2026-04-23
Queue new documents, when added during a conversion

almet:
When a document is added during a conversion, queue it. When a document is added after a conversion is finished, start afresh and display the settings pane.

This is a sequel to #549, and especially helpful on macOS, where — if Dangerzone is the default app to open PDFs — it will show a warning with the current implementation.


apyrgio:
I've managed to test this feature and it does work great. There's one thing though that's not covered by this PR, but we could add it in a subsequent one: if you select multiple files and do "Open with", only the first file is converted.

Other than that, it's good to merge, and certainly better than before.

apyrgio (approved):

PR #1466, opened 2026-04-23
Drop cx_Freeze pin

almet:
Fixes #1443


pyproject.toml
```diff
@@ -42,7 +42,7 @@ dangerzone = "dangerzone.gui.run:run"
 # Dependencies required for packaging the code on various platforms.
 [tool.poetry.group.package.dependencies]
 setuptools = "*"
-cx_freeze = { version = "<8.5.0", platform = "win32" }
+cx_freeze = { version = "*", platform = "win32" }
```
apyrgio, line 45: Given that we know the fixed version, let's make this slight change:

```suggestion
cx_freeze = { version = "^8.5.1", platform = "win32" }
```

almet: I don't think it's best to pin the version here again. Since we'll release with it and it's not a debian dep or something else, I find it slightly easier to deal with a lax version check (e.g. use the latest) unless we have a reason to pin.

apyrgio: Hm, that's not a pin actually, I've added the caret (`^`). It's mostly so that we don't get a `9.x.x` version, which may have breaking changes, and to signal that 8.5.0 cannot be used. Anything in between should work.

almet: As I mentioned in my previous message, I don't think it's best to use a specific version (by pin or by using the caret).

Keeping it lax allows us to detect problematic versions and potentially find a fix earlier, and currently doesn't have downsides from my POV.

Hit me up if you want to discuss that live!

apyrgio: Ok, got it! I don't have any strong opinion, I was just making sure we're on the same page. Let's disregard this then.

apyrgio (approved):

PR #1475, opened 2026-05-04
macOS: Fix packaging

### LINKED ISSUE #1470: Building macOS binaries from `main` is currently broken
We have recently made some changes in the `main` branch, that are not reflected in our packaging scripts. This means that creating an macOS app with `poetry run ./install/macos/build-app.py` will work, but the resulting application will fail with:

```python
[PYI-8545:ERROR] Failed to execute script 'dangerzone' due to unhandled exception: module 'dangerzone' has no attribute 'main'
[PYI-8545:ERROR] Traceback:
  File "dangerzone", line 13, in <module>
AttributeError: module 'dangerzone' has no attribute 'main'
```

Note that this does not affect the stable Dangerzone application that is available from our downloads page.

almet:
The location of the main entrypoints changed, but the macOS packaging scripts weren't updated accordingly. This fixes it.

Fixes #1470


apyrgio (approved):

apyrgio:
Υay, it works!

PR #1476, opened 2026-05-06
Remove scripts for building, reproducing or verifying images

apyrgio:
Remove some scripts and code from this repo that was used exclusively to build, reproduce or verify images. This task is now performed by the `freedomofpress/dangerzone-image` repo.

Also, fix some failing security scans in our repo that required a boilerplate `.grype.yaml`.

Refs freedomofpress/dangerzone-image#21


almet (approved):
Thanks, it looks great!

On top of what you did, I also removed the test path that used to check generated PDFs against their reference versions. This is currently done in the dangerzone-image repository, and the code duplication isn't adding value here.

So, let me know if that works for you and we're good to merge!

apyrgio:
Nice catch, merged!

PR #1481, opened 2026-05-22
test: probe GHCR write on fork PR

qsq8t65pxs:
Minimal security workflow probe. This only adds a non-functional comment to `dev_scripts/env.py` so the build-dev image hash changes without changing behavior.


qsq8t65pxs:
Closing minimal security workflow gating probe.

PR #1483, opened 2026-05-26
Bump versions

almet:
- Unpin python dependencies, unless we have a reason for it. It will fail the CI quicker, but will make it possible for us to find sooner than later.
- Add digests for GHA where it makes sense, as it's a security good practice nowadays.
- Remove the `generate-release-notes.py` script, as we now populate the CHANGELOG entries on the go.
- Remove the pymupdf specific version for Debian, it can now use any version.


almet:
Noting here that tests are passing just right before I remove Debian bullseye support (and python 3.9 with it)

.github/workflows/ci.yml
```diff
@@ -253,9 +251,11 @@ jobs:
   install-deb:
     name: "install-deb (${{ matrix.distro }} ${{ matrix.version }})"
     runs-on: ubuntu-latest
+    continue-on-error: true
```
apyrgio: I do get `fail-fast: false`, but I'm not sure about this. Won't this mean that the job will succeed even though a step fails?

dangerzone/util.py
```diff
@@ -82,7 +82,6 @@ def get_tessdata_dir() -> Path:
         Path("/usr/share/tessdata/"),  # on some Debian
         Path("/usr/share/tesseract/tessdata/"),  # on Fedora
         Path("/usr/share/tesseract-ocr/tessdata/"),  # ? (documented)
-        Path("/usr/share/tesseract-ocr/4.00/tessdata/"),  # on Debian Bullseye
```
apyrgio, line 85: Most likely, this is the reason why the Ubuntu 22.04 tests fail. Let's keep this.

apyrgio (commented):
That's amazing work Alexis. I have a few comments here, but they are mostly minor. Other than that, I think it's good to merge.

What I'd like to ask though before merging this PR is a couple of questions about dependency management in general, now that we have this in our cache:

1. How did you generate the pins for the GitHub actions? I tested your pins with the following command, but I'm wondering if you have a better way:

   ```
   podman run --rm -v $(pwd):/workflows docker.io/mheap/pin-github-action  --recursive .github/workflows/
   ```

2. Now that we have pinned GitHub actions everywhere, should we ask the infra team to enable [this option](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/) for our repo?

3. For dev/test dependencies, is it worth it to be on the bleeding edge, instead of sticking with a major version? I'm saying this because I see you had to update the tests to make them work after bumping some `pytest-*` dependencies. I don't have a strong opinion by the way security-wise, I'm mostly wondering about convenience here.

4. For the dependencies that we choose to always follow the latest version, should we perhaps wait a number of days before Poetry picks them up? (see [solver.min-release-age](https://python-poetry.org/docs/configuration/#solvermin-release-age))

Note that we can merge this PR without answering the above now, and convert the above questions to an issue.

.github/workflows/ci.yml
almet: I think I should remove it, it doesn't seem to work :/ that was an attempt to avoid stopping the tests for other variants when one is failing.

almet:
Nice! Thanks for the feedback!

Feel free to update this and merge it to `main`, as I won't be able to tomorrow, otherwise it will wait until I'm back.

>     1. How did you generate the pins for the GitHub actions? I tested your pins with the following command, but I'm wondering if you have a better way:
>        ```
>        podman run --rm -v $(pwd):/workflows docker.io/mheap/pin-github-action  --recursive .github/workflows/
>        ```

That's actually quite nice! I… did things manually and it was a pain, that's cool to see such a tool exists!

>     2. Now that we have pinned GitHub actions everywhere, should we ask the infra team to enable [this option](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/) for our repo?

It sounds like a good idea!

>     3. For dev/test dependencies, is it worth it to be on the bleeding edge, instead of sticking with a major version? I'm saying this because I see you had to update the tests to make them work after bumping some `pytest-*` dependencies. I don't have a strong opinion by the way security-wise, I'm mostly wondering about convenience here.

Yeah, to me that's the whole point. The known working set is encoded in the `poetry.lock` file and so being lax on the `pyproject.toml` seem the best thing to do… unless we find it too cumbersome in the future, but… let's try!

>     4. For the dependencies that we choose to always follow the latest version, should we perhaps wait a number of days before Poetry picks them up? (see [solver.min-release-age](https://python-poetry.org/docs/configuration/#solvermin-release-age))

100% agreed!

> Note that we can merge this PR without answering the above now, and convert the above questions to an issue.

I will be out most of tomorrow, so it's your call: either you have time to do this tomorrow and I can review when going to the city and back, or we let it like that and do it later :-) Either work for me!

apyrgio:
> I will be out most of tomorrow, so it's your call: either you have time to do this tomorrow and I can review when going to the city and back, or we let it like that and do it later :-) Either work for me!

I can give it a shot tomorrow. More specifically, I will:
* Rebase the PR and fix some conflicts with main.
* Address some of the stuff I mentioned in the comments and squash them into your commits.
* Experiment with the `solver.min-release-age` flag, and see if it changes something in our Poetry lock file.
* Merge, if there are no important differences
* (post-merge) Open an issue to infra to update this setting for pinned GHA for us.

dangerzone/util.py
apyrgio: Fixed in 95b68a2624ff641f68ed34bbe9c06a85ae43ffef

.github/workflows/ci.yml
apyrgio: Fixed in 74cb33ef17dbe90605ef7dc4bad3642b3d408e4f.

apyrgio (approved):

apyrgio:
Ok, I have:
* force-pushed the rebased branch,
* fixed a couple of issues I had commented above,
* introduced a dependency cooldown (see 825532da0558cd09faa931fcbdf46f31fb295bf5), and saw that it works, with minimal changes to the installed packages
* approved the PR

Once you I have the green light, I can squash the fixups, merge, and open an issue to infra.

almet:
Okay, I've just force-pushed and squashed your fixups. Let's merge when the CI is green!

PR #1484, opened 2026-05-27
Make it possible to output images

### LINKED ISSUE #629: Save Images as Images (instead of PDFs)
Dangerzone, can accept image formats. However it outputs them as a PDF. But it would be pretty simple to save them as images (even if it's in another format). That would much more useful. 

## User Story
(theoretical) A journalist before publishing images in a sensitive story, runs them through Dangerzone as a general practice. The journalist used a phone to crop the the face of a source from picture that was later found vulnerable to something like the [aCropalypse](https://en.wikipedia.org/wiki/ACropalypse). Dangerzone would have prevented a reversal of that crop.

However, because Dangerzone currently saves the images as PDF that wouldn't have been possible.

almet:
Fixes #629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 - Accept OCI image indexes as multi-arch container images, in addition to Docker manifest lists. BuildKit 0.31.0 and later pushes OCI image indexes by default, which made `dangerzone-image prepare-archive` fail with `InvalidMutliArchImage` against newly published images ([#1534](https://github.com/freedomofpress/dangerzone/pulls/1534)).
 
+### Added
+
+- Allow converting documents to images: a new "Output format" option
+  (`--output-format` in the CLI) can convert everything to PDFs, everything to
+  PNG images, or detect the format automatically (the default): documents are
+  converted to PDFs, and images to images
+  ([#629](https://github.com/freedomofpress/dangerzone/issues/629))
+
 ### Development changes
 
 - Add a security policy for our project
--- a/dangerzone/args.py
+++ b/dangerzone/args.py
@@ -37,9 +37,7 @@ def _validate_output_filename(
 ) -> str | None:
     if value is None:
         return None
-    filename = Document.normalize_filename(value)
-    Document.validate_output_filename(filename)
-    return filename
+    return Document.normalize_filename(value)
 
 
 # XXX: Click versions 7.x and below inspect the number of arguments that the
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -25,6 +25,12 @@ def print_header(s: str) -> None:
     callback=args.validate_output_filename,
     help=f"Default is filename ending with {SAFE_EXTENSION}",
 )
+@click.option(
+    "--output-format",
+    type=click.Choice(["pdf", "png"], case_sensitive=False),
+    default="pdf",
+    help="Output format for safe documents (pdf or png)",
+)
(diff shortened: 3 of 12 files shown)


apyrgio:
A couple of quick issues that I've encountered:
1. If I select "Save safe PDFs to ..." and also select an image as an output, I get the following error:

    ```python
      File "/home/user/dangerzone/dangerzone/gui/main_window.py", line 1590, in start_button_clicked
        document.suffix = self.safe_extension.text()
        ^^^^^^^^^^^^^^^
      File "/home/user/dangerzone/dangerzone/document.py", line 146, in suffix
        raise errors.SuffixNotApplicableException()
    dangerzone.errors.SuffixNotApplicableException: Cannot set a suffix after setting an output filename
    ```
    
2. While the "OCR document" checkbox is disabled, the language picker still works.
3. The "Open safe documents after converting" option no longer works, because the list there includes PDF readers. For example, in my environment, MuPDF cannot open the produced PNG.
4. The fact that `-page-<N>` is appended to the output should affect the "Save as" checkbox. This is necessary because we otherwise don't have a nice way to show in our UI where the sanitized files are stored.
6. This feature applies to all selected files, regardless of their filetype. This means that if a user wants to sanitize multiple files in a directory, and keep images as images, they would need to sanitize the images first, and then the documents.

Most importantly though, I'm not sure if this feature solves https://github.com/freedomofpress/dangerzone/issues/629 adequately. As a user of this tool, what I've wanted in several cases was a way to **preserve** the original file type. That is:
* If I give you an image, please give me back an image. It doesn't have to be exact same file type (JPG/TIFF/etc), just a reasonably popular image format (here, PNG works great).
* If I give you a document, give me back a PDF. Even if it's just a single page, since I may want OCR on it.

I have a feeling that this behavior could also be a default, but I'd like to know what you think here. We could of course have an option that converts everything to PDF, for advanced users who want the original behavior back. Finally, I think we can rework this PR to make this behavior possible, if we both like it. Very roughly:
1. We could check the extension on the client side and the number of pages we get back. In case it's an image type and the number of pages is 1, it's probably an image type.
    * Ideally, once the server-side sends back the file type, we can ditch this heuristic.
3. The safe document can be named `document-safe.png`, and we can communicate this in the Dangerzone UI.
4. We can add an extra line for the application that will open PNG files, with the similar logic we have for PDF files.

almet:
Thanks for the feedback!

> If I give you an image, please give me back an image. It doesn't have to be exact same file type (JPG/TIFF/etc), just a reasonably popular image format (here, PNG works great); If I give you a document, give me back a PDF. Even if it's just a single page, since I may want OCR on it.

An interesting feature of being able to convert images to PDFs is the fact that it does OCR on them.

We probably want to have an interface that makes it possible to select the output format. We could have a dropdown with these options:

Format of the output document:
- "Convert to images" ;
- "Convert to PDF" (in which case we offer OCR options) ;
- "Automatic detection" (the default), which will convert documents to PDFs, and images to images.

apyrgio:
> We probably want to have an interface that makes it possible to select the output format. We could have a dropdown with these options:
> 
> Format of the output document:
> - "Convert to images" ;
> - "Convert to PDF" (in which case we offer OCR options) ;
> - "Automatic detection" (the default), which will convert documents to PDFs, and images to images.

I agree, if we want to be exhaustive we can offer three options. In that case though, if we know that something is an image, it's probably best to skip the `-page-1` suffix altogether. But that's minor.

How do you want to handle this for this PR then? Also, I guess we're not considering it for inclusion in 0.11.0, right?

(Oh, btw, the option to convert to images could help with redactions btw, in the spirit of https://github.com/firstlookmedia/pdf-redact-tools)

PR #1493, opened 2026-06-09
Prepare 0.11.0

almet:
Bump versions and prepare the ground for the upcoming release


docs/developer/release/pre-release.md
```diff
@@ -14,7 +14,6 @@ run from the developer's laptop and are not tied to any build environment.
 - [ ] Update the "Version" field in `install/linux/dangerzone.spec`
 - [ ] (optional) Bump image version from `v1` to `v2`, if the API has changed.
 - [ ] Bump the Debian version by adding a new changelog entry in `debian/changelog`
-- [ ] Bump the dates and versions in the `Dockerfile.env`
```
apyrgio, line 17: We need to keep somewhere a mention that we should bump the dependencies of the container image. I think those are just the `Dockerfile.env` though. How do you propose we handle it?

docs/developer/release/pre-release.md
apyrgio:
```suggestion
- [ ] Check for new [WiX releases](https://github.com/wixtoolset/wix/releases) and update it if needed (see [#1200](https://github.com/freedomofpress/dangerzone/issues/1200))
```

docs/developer/release/pre-release.md
almet: As this is now handled by the dangerzone-image repository, I believe we should do that as part of releasing a new container image, don't you think?

docs/developer/release/pre-release.md
almet: Done, thanks!

dev_scripts/generate-release-tasks.py
apyrgio, line 1: This is not a comment about the file, but about the commit that includes that file. Let's change "stall link" to "stale link".

dangerzone/gui/log_window.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
apyrgio: Regarding this line, and other similar ones. My understanding is that the annotations feature has been implemented in a different way in Python 3.14, and will be deprecated in those versions. See https://docs.python.org/3/library/__future__.html#id2.

I'm not sure what's the way forward here for older Python versions (I guess stringify class names until Python 3.13 is EOL?), but we can defer this for later.

dangerzone/settings.py
```diff
@@ -114,11 +115,12 @@ def load(self) -> None:
...
-                    elif key == "updater_latest_version":
-                        if version.parse(get_version()) > version.parse(self.get(key)):
-                            self.set(key, get_version())
+                    elif key == "updater_latest_version" and version.parse(
+                        get_version()
+                    ) > version.parse(self.get(key)):
+                        self.set(key, get_version())
 
-            except Exception as e:
+            except json.JSONDecodeError as e:
```
apyrgio: If the settings file doesn't exist, we still need to load the default settings, and this exception does not catch that.

dangerzone/startup.py
```diff
@@ -255,7 +257,7 @@ def run(self) -> None:
             if report.container_image_bump:
                 self.handle_container_update(report)
         elif isinstance(report, ErrorReport):
-            raise RuntimeError(report.error)
+            raise TypeError(report.error)
```
apyrgio: Why `TypeError` here? Here we handle an error during runtime, not an operation that applies to a wrong object type.

dangerzone/updater/cli.py
```diff
@@ -71,7 +69,7 @@ def upgrade() -> None:
         click.echo(
             f"✅ The local image {DEFAULT_IMAGE_NAME}@{manifest_digest} is already up to date"
         )
-    except Exception as e:
+    except errors.UpdaterError as e:
```
apyrgio: This is a catch-all exception to pretty-print errors, so I would still use `except Exception` here.

dangerzone/updater/registry.py
```diff
@@ -13,16 +15,7 @@
...
-    [
-        "application/vnd.docker.distribution.manifest.v1+json",
-        "application/vnd.docker.distribution.manifest.v1+prettyjws",
-        "application/vnd.docker.distribution.manifest.v2+json",
-        "application/vnd.oci.image.manifest.v1+json",
-        IMAGE_LIST_MEDIA_TYPE,
-        IMAGE_INDEX_MEDIA_TYPE,
-    ]
-)
+ACCEPT_MANIFESTS_HEADER = f"application/vnd.docker.distribution.manifest.v1+json,application/vnd.docker.distribution.manifest.v1+prettyjws,application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json,{IMAGE_LIST_MEDIA_TYPE},{IMAGE_INDEX_MEDIA_TYPE}"
```
apyrgio: Sigh, this makes things less readable, and my intention was to easily cross-reference the supported manifest types versus the Podman/cosign/crane ones. May I suggest a multi-line string instead instead?

```
ACCEPT_MANIFESTS_HEADER = f"""\
application/vnd.docker.distribution.manifest.v1+json,\
application/vnd.docker.distribution.manifest.v1+prettyjws,\
application/vnd.docker.distribution.manifest.v2+json,\
application/vnd.oci.image.manifest.v1+json,\
{IMAGE_LIST_MEDIA_TYPE},\
{IMAGE_INDEX_MEDIA_TYPE}\
"""
```

dangerzone/updater/releases.py
```diff
@@ -105,18 +106,9 @@ def fetch_github_release_info() -> tuple[str, str]:
...
-            f"Encountered an exception while checking {GH_RELEASE_URL}: {e}"
-        )
 
-    if res.status_code != 200:
-        raise RuntimeError(
-            f"Encountered an HTTP {res.status_code} error while checking"
-            f" {GH_RELEASE_URL}"
-        )
+    res = requests.get(GH_RELEASE_URL, timeout=REQ_TIMEOUT)
+    res.raise_for_status()
```
apyrgio: Hm, I'm not sure what are the side effects of this change. I can say for sure that we will no longer raise RuntimeError, and that the message will be different. I think this message is what we show to the user, which is why I wanted to control it.

dangerzone/updater/installer.py
```diff
@@ -51,7 +46,7 @@ def apply_installation_strategy(strategy: Strategy) -> None:
         log.debug("Download and install a remote container image")
         container_name = runtime.expected_image_name()
 
-        remote_digest, remote_log_index, signatures = get_remote_digest_and_logindex(
+        remote_digest, _remote_log_index, signatures = get_remote_digest_and_logindex(
```
apyrgio: What's the best etiquette here? On new code, I think we just use `_` when we don't keep the result of a call. Should we do the same here as well, and in some other instances I saw?

dangerzone/updater/signatures.py
```diff
@@ -693,15 +665,15 @@ def install_local_container_tar(
     if bypass_signature_checks():
         return runtime.load_image_tarball(tarball_path)
     else:
-        image_name, image_digest = upgrade_container_image_airgapped(
+        _image_name, image_digest = upgrade_container_image_airgapped(
```
apyrgio: Another instance where we can use `_` instead.

tests/test_signatures.py
```diff
@@ -278,7 +274,7 @@ def test_get_remote_digest_and_logindex_when_remote_image_available(
     )
 
     # Call is_update_available
-    digest, log_index, signatures = get_remote_digest_and_logindex(
+    digest, log_index, _signatures = get_remote_digest_and_logindex(
```
apyrgio: Suggestion to replace it with `_` here.

dangerzone/updater/releases.py
apyrgio: Oh, I just saw that you reinstated the original logic in the last commit of the PR. Ignore this then.

dangerzone/updater/registry.py
almet: Okay, this was applied automatically by ruff and I agree with you: it's less readable. I've added a linter exception here.

dangerzone/gui/log_window.py
almet: You're correct! I completely missed this deprecation! I've changed and made use of stringified types instead for circular imports.

dangerzone/settings.py
almet: Good catch, thanks.

dangerzone/startup.py
almet: You're correct, this is fixed now.

dangerzone/updater/cli.py
almet: I guess we should let click handle this in the end! I went back to catching all exceptions for now!

dangerzone/updater/installer.py
almet: I was actually surprised when ruff proposed this prefixing. I guess it's a way to convey both what it is, and that we're not using it. Using hte bare `_` seem more readable to me, so I went away with this.

dangerzone/updater/signatures.py
almet: Thanks.

tests/test_signatures.py
almet: Applied.

apyrgio (approved):

PR #1494, opened 2026-06-10
chore(make): Remove Dockerfile job

apyrgio:
Remove the Make job that generates the Dockerfile, since this is now done in the freedomofpress/dangerzone-image repo. As a result, also remove the Jinja2 dependency, which is no longer used.


almet:
Thanks. I've included this in the `prepare-0.11.0` branch and pull request, and so am closing this one.

See https://github.com/freedomofpress/dangerzone/pull/1493

PR #1497, opened 2026-06-17
feat(dev_scripts): Create a Dangerzone environment from an APT/YUM repo

apyrgio:
Update `dev_scripts/env.py` to create an end-user Dangerzone environment, by installing Dangerzone from an APT/YUM repo, instead of installing it via a local package, as was the situation so far.

The script accepts an optional `--prod` or a `--qa` flag, in order to pick which kind of APT/YUM repo we want. For more info, check out https://github.com/freedomofpress/packages.


docs/developer/release/prepare-build-envs.md
```diff
@@ -17,10 +17,11 @@
     If they key isn't found, add it to the Keychain by running:
 
     ```bash
-    xcrun notarytool store-credentials dz-notarytool-release-key --apple-id <email> --team-id <team ID>
+    xcrun notarytool store-credentials dz-notarytool-release-key --apple-id <email> --team-id 94ZZGGGJ3W
```
almet, line 20: Do we want this to be public?

dev_scripts/env.py
```diff
@@ -708,9 +742,14 @@ def env_build_dev(args):
 def env_build(args):
     """Invoke the 'build' command based on the CLI args."""
     env = Env.from_args(args)
+    if args.qa and args.prod:
+        raise ValueError("You need to pass either --qa or --prod")
```
almet: I was expecting this to handle local `.deb` installations as well, as it's been useful for me in the past when developing.

My understanding is that this PR changes this and makes it so that you either need to install from QA or Release, only. Can we continue offering the default behavior in addition to the extended one that downloads from the official repos?

almet (changes requested):
Thanks, this will prove useful!

It seems that we're stopping offering the possibility to test a locally built .deb with these changes. Are we sure that's what we want?

dev_scripts/env.py
apyrgio: It still does. This is to avoid providing **both** `--qa` and `--prod`. I can maybe rephrase this to:

> You cannot pass both --qa and --prod

docs/developer/release/prepare-build-envs.md
apyrgio: It's already public, since it's part of our code-signing certificate (see a few lines above).

dev_scripts/env.py
almet: Oh wow, interesting to see how the actual phrasing tainted my reading of the code! I really read `if not args.qa or args.prod` 🙄 

The new phrasing is way better, yes.

apyrgio: Fixed in https://github.com/freedomofpress/dangerzone/pull/1497/changes/f4a6ada28318838ccf7ace47b27c505709f8b3f2

almet (approved):
That makes sense, thanks!

PR #1502, opened 2026-06-22
Small docs update

micahflee:
Hello. It's been a long time since I've looked at the Dangerzone code. I just started setting up a development environment in macOS, and I noticed a few small documentation issues.

- First, in the readme, it still says Windows and Mac uses Docker Desktop, though that's not true.
- And second, in the macOS docs, it doesn't include setting `DANGERZONE_DEV=1`, which prevents the instructions from working without it.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -378,6 +378,12 @@ command:
 poetry run mazette install
 ```
 
+Make sure you the `DANGERZONE_DEV` for development:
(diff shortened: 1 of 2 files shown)


apyrgio (approved):

apyrgio:
Nice catch, merged!

PR #1506, opened 2026-06-29
Bump last known logindex

almet:
When doing 0.11.0 QA, I found out that the last known log-index wasn't bumped properly. As a result, the remote image was always preferred to the local embedded one.


apyrgio (approved):

apyrgio:
Looks good. We should fix our docs as well and re-add a mention to bumping the LOG_INDEX, but we can do that in a separate PR.

almet:
Yes, there will be a separate PR with documentation fixes (I have a bunch of them)

PR #1507, opened 2026-06-29
chore(docs): Set DANGERZONE_DEV envvar when testing on Windows

apyrgio:


BUILD.md
```diff
@@ -451,6 +451,10 @@ poetry run mazette install
 Run the following command to download the latest container image, or
 [build it locally](#building-a-local-container-image):
 
+```sh
+set DANGERZONE_DEV 1
```
almet: Is this working? [The docs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/set_1) seem to mention there should be an `=` sign.

apyrgio: Hm, I may have used the `=` variant, but got side-tracked by a similar line we have in our docs that didn't have the `=` sign. I have force-pushed this branch to fix both occurences.

BUILD.md
almet: Ah, interesting! it seem that actually this whole section need to go, as now this is handled by the `dangerzone-image` repo!

apyrgio: Nice find, fixed in 7d8c93a3056f9784d3544437a90f7570e51dc4c5 (where I've squashed the other commits as well mostly for simplicity reasons).

deeplow (commented):
This is failing for me.

<details><summary> (old and wrong hypothesis)</summary>

and I think the reason is that at no point is the environment variable transformed into the `sys` setattr "hack" that `get_resource_path` assumes. Therefore it just tries to find it in the poetry virtualenv directory and then fails to load the image name.

**Update** the code path is the same on linux, and there the env var is loaded. Therefore it's a problem with the setting of environment variables.

</details>

Here's the traceback:

```python
PS C:\Users\user\dangerzone> poetry run dangerzone-image prepare-archive --output .\share\container.tar
  File "<string>", line 1, in <module>
    import sys; from importlib import import_module; sys.argv = ['C:\\Users\\user\\AppData\\Local\\pypoetry\\Cache\\virtualenvs\\dangerzone-ul5J_m7X-py3.13\\Scripts\\dangerzone-image.cmd', 'prepare-archive', '--output', '.\\share\\container.tar']; sys.exit(import_module('dangerzone.updater.cli').run())
                                                                                                                                                                                                                                                                 ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Python\pythoncore-3.13-64\Lib\importlib\__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1395, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 1023, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "C:\Users\user\dangerzone\dangerzone\updater\cli.py", line 17, in <module>
    DEFAULT_IMAGE_NAME = expected_image_name()
  File "C:\Users\user\dangerzone\dangerzone\container_utils.py", line 375, in expected_image_name
    return image_name_path.read_text().strip("\n")
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\user\AppData\Local\Python\pythoncore-3.13-64\Lib\pathlib\_local.py", line 546, in read_text
    return PathBase.read_text(self, encoding, errors, newline)
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Python\pythoncore-3.13-64\Lib\pathlib\_abc.py", line 632, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors, newline=newline) as f:
         ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Python\pythoncore-3.13-64\Lib\pathlib\_local.py", line 537, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\user\\AppData\\Local\\pypoetry\\Cache\\virtualenvs\\dangerzone-ul5J_m7X-py3.13\\Scripts\\share\\image-name.txt
```

BUILD.md
deeplow: I have confirmed that `=` is needed.

BUILD.md
```diff
@@ -433,10 +435,11 @@ poetry run mazette install
 ```
 
 Run the following command to download the latest container image, or
-[build it locally](#building-a-local-container-image):
+[use a local one](#using-a-local-container-image):
 
 ```sh
-poetry run dangerzone-image prepare-archive --output share/container.tar
+set DANGERZONE_DEV=1
```
deeplow: Turns out I was using powershell, where it needs to be [`$Env:DANGERZONE_DEV = 1`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.6).

Looking more carefully, first windows command we mention tells the user to use powershell, and the second says `cmd.exe`:

> Install [poetry](https://python-poetry.org/). Open PowerShell, and run:
>
> ```
> python -m pip install poetry
> ```
> 
> Install git from [here](https://git-scm.com/download/win), open a Windows terminal (`cmd.exe`) and clone this repository:
>
> [...]

So technically `set DANGERZONE_DEV=1` was correct. But we might consider also providing a way to do it in powershell, since everything is basically the same up until this point. Thoughts @almet?

almet: I've just tried and the following works on Windows powershell. As the default shell these days seem to be powershell (what a world), we might want to include these as default instructions for Windows devs:

```powershell

$Env:DANGERZONE_DEV = 1 ; poetry run dangerzone
```

deeplow: I have added this now and also set powershell to the default.

deeplow:
@almet this is ready for a review pass.

dev_scripts/qa.py
```diff
@@ -378,14 +380,16 @@
 
 Install Microsoft Visual C++ 14.0 or greater. Get it with ["Microsoft C++ Build Tools"](https://visualstudio.microsoft.com/visual-cpp-build-tools/) and make sure to select "Desktop development with C++" when installing.
 
-Install [poetry](https://python-poetry.org/). Open PowerShell, and run:
+Install git from [here](https://git-scm.com/download/win), open a Windows terminal (`cmd.exe`) and clone this repository:
```
almet: `cmd.exe` actually opens a "command prompt" interface, which is different from the PowerShell experience we want.

Instead, we could ask the user to open the `Terminal` application, which comes installed with Windows as the default since Windows 11 22H2. I believe that PowerShell is the default in there.

BUILD.md
```diff
@@ -405,14 +407,16 @@ Install the latest version of Python 3.13 (64-bit) [from python.org](https://www
 
 Install Microsoft Visual C++ 14.0 or greater. Get it with ["Microsoft C++ Build Tools"](https://visualstudio.microsoft.com/visual-cpp-build-tools/) and make sure to select "Desktop development with C++" when installing.
 
-Install [poetry](https://python-poetry.org/). Open PowerShell, and run:
+Install git from [here](https://git-scm.com/download/win), open a Windows terminal (`cmd.exe`) and clone this repository:
+
+Then open PowerShell, where all the remaining commands will be ran.
```
almet: Ditto

almet (changes requested):
I think we should default to ask the user to use the Windows "Terminal" app, and I've commented this way.

deeplow:
Just trying to wrap this one up. @almet I have now set it to Windows Terminal. I haven't confirmed that it runs powershell in Windows 11 (I don't have a device with it).

I have also moved the git installation to the top, before all the terminal commands. This way all programs needed for installation are grouped at the top.

almet (approved):
Thanks, the changes look good to me, let's merge 👍🏼

PR #1519, opened 2026-07-08
gui: Respect disabling update checks

### LINKED ISSUE #1517: Users who don't want sandbox updates are always prompted on startup to enable them
In the latest Dangerzone release, if a user has expressed that they don't want to check for updates, they will still get prompted every time they open Dangerzone.

The bug is in these lines:

https://github.com/freedomofpress/dangerzone/blob/42a6d41a1f76dcedbce861031a00d8d54074dffe/dangerzone/updater/releases.py#L162-L172

Besides the fact that the `if` / `elif` condition is the same, we never return `False` to the caller.

apyrgio:
If the user decides to disable update checks, we should respect their choice. The only exception should be in cases where there is no container image locally, or not bundled within Dangerzone. The current behavior is to always ask the user to enable update checks, but that is not helpful in airgapped installations.

This commit basically reinstates the v0.10.0 behavior in the `should_check_for_updates` function and `test_user_prompts` test.

Fixes #1517

### DIFF
--- a/dangerzone/updater/releases.py
+++ b/dangerzone/updater/releases.py
@@ -138,34 +138,33 @@ def fetch_github_release_info() -> tuple[str, str]:
 def should_check_for_updates(settings: Settings) -> bool:
     """Determine if we can check for release updates based on settings and user prefs.
 
-    Note that this method only checks if the user has expressed an interest for
-    learning about new updates, and not whether we should actually make an update
-    check. Those two things are distinct, actually. For example:
+    This method checks if the user has expressed an interest for learning about new
+    updates, or if updates are mandatory. Then, it either returns:
+    * True: if the user has explicitly requested to check for updates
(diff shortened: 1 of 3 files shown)


dangerzone/updater/releases.py
```diff
@@ -138,34 +138,33 @@ def fetch_github_release_info() -> tuple[str, str]:
...
 
-    if check is False or check is None:
-        if check is None:
-            log.debug("User has not been asked yet for update checks")
-
-        if not is_container_tar_bundled() and not is_container_image_installed():
-            log.debug("No container available, prompting user to enable updates")
-            raise errors.NeedUserInputNoContainer()
+    check = settings.get("updater_check_all")
+    if check is None:
```
almet, line 166: Thanks, this code path is easier to grasp.

almet (approved):
These changes look good to me, and the code makes more sense than the previous complicated decision tree.

PR #1521, opened 2026-07-14
ci: Bump versions of GHA dependencies

apyrgio:
This PR supersedes some PRs that Dependabot has opened in the past.


deeplow (approved):
LGTM. It was not super obvious what the version 7.0.0 was as they probably switch the tag for folks to keep it pinned to the specific version series. But I eventually got there :)

Issue #206, opened 2022-09-15
Consider adding support for Ubuntu Focal


eloquence:
Ubuntu 20.04 is still supported until April 2025, so we should consider adding it to the list of supported releases for users who haven't moved to Jammy yet.

deeplow:
Apparently podman (on which the linux version depdends) is [not available](https://packages.ubuntu.com/focal/allpackages) on ubuntu focal and it requires [external repos](https://www.atlantic.net/dedicated-server-hosting/how-to-install-and-use-podman-on-ubuntu-20-04/) (linked [from the podman blog](https://podman.io/blogs/2021/06/16/install-podman-on-ubuntu.html)). 

And actually we do test podman on this exact ubuntu version (it's [where the CI tests the document conversion process](https://github.com/freedomofpress/dangerzone/blob/main/.circleci/config.yml#L130)).  And dangerzone had been available in the past for that ubuntu version.

@apyrgio will add the documentation to the installation instructions after successful testing & confirmation the packagecloud deb works for that version.

deeplow:
It doesn't support python3.10 ([which we now need)](https://github.com/freedomofpress/dangerzone/pull/161) unless we add the deadsnakes as a dependency

apyrgio:
Our Packagecloud repos for Ubuntu Focal do have a Dangerzone package, but the latest version is 0.2.1, which is the last version that supported Ubuntu Focal. See also https://github.com/freedomofpress/dangerzone/issues/160#issuecomment-1138970038. This means that we need to partially revert 229ebbda1499d29ea44099c28db25a921513cf9d, and create a 0.4 Ubuntu Focal package, besides adding Podman instructions.

Aside from that, I'd like to ask a quick question:

> It doesn't support python3.10 ([which we now need)](https://github.com/freedomofpress/dangerzone/pull/161) unless we add the deadsnakes as a dependency 

I believe this is no longer accurate, right? We had some minor Mypy issues in Python 3.9, but they are not important enough to switch to Python 3.10.

apyrgio:
I can verify that we can install Dangerzone on Ubuntu Focal, if we install Podman first, with the instructions that we have in our CircleCI config. As for the .deb itself, I just downloaded it from Packagecloud, without adding the repo in the sources.

I haven't managed to test Dangerzone successfully though, because I encountered this issue:

```
[...]

PR #1528, opened 2026-07-21
Accept files from stdin, write safe PDFs to stdout

### LINKED ISSUE #1522: Accept files from stdin
**What is the feature you think should be a good addition to Dangerzone?**

Dangerzone CLI only accepts file paths as arguments. There's no way to pipe documents through stdin (e.g., `cat file.pdf | dangerzone` or `dangerzone - < file.pdf`), which is a standard Unix pattern needed for scripting and tool composition.

### Proposed Solution

- Accept `-` as a filename to explicitly signal stdin, or detect implicit stdin when no filenames are given and stdin is not a TTY
- Extend the `Document` class to accept raw `data: bytes` instead of a filename, with an `.open()` method that returns the appropriate `BinaryIO` (real file or `BytesIO`)
- When no `--output-filename` is given, write the safe PDF directly to stdout (via the isolation provider); refuse with a clear error if stdout is a TTY
- Route all non-PDF output to stderr via `click.echo(..., err=True)` so stdout stays clean
- Reject `--archive` with stdin input (no file to archive)

**Is your feature request related to a problem? Please describe.**

Nope, it's a good to have.

**Additional context**

N/A

apyrgio:
Dangerzone CLI now accepts documents via stdin and writes safe PDFs to stdout, following standard Unix pipe conventions:

```
cat file.pdf | dangerzone-cli > safe.pdf
dangerzone-cli - < file.pdf > safe.pdf
dangerzone-cli < file.pdf > safe.pdf
```                                                                                                                                                               

All status output (banner, progress, errors) goes to stderr so stdout stays clean for PDF data. We also take extra care to detect when we can read from stdin and when we can write to stdout, so that we don't mess with the user's TTY.

Fixes #1522
Refs #206

### DIFF
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -15,8 +15,8 @@ from .util import get_version, replace_control_chars
 
 
 def print_header(s: str) -> None:
-    click.echo("")
-    click.echo(Style.BRIGHT + s)
+    click.echo("", err=True)
+    click.echo(Style.BRIGHT + s, err=True)
 
 
 @click.command()
@@ -88,13 +88,16 @@ def run(
         if set_container_runtime == "default":
             settings.unset_custom_runtime()
             click.echo(
-                "Instructed Dangerzone to use the default container runtime for this OS"
+                "Instructed Dangerzone to use the default container runtime for this OS",
+                err=True,
             )
         else:
             container_runtime = settings.set_custom_runtime(
                 set_container_runtime, autosave=True
             )
-            click.echo(f"Set the settings container_runtime to {container_runtime}")
+            click.echo(
+                f"Set the settings container_runtime to {container_runtime}", err=True
+            )
         sys.exit(0)
     elif not filenames:
         raise click.UsageError("Missing argument 'FILENAMES...'")
@@ -123,9 +126,9 @@ def run(
                 valid = True
                 break
         if not valid:
-            click.echo("Invalid OCR language code. Valid language codes:")
+            click.echo("Invalid OCR language code. Valid language codes:", err=True)
             for lang in dangerzone.ocr_languages:
-                click.echo(f"{dangerzone.ocr_languages[lang]}: {lang}")
+                click.echo(f"{dangerzone.ocr_languages[lang]}: {lang}", err=True)
             sys.exit(1)
 
     tasks = []
@@ -150,7 +153,8 @@ def run(
                 + "No container image found."
                 + Style.RESET_ALL
                 + " Please initialize Dangerzone by running:\n\n"
-                "    dangerzone-image upgrade\n"
+                "    dangerzone-image upgrade\n",
+                err=True,
             )
             sys.exit(1)
         print_header("Converting document(s) to safe PDF")
@@ -178,7 +182,7 @@ def run(
     if documents_failed != []:
         print_header("Failed to convert document(s)")
         for document in documents_failed:
-            click.echo(replace_control_chars(document.input_filename))
+            click.echo(replace_control_chars(str(document)), err=True)
         sys.exit(1)
     sys.exit(0)
 
--- a/dangerzone/cli.py
+++ b/dangerzone/cli.py
@@ -231,7 +231,10 @@ def display_banner() -> None:
     ╰──────────────────────────╯
     """
 
-    print(Back.BLACK + Fore.YELLOW + Style.DIM + "╭──────────────────────────╮")
+    print(
+        Back.BLACK + Fore.YELLOW + Style.DIM + "╭──────────────────────────╮",
+        file=sys.stderr,
+    )
     print(
         Back.BLACK
         + Fore.YELLOW
@@ -242,7 +245,8 @@ def display_banner() -> None:
         + "           ▄██▄           "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -254,7 +258,8 @@ def display_banner() -> None:
         + "          ██████          "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -266,7 +271,8 @@ def display_banner() -> None:
         + "         ███▀▀▀██         "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -278,7 +284,8 @@ def display_banner() -> None:
         + "        ███   ████        "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -290,7 +297,8 @@ def display_banner() -> None:
         + "       ███   ██████       "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -302,7 +310,8 @@ def display_banner() -> None:
         + "      ███   ▀▀▀▀████      "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
@@ -314,7 +323,8 @@ def display_banner() -> None:
         + "     ███████  ▄██████     "
         + Fore.YELLOW
         + Style.DIM
-        + "│"
+        + "│",
+        file=sys.stderr,
     )
     print(
         Back.BLACK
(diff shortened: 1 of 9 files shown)


dangerzone/isolation_provider/base.py
```diff
@@ -315,10 +315,8 @@ def drain_ocr_futures(block_until_below: int | None = None) -> None:
         # Ensure nothing else is read after all bitmaps are obtained
         p.stdout.close()
 
-        # Saving it with a different name first, because PyMuPDF cannot handle
-        # non-Unicode chars.
-        safe_doc.save(document.sanitized_output_filename)
-        os.replace(document.sanitized_output_filename, document.output_filename)
+        # Write the safe PDF to the document's destination.
+        document.write(safe_doc.tobytes())
```
almet, line 319: Does this changes the default CLI user experience?

I believe we should output the files somewhere the user has specified, and if nothing was specified, error out and ask where this should be.

Presenting the user with the contents of the file should be done only if it was asked explicitely, e.g. by setting input and output to `-` (to mean it's should be read and written to a STDIN and STDOUT)

dangerzone/cli.py
```diff
@@ -15,11 +15,56 @@
...
+            raise click.BadOptionUsage(
+                option_name="--output-filename",
+                message="--output-filename can only be used with one input file",
+            )
+
+    if filenames == [args.STDIO_DESCRIPTOR] or not filenames:
+        # We are reading a single document from stdin.
+        if archive:
+            raise click.UsageError("--archive cannot be used with input from stdin")
+        if sys.stdin.isatty():
```
almet: I had to look up how this works, and so suggest adding a comment (it wasn't clear with the error message on its own for me).

```suggestion
        # if stdin is a TTY, it's not possible to read from it.
        if sys.stdin.isatty():
```

dangerzone/cli.py
```diff
@@ -15,11 +15,56 @@
...
+        dangerzone.add_document_from_filename(filename, output_filename, archive)
+
+
+@click.command(
+    help=(
+        "Convert potentially dangerous documents to safe PDFs.\n\n"
+        "Accepts file paths as arguments, or reads from stdin when no\n"
+        "files are given (or when '{args.STDIO_DESCRIPTOR}' is passed as a filename).\n"
+        "When reading from stdin, the safe PDF is written to stdout unless\n"
+        "--output-filename is specified. All status output goes to stderr."
```
almet: Can this be added in the docstring instead? (Click picks it up)

dangerzone/cli.py
```diff
@@ -168,7 +208,10 @@ def run(
     if documents_safe != []:
         print_header("Safe PDF(s) created successfully")
         for document in documents_safe:
-            click.echo(replace_control_chars(document.output_filename))
+            # When writing to stdout (data-based doc, no output filename),
+            # skip printing the filename — the PDF is already on stdout.
```
almet: We might want to print to `STDERR` the fact that the output will be available to `STDOUT` rather than skipping.

dangerzone/document.py
```diff
@@ -173,8 +192,33 @@ def default_output_filename(self) -> str:
         return f"{os.path.splitext(self.input_filename)[0]}{self.suffix}"
 
     def announce_id(self) -> None:
-        sanitized_filename = util.replace_control_chars(self.input_filename)
-        log.info(f"Assigning ID '{self.id}' to doc '{sanitized_filename}'")
+        if self._input_filename is not None:
+            log.info(
+                f"Assigning ID '{self.id}' to doc '{self.sanitized_input_filename}'"
```
almet: nit: We could leverage the `__str__` method here rather than having two "if" branches.

dangerzone/logic.py
```diff
@@ -42,6 +42,10 @@ def add_document_from_filename(
         doc = Document(input_filename, output_filename, archive=archive)
         self.add_document(doc)
 
+    def add_document_from_stdin(self, output_filename: str | None = None) -> None:
```
almet: There seem to be a pattern repeating here, where we seem to be passing plates around, as convenience.

Should we try to remove a layer of indirection?

almet (changes requested):
Amazing, thanks :-) I've stumbled onto this need in the past when building a web interface on top of DZ, and I can clearly see how that can be useful in general.

I'm requesting some changes on this PR, but it's headed the right direction! Thanks.

dangerzone/isolation_provider/base.py
apyrgio: Ok, I've made the input from stdin and output to stdout a conscious user choice in 6b065a6c. This means that:

```
dangerzone-cli < file.pdf > safe.pdf
```

becomes:

```
dangerzone-cli - -o - < file.pdf > safe.pdf
```

which is more dash-y, but ultimately fine.

Note that with the above change, we no longer check for TTYs, since this no longer happens implicitly.

dangerzone/cli.py
apyrgio: This no longer applies to the new changes, since we no longer check for TTYs (see https://github.com/freedomofpress/dangerzone/pull/1528#discussion_r3720960495)

dangerzone/cli.py
apyrgio: This was done on purpose actually (also, I think Click shows the docstring as a help message as well). I wanted a quick way to explain how to use `dangerzone-cli`, for those who encounter it for the first time. To better explain my end goal, I've rephrased and added extra stuff in 3391a26d. It now looks like this:

```
[user@dangerzone-dev dangerzone]$ poetry run dangerzone-cli --help
Usage: dangerzone-cli [OPTIONS] [FILENAMES]...

  Convert potentially dangerous documents to safe PDFs.

  Documents are converted inside a sandbox, so that any embedded threats are
  neutralized. You can also use OCR to add a searchable text layer to the safe
  PDF, via the --ocr-lang option.

  Pass one or more file paths as arguments, or use '-' to read a document from
  standard input. For single-document conversions, you can write the safe PDF
  to a file with -o <file>, or to standard output with -o -.

  Examples:

      dangerzone-cli evidence.odt                      # create evidence-safe.pdf
      dangerzone-cli doc1.pdf doc2.docx                # create doc1-safe.pdf and doc2-safe.pdf
      dangerzone-cli --archive report.docx             # create report-safe.pdf and move the original under 'unsafe/'
      dangerzone-cli --ocr-lang eng scan.pdf           # add a searchable text layer
      dangerzone-cli -o safe.pdf evidence.odt          # create safe.pdf
      dangerzone-cli -o - evidence.odt > safe.pdf      # write the safe PDF to stdout
      cat evidence.pdf | dangerzone-cli - -o safe.pdf  # read from stdin and write to safe.pdf
      cat in.pdf | dangerzone-cli - -o - > out.pdf     # read from stdin to write to stdout
```

It's probably a little verbose and I can drop some examples, but I mostly want to know if you find this valuable.

dangerzone/cli.py
apyrgio: Done in 6b065a6cccf4a745a8651ab2c3656e4bbdcb0454. The `dangerzone-cli` now reports:

```
Safe PDF written successfully to stdout
```

dangerzone/document.py
apyrgio: Done in 32d2ac07b2f7d6d30838ff781ad53a2dd82aba2f

dangerzone/logic.py
apyrgio: Done in a88ba4c1106f16a42b933c5be31b3dfb53b1d124.

dangerzone/isolation_provider/base.py
almet: If you'd like to get a less dashy experience, it's also possible to have a `--stdio` argument, for instance.

dangerzone/cli.py
almet: Can you clarify why using docstrings didn't work? From my perspective, it looks like a way to handle the spacing and returns automatically, but I might be missing something.

Otherwise, I really like the idea of having documentation in the CLI like that, with examples it's great :-)

The example that reads "# write the safe PDF to stdout" should probably be renamed to "use unix pipes to convert to safe.pdf", or we should remove the last `>`.

dangerzone/isolation_provider/base.py
apyrgio: Hm, people may be more familiar with the dashes than with a custom `--stdio` flag, so I'm ultimately ok with it.

dangerzone/cli.py
apyrgio: > Can you clarify why using docstrings didn't work? From my perspective, it looks like a way to handle the spacing and returns automatically, but I might be missing something.

Oh I'm sorry! When I read your initial message, I thought that you _didn't_ want the examples in the help message, and wanted to move them in the docstring (hence my comment).

I agree, I've moved them to the docstring and tweaked a bit the sentence you pointed out, as well as the others. See a1a8d3efce698d2f418825c40270646578993f57.

almet:
I'm not sure what's the status of this PR now. @apyrgio let me know if you need anything from me here.

apyrgio:
I've addressed your comments in this PR, and from our live discussion, I _think_ the resolution of those comments looked good. I've moved forward then and rebased on top of `main`, fixed conflicts, and force-pushed a clean version of the branch here.

almet (approved):
This LGTM, thanks, let's merge!

PR #1534, opened 2026-07-29
Make `dangerzone-image prepare-archive` work with new buildkit versions

almet:
Accept OCI image indexes as multi-arch container images In addition to Docker manifest lists.

BuildKit 0.31.0 and later pushes OCI image indexes by default, which made `dangerzone-image prepare-archive` fail with `InvalidMutliArchImage` against newly published images.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,15 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## [Unreleased](https://github.com/freedomofpress/dangerzone/compare/v0.11.0...HEAD)
 
+### Fixed
+
+- Accept OCI image indexes as multi-arch container images, in addition to Docker manifest lists. BuildKit 0.31.0 and later pushes OCI image indexes by default, which made `dangerzone-image prepare-archive` fail with `InvalidMutliArchImage` against newly published images ([#1534](https://github.com/freedomofpress/dangerzone/pulls/1534)).
+
 ### Development changes
 
 - Add a security policy for our project
   ([#1278](https://github.com/freedomofpress/dangerzone/issues/1278))
+
 
 ## [0.11.0](https://github.com/freedomofpress/dangerzone/releases/tag/v0.11.0)
 
--- a/dangerzone/updater/registry.py
+++ b/dangerzone/updater/registry.py
@@ -124,8 +124,8 @@ def get_digest_for_arch(image_str: str, architecture: str) -> str:
     """Return the digest of the matching architecture, for the specified image, without the sha256: prefix"""
     manifest = get_manifest(image_str).json()
 
-    if manifest.get("mediaType") != IMAGE_LIST_MEDIA_TYPE:
-        raise errors.InvalidMutliArchImage()
+    if manifest.get("mediaType") not in (IMAGE_LIST_MEDIA_TYPE, IMAGE_INDEX_MEDIA_TYPE):
+        raise errors.InvalidMultiArchImage()
 
     arch_manifests = [
         m["digest"].replace("sha256:", "")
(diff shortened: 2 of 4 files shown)


apyrgio:
Other than `dangerzone-image prepare-archive`, is regular Dangerzone affected? Asking to make sure that previous Dangerzone releases can handle the new image types.

almet:
They're not affected, this only affects the "prepare-archive" codepath

deeplow:
Look at the code, that's also the only code path I see directly into `IMAGE_LIST_MEDIA_TYPE`.

But `get_manifest` has an `Accept` header that I supposed could lead to a refusal in fetching the resource:

```python
def get_manifest(image_str: str) -> requests.Response:
    """Get manifest information for a specific tag"""
    image = parse_image_location(image_str)
    if image.digest:
        manifest_url = f"{_url(image)}/manifests/{image.digest}"
    else:
        manifest_url = f"{_url(image)}/manifests/{image.tag}"
    headers = {
        "Accept": ACCEPT_MANIFESTS_HEADER,
    }
    headers.update(_get_auth_header(image))

    ...
```

But I'm not sure which comment this is ultimately called in.

deeplow:
@apyrgio for now the plan is to release with a buildkit version that regenerated the old headers, so I don't think we're run the risk download failures in this next release, but we should definitely have this in the release test plan (just as a sanity-check).

almet:
> But get_manifest has an Accept header that [...] could lead to a refusal in fetching the resource:

Don't worry :-) 

The client provides the `Accept` header of all the resources it can accept, and the server decides to return them if it has content types that match. Here, `IMAGE_INDEX_MEDIA_TYPE` and `IMAGE_LIST_MEDIA_TYPE` are part of the `Accept` header provided to the server.

I validated that in practice, upgrading Dangerzone to images built with BuildKit v0.31.0 (using the new OCI Media Type, by changing `share/dangerzone-image.txt`) works properly, and so only the `prepare-archive` code path would be impacted.

We decided to pin images to BuildKit 0.30.0 for now, until we have a way to prepare images with at least one version of DZ (that will be 0.11.1).

deeplow:
OK. Great. Thanks for the due diligence!

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ since 0.4.1, and this project adheres to [Semantic Versioning](https://semver.or
 
 ## [Unreleased](https://github.com/freedomofpress/dangerzone/compare/v0.11.0...HEAD)
 
+### Fixed
+
+- Accept OCI image indexes as multi-arch container images, in addition to Docker manifest lists. BuildKit 0.31.0 and later pushes OCI image indexes by default, which made `dangerzone-image prepare-archive` fail with `InvalidMutliArchImage` against newly published images.
```
apyrgio:
```suggestion
- Accept OCI image indexes as multi-arch container images, in addition to Docker manifest lists. BuildKit 0.31.0 and later pushes OCI image indexes by default, which made `dangerzone-image prepare-archive` fail with `InvalidMutliArchImage` against newly published images ([#1534](https://github.com/freedomofpress/dangerzone/pulls/1534)).
```

apyrgio (approved):

apyrgio:
Tested it out locally, and I think we're good, aside from some small comments. Thanks!

dangerzone/updater/registry.py
```diff
@@ -124,7 +124,7 @@ def get_digest_for_arch(image_str: str, architecture: str) -> str:
     """Return the digest of the matching architecture, for the specified image, without the sha256: prefix"""
     manifest = get_manifest(image_str).json()
 
-    if manifest.get("mediaType") != IMAGE_LIST_MEDIA_TYPE:
+    if manifest.get("mediaType") not in (IMAGE_LIST_MEDIA_TYPE, IMAGE_INDEX_MEDIA_TYPE):
         raise errors.InvalidMutliArchImage()
```
apyrgio: Unrelated, there's a pre-existing typo here:


```suggestion
        raise errors.InvalidMultiArchImage()
```

We can merge as is of course, I'll leave it up to you.

PR #1539, opened 2026-08-04
Fix new ruff rule violations

almet:
ruff recently introduced some new rules by default. This PR aims to fix the ones that were now failing.


apyrgio (approved):

apyrgio:
Those changes look good, let's see if our CI tests agree.

PR #1544, opened 2026-08-05
Respect user choice when skipping updates

### LINKED ISSUE #1543: Skipping updates is not respected
If a new container update is available, you asked to be prompted before applying the updates, and you ask to skip this update, your choice is currently not taken into account and the container is downloaded anyway.

It's possible that this will be fixed by @apyrgio's work fixing #1517.

almet:
To merge after #1519, as it touches code in the same places.

Fixes #1543


apyrgio (approved):

apyrgio:
Nice catch about the missing return. I've approved the PR, even though the CI tests don't work. I'll try to figure out what's wrong with them, else you can give it a shot when you're back.

almet:
We might just want to re-apply these changes manually on top of `main`. One commit is here and shouldn't be, and the fix here is really a `return` statement, and so shouldn't make the tests fail like we see in the current CI.

almet:
(Just rebased this, let's see how the CI goes)

PR #1548, opened 2026-08-11
tests: Destroy Qt objects left behind by GUI tests

### LINKED ISSUE #1547: Flakiness in GUI tests during teardown
Our GUI tests have started failing on process teardown, especially on Windows 2022. This happens even though we run each GUI test in a separate process (see #493). Now that we add more and more GUI tests, we may see this racy behavior more often. We should find out the root cause and fix it.

apyrgio:
By the time a GUI test is over, pytest-qt has called `close()` and `deleteLater()` on every widget registered with `qtbot.addWidget()`, and has run `processEvents()` a few times. That is not enough to actually destroy them:

* `deleteLater()` merely posts a DeferredDelete event, and Qt delivers those only from within a running event loop. There is none at teardown time, so the deletions pile up unprocessed.
* The widgets that the application creates without a parent, such as `MainWindow.file_dialog` or the various `Alert` dialogs, are not registered with pytest-qt in the first place.

The result is that a Qt test leaves roughly a dozen live widgets behind, and since pytest holds on to the fixtures that reference them for the whole session, they are only ever destroyed by the interpreter's final garbage collection pass -- in an order that PySide does not control, and possibly after the QApplication itself is gone. On Windows this occasionally faults *after* pytest has already reported the test as passed, which fails the CI run with no indication of which test process died.

So we destroy them here instead, while the QApplication is still alive and we are still in control of the ordering.

Refs #493
Closes #1547

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,11 +11,7 @@ fix: ## apply all the suggestions from ruff
 
 .PHONY: test
 test: ## Run the tests
-	# Make each GUI test run as a separate process, to avoid segfaults due to
-	# shared state.
-	# See more in https://github.com/freedomofpress/dangerzone/issues/493
-	pytest --co -q tests/gui | grep -e '^tests/' | xargs -n 1 pytest -v
-	pytest -v --cov --ignore dev_scripts --ignore tests/gui
+	pytest -v --cov --ignore dev_scripts
 
 .PHONY: poetry-install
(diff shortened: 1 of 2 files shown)


almet (approved):
If this indeed passes the CI and we can now run the tests in real parallel with pytest (and maybe pytest xdist then?), I'm all for it!

PR #1552, opened 2026-08-22
Default to the production apt repository

almet:
Before this change, `env.py` is always and only referring to the QA packages.


apyrgio (approved):

apyrgio:
D'oh, thanks!

PR #1557, opened 2026-09-07
Mention Winget in the Windows installation instructions

### LINKED ISSUE #1498: Add a mention to Winget as a possible installation method.
@almet found out that a user has added us to the Winget repository, similarly to how others have added us in the Homebrew repo: https://github.com/microsoft/winget-pkgs/tree/d7fb21f30069e0cc5a1e2dabf64df826474523cb/manifests/f/FreedomofthePressFoundation/Dangerzone/0.10.0

We should feature this installation method in our docs, and possibly in our https://dangerzone.rocks website as well.

marekl11:
Fixes #1498.

The Windows section of INSTALL.md only points at the release page, while the macOS section right above it already offers Homebrew as an alternative. This adds the equivalent line for Winget, in the same shape as the Homebrew one.

I checked that `FreedomofthePressFoundation.Dangerzone` is the current package identifier in winget-pkgs and that its latest manifest is 0.11.0, matching the current release.

No CHANGELOG entry: this documents an installation route that already exists rather than changing anything, so I read it as falling under the `no changelog` label. Happy to add an entry instead if you would rather have one.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,6 +53,8 @@ You can also install Dangerzone for Mac using [Homebrew](https://brew.sh/): `bre
 > [!TIP]
 > We generally support Windows releases that are still within [Microsoft’s servicing timeline](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet).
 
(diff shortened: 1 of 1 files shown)


apyrgio (approved):

apyrgio:
Thanks for the PR. We plan to also add this information in https://dangerzone.rocks, alongside a note about the Microsoft Store (we plan to publish soon there).

